PR #8, opened 2020-06-07
Make application copy less ambiguous

ammongit:
The current copy is too casual and introduces unnecessary uncertainty.


ammongit:
https://scuttle.atlassian.net/browse/WDBUGS-5

PR #9, opened 2020-06-07
Allow [[/div_]] to end blocks

ammongit:
Since `[[div_]]` is a modified wikidot div that is syntactically permissible, `[[/div_]]` should be permitted as well.

### DIFF
--- a/lib/Text_Wiki/Text/Wiki/Parse/Default/Div.php
+++ b/lib/Text_Wiki/Text/Wiki/Parse/Default/Div.php
@@ -1,53 +1,53 @@
 <?php
 
 /**
- * 
- * Parses divs.
- * 
- * @category Text
- * 
- * @package Text_Wiki
- * 
- * @author Michal Frackowiak
- * 
- * @license LGPL
- * 
- * @version $Id$
- * 
- */
-
-/**
- * 
+ *
  * Parses divs.
  *
  * @category Text
- * 
+ *
  * @package Text_Wiki
- * 
+ *
  * @author Michal Frackowiak
- * 
+ *
+ * @license LGPL
+ *
+ * @version $Id$
+ *
+ */
+
+/**
+ *
+ * Parses divs.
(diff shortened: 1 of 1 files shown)


ammongit:
https://scuttle.atlassian.net/browse/WDBUGS-35

pxdnbluesoul:
I haven't actually used or seen [[/div_]] used, what is the intent of the PR?

ammongit:
The intent is to allow a syntactic construction that should apparently work but does not.

Wikidot allows the use of `[[div_]]`, which is like a wikidot div except it doesn't pad the internals. That is:
* `[[div]]` -> `<div><p> ... </p></div>`
* `[[div_]]` -> `<div> ... </div>`

However the ending tag does not have this flexibility. Thus:
```
[[div_]]
contents1
[[/div]]
----
[[div_]]
contents2
[[/div_]]
```
appears as:
image

rossjrw:
I feel like if the "do not pad interior paragraphs" setting were an attribute on the div tag rather than a modification to its name, this PR wouldn't be desirable behaviour. For example if the proper way to produce this were `[[div pad-interior="false"]]`, then `[[/div]]` would clearly be the proper closing tag. If you view `[[div_]]` not as a separate tag altogether but as an alias for `[[div pad-interior="false"]]`. Much the same as how you'd never expect to add a div's classes to its closing tag.

Wanting `[[/div_]]` feels to me like the original poor design choice of implementing this feature as `[[div_]]` propagating forward. I say keep the closing tag clean.

ammongit:
That's a valid point. (I personally feel like `pad-interior` is kind of long for how useful of a behavior it is, but this is a nitpick)

I appreciate the focus on considering how to reform the syntax as we go forward to improve clarity and usability.

ammongit:
Based on this I have filed https://scuttle.atlassian.net/browse/WDBUGS-95 and will be closing this PR

PR #16, opened 2020-06-12
Use codesniffer to autoformat PHP files

ammongit:
Decided to not use Prettier as discussed on the issue. 

Of the autoformatting options I tried, this was the most comprehensive and reasonable. There are still some unusual outputs due to Wikidot source being `;-)` but overall it's a big improvement.


ammongit:
https://scuttle.atlassian.net/browse/WDBUGS-76

PR #28, opened 2020-07-04
Enable HTTPS and SameSite compatibility.

pxdnbluesoul:
This enables the ability for the server administrator to require HTTPS for all traffic, and runs cookie requests through a new method that enables the SameSite=None flag, allowing for auth on custom domains. This also contains a few bugfixes and removes some sources of benign warnings.


templates/modules/managesite/ManageSiteSecureAccessModule.tpl
```diff
@@ -42,3 +43,4 @@
 <p style="font-size: 120%; font-weight: bold;">
 	In case of troubles - please contact <a href="mailto:{$SUPPORT_EMAIL}">{$SUPPORT_EMAIL}</a>.
 </p>
+{/if}
```
ammongit, line 46: Trailing newline

templates/modules/managesite/ManageSiteSecureAccessModule.tpl
```diff
@@ -28,7 +29,7 @@
 </form>
 
 <p>
-	The secure mode (SSL/TSL via HTTPS) basicly mean that the whole
+	The secure mode (SSL/TSL via HTTPS) basically means that the whole
```
ammongit, line 32: Suggested copy change:
```suggestion
	The secure mode (SSL/TSL via HTTPS) means that the whole
```

web/index.php
```diff
@@ -31,7 +31,11 @@
 
 try {
     // set anti-session-riding token
-    setcookie("wikidot_token7", md5(rand(0, 10000)), null, '/');
+    setsecurecookie("wikidot_token7", md5(rand(0, 10000)), 0, '/', GlobalProperties::$SESSION_COOKIE_DOMAIN);
```
ammongit, line 34: Is `rand()` a CSPRNG? Also I feel this would be cleaner if the cookie value was created in a separate line.

php/actions/Login2Action.php
```diff
@@ -99,8 +99,8 @@ public function loginEvent($runData)
             $runData->ajaxResponseAdd('originalUrl', $originalUrl);
         }
 
-            setcookie("welcome", $user->getUserId(), time() + 10000000, "/", GlobalProperties::$SESSION_COOKIE_DOMAIN);
-            setcookie(GlobalProperties::$SESSION_COOKIE_NAME_IE, $runData->getSessionId(), null, "/");
+            setsecurecookie("welcome", $user->getUserId(), time() + 10000000, "/", GlobalProperties::$SESSION_COOKIE_DOMAIN);
```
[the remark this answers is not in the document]
pxdnbluesoul: That appears to be the default lifespan for the session cookie, about 2.5 months, refreshed through one of the flow controllers.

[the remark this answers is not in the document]
ammongit: I see. Eventually we should split this out into a config setting and avoid inlining the calculation

php/db/base/DB/SiteSettingsBase.php
```diff
@@ -138,7 +138,17 @@ public function setMaxPrivateViewers($v1, $raw = false)
...
+         * We are deprecating any usage of insecure HTTP.
+         * The `allow_http` flag in wikidot.ini should only be set to true for local dev purposes.
+         * If you do so, SSL is controlled on a per-site basis in the site_settings table.
+         * The only difference between `ssl_only` and `ssl_only_paranoid` is the latter forces a secure cookie.
+         * This may be removed entirely in a future release.
+         */
+        if(\GlobalProperties::$ALLOW_ANY_HTTP == true) {
+            return $this->getFieldValue('ssl_mode');
+        }
+        return "ssl_only_paranoid";
```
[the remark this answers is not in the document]
pxdnbluesoul: Yeah this will likely get overhauled down the road, there are a few places in the code that are expecting to receive a string, and there are very few differences between a couple of them such that this could end up probably being distilled down to a bool.

web/index.php
pxdnbluesoul: If this were for any other purpose I would agree with you, but the token7 is an arcane bit of crap that hopefully we can live without or at least better understand the reason for here soon, and we can either enhance it or remove it altogether.

PR #29, opened 2020-07-07
Rename project to Wikijump

pxdnbluesoul:


rossjrw:
I notice that you've stripped copyright headers from a lot of files - I presume that's just to make it easier to add corrected ones later on?

PR #30, opened 2020-07-07
Decrapify master

pxdnbluesoul:
This commit removes a bunch of stuff not actually used in day-to-day usage. The setup method doesn't work on a modern OS, and a new method will be created later using the wikidot-infra repo and perhaps eventually moved into an installer package.


files/dump/sites/profiles/start.page
```diff
@@ -1,25 +0,0 @@
...
-
-Go to [[[admin:users | Users management panel]]] and create accounts for wiki users.
-
-++ Visit Wikidot.org
-
-Go to **[http://www.wikidot.org www.wikidot.org]** -- home of the Wikidot software -- for extra documentation, howtos, tips and support.
-
-++ Explore more at Wikidot.com
-
-If you want more feature rich wiki reliably hosted, consider creating a wiki within **[http://www.wikidot.com/ Wikidot.com]**
```
ammongit, line 25: Do these files already exist in another location? What gets populated for new wikis?

tests/OzoneTest.php
```diff
@@ -0,0 +1,92 @@
...
+        $ncat->setNew ( true ) ;
+        $ncat->setCategoryId ( null ) ;
+        $ncat->setName ( "auth" ) ;
+        $ncat->save () ;
+
+        addPage ( $ncat, "test", "Test page.", "Test page" ) ;
+
+        $od->recompileWholeSite ( DB_SitePeer::instance ()->selectByPrimaryKey ( 1 ) ) ;
+
+//        $db->commit () ;
```
ammongit, line 89: I'm guessing these were commented in the original? Do we know anything about whether these used to be required, conditionally needed, etc?

files/dump/sites/profiles/start.page
pxdnbluesoul: These are basically just dumps of stuff that's already in the database. If we want to bootstrap another way of doing the setup down the road that's fine, but I don't want to keep this stuff around for now.

tests/OzoneTest.php
pxdnbluesoul: This didn't actually exist as a test before, I stole this from one of the update scripts as it looks to be the workflow for saving a page. For now it's not actually ready for phpunit, but I wanted to keep the logic somewhere and it will probably be a setup method for much of our test suite to come.

PR #33, opened 2020-07-12
Halfway

ealevine:
Alright so this time with feeling, apologies for the issues.

-Replaced "can not" with "can"

### DIFF
--- a/php/actions/MembershipApplyAction.php
+++ b/php/actions/MembershipApplyAction.php
@@ -53,7 +53,7 @@ class MembershipApplyAction extends SmartyAction
 
         $user = $runData->getUser();
         if ($user == null) {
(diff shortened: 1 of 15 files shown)


php/actions/MembershipApplyAction.php
```diff
@@ -122,7 +122,7 @@ public function applyByPasswordEvent($runData)
 
         $user = $runData->getUser();
         if ($user == null) {
-            throw new ProcessException(_("Sorry, you are not logged in. Anonymous users can not apply ;-)"));
+            throw new ProcessException(_("Sorry, you are not logged in. Anonymous users cannot apply ;-)"));
```
ammongit, line 125: I would suggest removing the `;-)` both here and above

PR #34, opened 2020-10-15
[Web] Convert to TypeScript + Webpack

rossjrw:
The intent of this PR is to start preparing the codebase for a Javascript refactor. Eventually, I'd like everything to be Typescript with Babel. A build process will definitely be needed, and I'm thinking the best approach to that will be to use Webpack to generate a single bundle.js with everything the site needs - no more importing eight different JS files at the top of every layout template.

Right now the codebase is structured to have Javascript files interpreted literally, which is not sustainable nor desirable in the world of modern Javascript. A structural rework will be required before we start refactoring anything, and that's what this PR is for.

Eventually I would like to dump the included libraries, especially Yahoo UI which is mostly just used to add DOM event listeners (which is now built-in and has been since IE 9), but that is out of scope for this PR.

By the end of this PR, Wikijump will not be type-safe. That won't happen until we ditch old libraries. After that, hopefully, yes.

Noting that eventually modules CSS should also be bundled.

- [x] Add record of removed functions
- [x] Type OZONE.js
- [x] Type OZONE.dialog.js
- [x] Rearrange OZONE JS for type-safe imports and exports
- [x] Type Wikijump.js
- [x] Type Wikijump.editor.js
- [x] Type Wikijump.page.js
- [x] Rearrange Wikijump JS for type-safe imports and exports

Note to self: avoid refactoring where not absolutely necessary while typing!

- [x] Set up bundling and transpiling
- [x] Change templates to point to the bundle
- [x] Add build instructions
- [x] Change dockerfile/install script to install node, npm and npm packages, and run the build script

Tasks originally intended to be completed in this PR, but now scheduled for a later one:

* Type modules JS
* Change PHP module handlers to not fetch module JS (it will be bundled)
* Change JS module handlers to not fetch module JS (`OZONE.ajax.requestModule`)


rossjrw:
The main JS files are kept in `web/files--common/javascript`. These files are requested by the page templates in `templates/layouts`.

`combined.js` is the result of `yahooui/combine.sh` which concatenates some (but not all) of the Yahoo UI and Moo FX files into one. `combined.js` is used by the WikiLayout template. Other layout templates - PrintLayout and Default - use subsets of the scripts contained in `combined.js` and import them individually.

Every module has its own CSS and JS in `web/files--common/modules/[js|css]`. These attach a listener to the main Wikijump object, which when activated triggers a callback. The listener is attached to part of the page using the YAHOO object. I'm not sure where the modules JS/CSS is called from, yet. I don't think it's Wikijump.js.

rossjrw:
The way I've had Webpack configured for other projects is to bundle everything - JS, CSS, TOML, SVG, whatever - into a single bundle, slap an anti-cache hash on it, and then affix that to a basic index.html with `HtmlWebpackPlugin`. That's no longer appropriate - a simple HTML template will no longer suffice, Wikijump uses Smarty templates, which are kept elsewhere and do not have a src/dist version.

One solution would be to simple output a bundle.js with no hash, import that into each template, and hope that it's fine. It should be, but it might cause caching issues. Updating the JS and refreshing local wikijump.test so far has been fine, though, so it's probably fine for now. Might be something worth implementing before we hit production, though.

Another problem is unbundled files. Under the current system, modules JS are explicitly excluded by default and called only when the module is required by a template. I see no reason to fuck with this system right now, it's working fine, but it does mean that those scripts have to stay in their unique locations and can't be bundled along with everything else. It's a stupid system, really, I'd much rather take the JS framework route and attach everything to where it belongs instead of asking for each script specifically, but it does work. I need to work out how to configure Webpack to not bundle those files and instead compile them in-place (or to a parallel of that place, at least).

rossjrw:
The core JS weighs 2.2M. The modules JS weighs 720K in total. That's... that's nothing.

I'm not sure that fetching modules JS individually when they're needed is even a good idea, and not something to continue, even if it works. It can all be bundled.

ammongit:
> But this is fine for now. There's nothing to do except to register WIKIREQUEST as an external.
Seems worthwhile to consider whether later on we want to change what actually interfaces with the client JS 

> The big question is: where do I want to keep compiled Javascript files as opposed to their source Typescript files? It's typical to differentiate them with dist/src directories, but of course, right now there's no such distinction.
This is a necessity as part of the build process changes, imo. All existing JS files are moved as sources, and their "compilation" is simple copying. TS files are transpiled appropriately and the built JS is placed in dist, which is served in the bundle/to the client

> The way I've had Webpack configured for other projects is to bundle everything - JS, CSS, TOML, SVG, whatever - into a single bundle, slap an anti-cache hash on it, and then affix that to a basic index.html with HtmlWebpackPlugin
"anti-cache hash"? as in the filename has the sha256 or whatever to avoid cache issues? 
for now let's just have the single bundle file, we can worry about the naming stuff when we are refactoring smarty stuff

web/files--common/javascript/OZONE.js
```diff
@@ -1,767 +1,730 @@
...
+    setInnerHTMLContent: function (elementId, content) {
+      var el = $(elementId)
+      if (el) {
+        el.innerHTML = content
+        OZONE.utils.formatDates(el)
+        OZONE.dialog.hovertip.dominit(el)
+      }
+    },
+
+    disableEnterKey: function (e) {
```
ammongit, line 287: wtf is this function used for

web/files--common/javascript/OZONE.js
```diff
@@ -1,767 +1,730 @@
...
+      if (text === null || text === '') {
+        return ''
+      }
+      return text.split('&').join('&amp;').split('<').join('&lt;').split('>').join('&gt;')
+    },
+    unescapeHtml: function (text) {
+      if (text === null || text === '') {
+        return ''
+      }
+      return text.split('&gt;').join('>').split('&lt;').join('<').split('&amp;').join('&')
```
ammongit, line 314: oh my god lol

.eslintrc.json
```diff
@@ -0,0 +1,27 @@
...
+    "@typescript-eslint"
+  ],
+  "rules": {
+    "@typescript-eslint/no-unused-vars": [
+      "warn", { "argsIgnorePattern": "^_" }
+    ],
+    "@typescript-eslint/no-non-null-assertion": "off",
+    "quotes": "off"
+  }
+}
```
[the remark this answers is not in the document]
rossjrw: That's fine, I'll set up my editor to do that

package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "wikijump",
+  "main": "web/common--files/index.js",
+  "directories": {
+    "test": "tests"
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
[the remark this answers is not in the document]
rossjrw: It's the default from `npm init` lmao

web/files--common/javascript/OZONE.js
```diff
@@ -1,767 +1,730 @@
...
+    _callbackArrayIndex: 0,
+    _javascriptLoadLock: false,
+
+    /**
+     * arg - extra parameter passed to the callback as a second parameter
+     */
+    requestModule: function (moduleName, parameters, callback, arg, options) {
+      OZONE.visuals.cursorWait()
+      if (parameters === null) {
+        parameters = {}
```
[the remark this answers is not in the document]
rossjrw: This is something we need to discuss, then. I'm using [JavaScript Standard Style] (https://standardjs.com/), which is a linter and fixer that enforces a 'standard' style and crucially does not have any configuration. The point of it is to prevent wasting time on this sort of discussion. 

However, standard doesn't work with Typescript out of the box and its own FAQ recommends you replace it with [standardx](https://github.com/standard/standardx), a fork of it that also enables it to be configured. So while it's against the ideology of standard, it is possible to enable semicolons. 

It's very rare that a semicolon in JS is syntactically necessary. I'd strongly prefer to go without - if we're writing code that needs semicolons in order not to be ambiguous, whether that's syntactically or visually, we're writing bad code. 

Equally, though, I wouldn't merge a commit that removes semicolons without at least looking through the code first.

rossjrw:
Noting that there are some WIKIREQUEST specifics that I'll need to resolve. Currently WIKIREQUEST is loaded into the template _after_ OZONE.js but _before_ Wikijump.js. That's because at the same time as the WIKIREQUEST object is constructed, a couple of flags are set on the OZONE object (current date and site language), which requires OZONE to already be loaded. But it needs to come before Wikijump.js because the information in the object is used for doing stuff.

The bundle contains both OZONE and Wikijump.js, so now WIKIREQUEST has to be defined either before or after both of them.

What I'll have to do is make sure WIKIREQUEST is defined before the bundle has been loaded, and then either define OZONE properties in a separate script afterwards, or have those properties be defined from WIKIREQUEST. The latter makes more sense, as OZONE is supposed to be an implementation-agnostic framework, I think. I will probably end up deprecating `OZONE.lang` in favour of `WIKIREQUEST.lang`.

rossjrw:
There are 135 modules in the namespace `Wikijump.modules`. This is fine and, for now, type-safe, as I can just (for now) define `Wikijump.modules` as being an object with an unknown number of string keys.

There are 4 modules attached directly to the `Wikijump` object. This is bad, and disrupts the type safety of the Wikijump object. They should be moved into `Wikijump.modules`.

Module | Notes
--- | ---
`CreateAccountModule` | So far as I can tell this is unused in favour of `Wikijump.modules.CreateAccountModule` which is in either the `createaccount` or the `createaccount2` dir - unsure yet. It seems to refer only to itself, and `Wikijump.page.createAccount` contains a commented-out reference to it. There is also no equivalent module on the PHP side. I think it is deprecated.
`CreateSiteModule` | 
`wikijumpadmin/ManageSuperUserModule` |
`wikijumpadmin/ManageUsersModule` |

rossjrw:
I deadcode'd `OZONE.ajax.requestQuickModule` because it was unused, but it turns out that modules do actually request quickmodules - just not with that method.

How `requestQuickModule` makes a request:
```javascript
YAHOO.util.Connect.asyncRequest('POST','/quickmodule.php?module='+moduleName,internalCallback,postdata);
```

How modules actually request quickmodules:
```javascript
var myDataSource = new YAHOO.widget.DS_XHR("/quickmodule.php", ['pages', 'unix_name', 'title']);
myDataSource.scriptQueryParam="q";
myDataSource.scriptQueryAppend = "s="+WIKIREQUEST.info.siteId+"&module=PageLookupQModule";
var myAutoComp = new YAHOO.widget.AutoComplete("sm-nav-top-bar","sm-nav-top-bar-list", myDataSource);
```

So far as I can tell it looks like quickmodules are used to populate the crappy 'autocomplete' fields that fill in the name of a message's recipient, or the name of page when setting a parent. Without having exhaustively checked it does look like they're only used for autocomplete. There are only three quickmodules: `MemberLookupQModule`, `PageLookupQModule`, and `UserLookupQModule`. I think I'd like to change the name of this concept eventually, assuming that we don't scrap it in favour of something else altogether.

rossjrw:
This PR is in a stable state now so I'm marking it ready for review.

Tasks to be completed in a future PR:
* Modules typing: modules are still Javascript right now.
* Modules bundling: module JS is still in a bunch of separate files, and not bundled. This is stable and how Wikidot has always worked, so it's fine for now.
* Modules testing: things seem to work at a surface level, but there's no way to tell what's broken, what didn't pass through the typing process unscathed, etc. This is the same situation we were in before, though, so I don't consider this a great loss.

web/web/files--common/javascript/OZONE/dialog.ts
```diff
@@ -0,0 +1,675 @@
...
+// attached.
+type HovertipOptions = {
+  context?: string
+  text?: string
+  style?: CSS.Properties
+  noCursorHelp?: boolean
+  delay?: number
+  smartWidthLimit?: number
+  valign?: 'center'
+}
```
ammongit, line 26: Related to the point about `interface`, shouldn't we end each of these in `;`?

web/web/files--common/javascript/OZONE/dialog.ts
```diff
@@ -0,0 +1,675 @@
...
+      const el = document.getElementById(options.context);
+      if (el === null) {
+        // XXX This should probably throw an error
+        return;
+      }
+      tipEl = el;
+    } else if (options && options.text) {
+      // create a new div
+      tipEl = document.createElement('div');
+      tipEl.innerHTML = `<div class="content">${options.text}</div>`;
```
ammongit, line 408: Add `TODO` for potential HTML injection issue

web/web/files--common/javascript/OZONE/dialog.ts
```diff
@@ -0,0 +1,675 @@
...
+            el.attributes[x].nodeValue = '';
+          }
+        }
+      }
+      if (title === null) {
+        // XXX Should probably throw an error
+        return;
+      }
+      tipEl = document.createElement('div');
+      tipEl.innerHTML = `<div class="content">${title}</div>`;
```
ammongit, line 428: Ditto

web/web/files--common/javascript/OZONE/dialog.ts
```diff
@@ -0,0 +1,675 @@
...
+    // fix if not "content" div inside.
+    const subDivs = tipEl.getElementsByTagName('div');
+    let hasContent = false;
+    for (let i = 0; i < subDivs.length; i++) {
+      if (YAHOO.util.Dom.hasClass(subDivs[i], 'content')) {
+        hasContent = true;
+      }
+    }
+    if (!hasContent) {
+      tipEl.innerHTML = `<div class="content">${tipEl.innerHTML}</div>`;
```
ammongit, line 448: Ditto

web/web/files--common/javascript/OZONE/utils.ts
```diff
@@ -0,0 +1,481 @@
...
+          }
+        }
+      }
+      if (dateString) {
+        dateElement.innerHTML = dateString;
+        dateElement.style.visibility = 'visible';
+        dateElement.style.display = 'inline';
+      }
+    });
+  },
```
ammongit, line 385: A lot of this stuff should either be in some shared TS library or in the backend

web/web/files--common/javascript/Wikijump/render.ts
```diff
@@ -0,0 +1,37 @@
...
+    const link = compress`
+      href="javascript:;"
+      onclick="Wikijump.page.listeners.userInfo(${userId})"
+    `;
+    return compress`
+      <span class="printuser">
+      ${showAvatar
+        ? `<a ${link}>
+          <img class="small"
+               src="${AVATARS}/${Math.floor(userId / 1000)}/${userId}/a16.png"
```
ammongit, line 29: This is logic being necessary why I'm in favor of just `/avatars/[user-id]/[size].png`

ammongit (approved):

.gitignore
```diff
@@ -1,12 +1,12 @@
...
-.idea/
-.*.swp
-
-# Miscellaneous
-*.bak
-*~
+# Keys and secrets
+files/public.pem
+files/modulus.pem
+*.pem
```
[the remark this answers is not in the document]
rossjrw: That's a commit for master

web/DEADCODE.md
```diff
@@ -17,3 +17,79 @@ If possible, add to this log in the same commit in which the code is removed.
 * Where it was: [UnixifyString](https://github.com/scpwiki/wikijump/blob/571cd42cb810223a5dfe5f594b66adc39cb5295e/web/php/utils/Wikijump/Util/UnixifyString.php), [Unixify](https://github.com/scpwiki/wikijump/blob/571cd42cb810223a5dfe5f594b66adc39cb5295e/web/php/quickmodules/Unixify.php)
 * What it did: It duplicated the code found in `WDStringUtils`, with some minor changes.
 * Why it was removed: It was unused.
+
+## JS: `OZONE.visuals.highlightText`
```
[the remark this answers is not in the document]
rossjrw: There's no issue that's specific to `highlightText`, other than WJ-218 which covers the whole list, and I'm loathe to create a new issue for it when it's already resolved

web/web/files--common/javascript/OZONE/dialog.ts
rossjrw: With ESLint set to be uncompromising on semicolons, it doesn't complain about this. The single-line form of an object type declaration is delimited by commas, but ESLint would normally complain about that too and it doesn't here. Maybe because this never makes it to Javascript?

web/web/files--common/javascript/OZONE/dialog.ts
rossjrw: Created [WJ-248](https://scuttle.atlassian.net/browse/WJ-248)

web/bin/utilities/jsts.sh
```diff
@@ -0,0 +1,3 @@
+standardx --fix $1
+(printf "import Wikijump from \"@/javascript/Wikijump\";\nimport OZONE from \"@/javascript/OZONE\";\ndeclare const YAHOO: any;\ndeclare const fx: any;\n"; cat $1) > ${1%.js}.ts
+nvim -O ${1%.js}.ts $1
```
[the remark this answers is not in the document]
rossjrw: It only has to work for as long as it takes to convert modules to TS, and then it'll be gone, so I'm not too concerned about making it perfect. Honestly this script probably shouldn't be in version control at all.

web/web/files--common/javascript/OZONE/ajax.ts
```diff
@@ -0,0 +1,141 @@
...
+      return;
+    }
+    parameters.wikijump_token7 = token;
+
+    YAHOO.util.Connect.asyncRequest(
+      'POST',
+      '/ajax-module-connector.php',
+      ajax.requestModuleCallback,
+      OZONE.utils.arrayToPostData(parameters)
+    );
```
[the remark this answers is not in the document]
rossjrw: Actually I don't know why I left it as `unknown` - it's either `void` or `Promise<void>`. As there's no return value I'll change it to `void`

web/web/files--common/javascript/Wikijump/render.ts
rossjrw: Created [WJ-249](https://scuttle.atlassian.net/browse/WJ-249)

web/DEADCODE.md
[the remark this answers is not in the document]
ammongit: Oh, I just mean adding 
```suggestion
## JS: `OZONE.visuals.highlightText`
* Relevant Issues: [WJ-218](https://scuttle.atlassian.net/browse/WJ-218)
```
to each of these sections. I'm not saying you need to create separate issues for each, just that we record which issue these changes came as part of for easy tracking.

web/bin/utilities/jsts.sh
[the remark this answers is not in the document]
ammongit: Okay. I would leave a comment to that effect somewhere, even if it's just `# TODO: remove when all modules are converted`

ammongit (approved):

web/DEADCODE.md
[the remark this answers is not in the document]
rossjrw: I've added a link to 218 at the top of DEADCODE, but I don't think it's necessary to add to each entry

ammongit:
Eh, I'd prefer to be more explicit, but I'm not dying on this hill.

ammongit:
https://scuttle.atlassian.net/browse/WJ-250

PR #37, opened 2020-10-27
Remove old log files

ammongit:
These shouldn't be in the source, and haven't been touched for years anyways.


rossjrw:
logs/jobs.log is used by bin/job.php and has logs from this April - are we sure it's unused? Do we know what bin/job.php does?

ammongit:
If we keep the directory it should be fully `.gitignore`'d, logs shouldn't be in the repo still. It's runtime stuff that's not relevant to some arbitrary person running the code.

rossjrw (approved):
Should this be in the main gitignore

ammongit:
thanks @rossjrw

PR #38, opened 2020-11-06
Add new forbidden regular expressions in default configuration

ammongit:
https://scuttle.atlassian.net/browse/WDBUGS-112


conf/forbidden_user_names.conf
```diff
@@ -5,17 +5,31 @@
...
 /^root$/
+/^error$/
+/^null$/
+/^undefined$/
 /^bot$/
-/^support$/
\ No newline at end of file
+/^robot$/
+/^O5-\w+$/
+/^SCP-\w+$/
```
pxdnbluesoul, line 35: Do we want to extend sites the ability to disallow certain nicknames from joining? We could also check renames against site specific configurations at that point as well.

ammongit: I think that makes sense, but would be a separate issue outside the scope of this PR. I was thinking these make sense since, our main Wikijump installation would be for EN, official sandboxes, etc.

pxdnbluesoul: Alright, I don't want this to block this pull, I think when we've got a bit more new code written we might circle back on this and see if we can pull these site-specific requirements out of the globals.

PR #39, opened 2020-11-07
Remove the "simple todo" module and associated code

ammongit:
As discussed, this feature isn't really used anywhere, is poorly designed, and isn't worth keeping around given the necessity of code maintenance. This PR removes it.

https://scuttle.atlassian.net/browse/WDBUGS-219


rossjrw (approved):
This seems fine to me. Confirming that the phrase "simpletodo" no longer exists in the repo. Sadly "todo" does so we've know way of knowing if RegularToDo or ComplicatedToDo ever existed

ammongit:
thanks @rossjrw

PR #42, opened 2020-11-11
[Web] Javascript testing

rossjrw:
This PR sets the groundwork for JS tests.

- [x] Use Jest for testing
- [x] Configure Jest to play nicely with Webpack
- [x] Configure Jest to play nicely with Typescript (via Babel)
- [x] Add a GitHub Action to test JS automatically on push to master, or on any PR
- [x] Split the tests directory into PHP and JS
- [x] Write a first test, to make sure everything is working

Noting that this PR is made against `javascript-init` (#34) which is pre-monorepo. During the merge of `javascript-init` into `master`, most changes will affect the `web/` subrepo, but the Action workflow from this PR will need to be moved into the root `.github/`. There is a commented-out `paths` filter in the workflow to restrict the test to PRs that affect the `web/` subrepo, which should be uncommented during the merge of #34.


.github/workflows/test-js.yml
```diff
@@ -0,0 +1,21 @@
+name: JS tests
```
ammongit, line 1: Would prefer a more obvious name, like "Frontend Javascript tests"

.github/workflows/test-js.yml
```diff
@@ -0,0 +1,21 @@
...
+on:
+  # Run this on push to master
+  push:
+    branches:
+      - master
+  # Run this on push to PRs, regardless of branch
+  pull_request:
+  # [Monorepo] Only run this if there are changes to web
+  # paths:
+  #   - "web/**"
```
ammongit, line 11: This is commented out, the `pull_request:` object is empty. Is this valid YAML?

package.json
```diff
@@ -6,21 +6,39 @@
...
+    "babel": {
+        "presets": [
+            [ "@babel/preset-env", { "targets": {"node": "current"} } ],
+            "@babel/preset-typescript"
+        ]
+    },
+    "jest": {
+        "moduleFileExtensions": ["js", "ts"],
+        "moduleNameMapper": {
+            "^@/(.*)$": "<rootDir>/web/files--common/$1"
```
ammongit, line 25: Curious, what does this do?

ammongit (approved):

.github/workflows/test-js.yml
rossjrw: My reason for making the name shorter is actually for the possibility of future README badges - I think it looks better to have a shorter, lowercase-oriented name

JS tests

ammongit: Hmm, okay

.github/workflows/test-js.yml
rossjrw: It's valid: ["You must append a colon (:) to all events, including events without configuration."](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#on)

Commented out section is explained in the OP

package.json
rossjrw: Throughout the typescript I'm using `@` as an alias for `./web/files--common`, so that I can do `import whatever from "@/javascript/thing"` (and eventually `"@/modules/thing"`)

https://github.com/scpwiki/wikijump/blob/197ce822449b94b0428fb428615ceda42d045953/webpack.config.js#L23-L24

https://github.com/scpwiki/wikijump/blob/197ce822449b94b0428fb428615ceda42d045953/tsconfig.json#L19-L20

That just sets it up so that Jest understands that alias, too, but for Jest it's a bit lower-level and needs a regex replacement.

.github/workflows/test-js.yml
rossjrw: For reference, when that section is uncommented, the indentation of that snippet would look like this:
```
pull_request:
paths:
  - "web/**"
```

package.json
ammongit: I would consider changing it to `@web` in case we need to add other special paths in the future. But this is fine.

PR #45, opened 2020-11-12
Convert to UNIX newlines

ammongit:
These files previously used CRLF. Note that the added/removed lines in the PR are identical.


ammongit:
Approved by @pxdnbluesoul in chat

PR #47, opened 2020-11-13
[Installer] Update Dockerfile and installer script for monorepo structure

rossjrw:
The Dockerfile is set up for the old web-only repository structure, not the new monorepo structure (#41). This PR intends to update the Dockerfile and install.sh such that they work for the new structure.

- [x] Update Dockerfile
- [x] Update install.sh
- [ ] Get Docker install to work

Trying to diagnose what's happening inside the Docker container:
* `localhost` returns nginx 404
* `https://www.wikijump.test` returns nothing
* `https://wikijump.test` returns nothing
* `http://www.wikijump.test` returns nothing with `allow_http` set to `true`
* `http://wikijump.test` returns nothing
-----
* Services (postgresql, php7.4-fpm, memcached, nginx) were not started inside Docker, but starting them manually doesn't change anything
* Local versions of those services were stopped outside of the container, so they shouldn't be conflicting


installer/Dockerfile
```diff
@@ -87,7 +87,6 @@ RUN sed -i "s/%PHP%/${PHP}/g" /app/services.sh
 WORKDIR /var/www
 
 RUN git clone "${WIKIJUMP_REPO}" "${WIKIJUMP_REPO_DIR}"
-RUN chmod -R 777 "${WIKIJUMP_REPO_DIR}/tmp/smarty_templates_c"
```
ammongit, line 90: I'm pretty sure bluesoul said that Wikidot will refuse to run if `tmp/smarty_templates_c` doesn't have permissions of `777`

installer/Dockerfile
```diff
@@ -120,15 +119,15 @@ RUN scripts/setup-postgres.sh
...
-RUN cp -av "./var/www/wikijump/conf/wikijump.ini" "${WIKIJUMP_DIR}/conf/wikijump.ini"
+RUN cp -av "./var/www/wikijump/conf/wikijump.ini" "${WIKIJUMP_DIR}/web/conf/wikijump.ini"
 RUN cp -av "./etc/php/${PHP_VERSION}/cgi/conf.d/"* "/etc/php/${PHP_VERSION}/cgi"
 RUN cp -av "./etc/php/${PHP_VERSION}/cgi/conf.d/"* "/etc/php/${PHP_VERSION}/fpm"
 
 # Inject values in the wikijump.ini configuration file
 
-RUN sed -i "s/BASEDOMAIN/${MAIN_DOMAIN}/g" "${WIKIJUMP_DIR}/conf/wikijump.ini"
-RUN sed -i "s/MAINWIKI/${WWW_DOMAIN}/g" "${WIKIJUMP_DIR}/conf/wikijump.ini"
-RUN sed -i "s/FILEDOMAIN/${FILES_DOMAIN}/g" "${WIKIJUMP_DIR}/conf/wikijump.ini"
```
ammongit, line 131: Instead of `${WIKIJUMP_DIR}/web`, we can have `${REPO_DIR}` be the root path of the repository, and `${WIKIJUMP_DIR}` be `${REPO_DIR}/web`

rossjrw:
I've had no luck running dockerd and connecting to it from WSL2, so this is my installation process from windows:
* Install docker on WSL
* Install docker desktop on windows https://docs.docker.com/docker-for-windows/install-windows-home
* In WSL change `credsStore` to `credStore` in `~/.docker/config.json` https://forums.docker.com/t/docker-credential-desktop-exe-executable-file-not-found-in-path-using-wsl2/100225/4
* Edit the config file as needed
* Build the image: `docker build . -t scpwiki/wikijump:local --no-cache`
* In docker desktop, run the image, with 443:443 and 80:80 set in optional settings
* Click "open in browser" on the container or navigate to localhost

image

Next job is getting it to show wikijump

installer/Dockerfile
rossjrw: Turns out that is correct; fixed in 34eb5643c11b591c304bd4fe0b28198c26ae988b

pxdnbluesoul (approved):

ammongit (approved):

PR #49, opened 2020-11-15
[WJ-243] [Docs] More explicit local development documentation

rossjrw:
This PR adds more explicit documentation to development.md, aimed at future contributors like me who aren't too familiar with the tools we use, how they work, or how to use them.

I welcome commits that add more detail, remove superfluous detail, clean up wording, are technical corrections, or anything else.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -43,6 +43,8 @@ The recommended way to install Wikijump is via Docker. Docker is a container sys
 
 You will need Docker installed and running:
 
+https://github.com/rossjrw/dotfiles/blob/3c5445abb138b735cc3caf61f070c9125fa87d2f/.profile#L28
+
 <table>
   <thead><tr><th>Ubuntu / Ubuntu VM</th><th>Windows via WSL2</th></tr></thead>
   <tbody valign="top"><tr>
@@ -54,7 +56,7 @@ You will need Docker installed and running:
     <td>
       <p><pre># apt install docker</pre></p>
       <p>Install <a href="https://docs.docker.com/docker-for-windows/install-windows-home">Docker Desktop</a>, which you'll be using as the Docker daemon, and leave it running in the background.</p>
-      <p>If at any point you hit an error along the lines of "<code>"docker-credential-desktop.exe": executable file not found in $PATH</code>", edit <code>~/.docker/config.json</code> and change "credsStore" to "credStore".</p>
+      <p>If you have set <code>appendWindowsPath=false</code> in your WSL config, then you may hit an error along the lines of <code>"docker-credential-desktop.exe": executable file not found in $PATH</code>. In this case you should either add <code>/mnt/c/Program\ Files/Docker/Docker/resources/bin</code> to your PATH, or <a href="https://github.com/rossjrw/dotfiles/blob/3c5445abb138b735cc3caf61f070c9125fa87d2f/.profile#L28">create a symlink</a> in a directory that is in your PATH already.</p>
     </td>
   </tr></tbody>
 </table>
--- a/docs/development.md
+++ b/docs/development.md
@@ -43,8 +43,6 @@ The recommended way to install Wikijump is via Docker. Docker is a container sys
 
 You will need Docker installed and running:
 
-https://github.com/rossjrw/dotfiles/blob/3c5445abb138b735cc3caf61f070c9125fa87d2f/.profile#L28
-
 <table>
   <thead><tr><th>Ubuntu / Ubuntu VM</th><th>Windows via WSL2</th></tr></thead>
   <tbody valign="top"><tr>
--- a/docs/development.md
+++ b/docs/development.md
@@ -54,7 +54,7 @@ You will need Docker installed and running:
     <td>
       <p><pre># apt install docker</pre></p>
       <p>Install <a href="https://docs.docker.com/docker-for-windows/install-windows-home">Docker Desktop</a>, which you'll be using as the Docker daemon, and leave it running in the background.</p>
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -1,32 +1,136 @@
...
+
+```
+127.0.0.1 localhost
+127.0.0.1 wikijump.test
+127.0.0.1 www.wikijump.test
+127.0.0.1 profiles.wikijump.test
+127.0.0.1 template-en.wikijump.test
+127.0.0.1 wjfiles.test
+127.0.0.1 www.wjfiles.test
+```
```
ammongit, line 32: Would also add equivalents for `::1` ([IPv6](https://scuttle.atlassian.net/browse/WJ-172))

Also noting the order / formatting I use:
```
# Development
127.0.0.1                   wikijump.test               Wikijump
127.0.0.1                   www.wikijump.test           Wikijump
127.0.0.1                   wjfiles.test                Wikijump
127.0.0.1                   profiles.wikijump.test      Wikijump
127.0.0.1                   template-en.wikijump.test   Wikijump
127.0.0.1                   sandbox.wikijump.test       Wikijump
127.0.0.1                   scp-wiki.wikijump.test      Wikijump
::1                         wikijump.test               Wikijump
::1                         www.wikijump.test           Wikijump
::1                         wjfiles.test                Wikijump
::1                         profiles.wikijump.test      Wikijump
::1                         template-en.wikijump.test   Wikijump
::1                         sandbox.wikijump.test       Wikijump
::1                         scp-wiki.wikijump.test      Wikijump
```

docs/development.md
```diff
@@ -1,32 +1,136 @@
...
 
-You can create a docker container using the following:
+The first step is to build the Docker image, which is a single source of information for Docker to use in a container later:
 
 ```bash
-$ docker build . -t scpwiki/wikijump:local
-$ docker create --name wikijump -p 80:80 scpwiki/wikijump:local
-$ docker start wikijump
+$ cd installer
+$ docker build . -t scpwiki/wikijump:local --no-cache
```
ammongit, line 66: Do we want to necessarily tell people to use `--no-cache` every time? It really speeds up builds if you're doing it a lot

docs/development.md
```diff
@@ -1,32 +1,136 @@
...
 ```
 
-Instead of building Wikijump locally, you can also pull the image from the Docker Hub:
+Then load that image into a container, which is a VM that runs only the image you created:
 
 ```bash
-$ docker create --name wikijump -p 80:80 scpwiki/wikijump:latest
+$ docker create --name wj --publish 80:80 --publish 443:443 -it -d scpwiki/wikijump:local
```
ammongit, line 72: `-it` and `-d`?

ammongit: Also what's wrong with specifying `docker run` instead?

docs/development.md
```diff
@@ -1,32 +1,136 @@
...
 ```
 
-Then terminate it:
+The container is already running, but to start the services inside it that run Wikijump you will need to send a start command:
```
ammongit, line 75: Not true, the container is created but not started.

docs/development.md
```diff
@@ -1,32 +1,136 @@
...
 
 ```bash
-$ docker stop wikijump
+$ docker start wj
 ```
 
-Alternatively, you can install to your local system using `install.sh`. This may require tinkering depending on your exact platform and environment.
+Then navigate to https://www.wikijump.test in your browser.
```
ammongit, line 81: Add a note about self-signed certificates.

docs/development.md
```diff
@@ -1,32 +1,136 @@
...
+```bash
+$ docker container ls  # List containers
+$ docker rm [ID]       # Destroy the container with this ID
+$ docker images        # List images
+$ docker rmi [ID]      # Remove the image with this ID
+```
+
+### Installation via Docker from the Docker Hub
+
+Instead of building Wikijump locally, you can also pull the `latest` image from the Docker Hub:
```
emmiegit, line 102:
```suggestion
Instead of building Wikijump locally, you can also pull the `latest` image from the Docker Hub:

(Note that this image may not be fully up-to-date with the current `develop`)
```

docs/development.md
```diff
@@ -1,32 +1,136 @@
...
+On Linux with systemd:
+
+```bash
+# systemctl restart nginx
+```
+
+On Linux with sysv-init, Windows via WSL2, or from within a Docker container:
+
+```bash
+# service nginx restart
```
ammongit, line 135: These get formatted as comments rather than root shells, would change to just \`\`\`

docs/development.md
rossjrw: I'm not really sure - I've found that if I've changed something and don't add `--no-cache`, it just rebuilds the same thing without my changes. I'd argue that it's better to add the option and give the reader the option of omitting it, rather than them potentially not knowing what to do if they do need it.

docs/development.md
rossjrw: I liked the explicitness of having one command to create a container and another to start it. It wasn't obvious to me that `run` created a container and didn't destroy it afterwards - I had quite the buildup. I know `--rm` exists, but that's another layer of implicitness.

So long as it's documented, I guess it doesn't really matter either way. This one makes slightly more sense to me.

docs/development.md
rossjrw: I added a note about ignoring the browser's warning - is this what you meant?

rossjrw:
This PR isn't picking up on new commits to `doc-docker`, so closing and opening a new PR.

emmiegit:
New PR: https://github.com/scpwiki/wikijump/pull/57

PR #53, opened 2020-11-17
Set shell scripts to use UNIX newlines

ammongit:


rossjrw (approved):

ammongit:
thanks @rossjrw @pxdnbluesoul

PR #59, opened 2020-11-19
[WJ-254] Add logo

rossjrw:
This PR adds the current working Wikijump logo, which I've deemed as 'working' unlike the other logos we've come up with because the J is currently being used as our JIRA logo.

Obviously this logo is not finalised or set in stone.

<p align="center"></p>

I've added a new top-level directory `assets` and the logo is `logo.svg`.

However, I considered naming it `logo.inkscape.svg`, because it's not minimised at all and contains many Inkscape-specific properties. We'll be able to use this safely on our frontend via Webpack's SVG minimisation, but it might be nice to have a canonical minimised SVG on hand, or even a canonical PNG. Is this desirable? We wouldn't commit any other minified code to VC.

For posterity's sake, previous logos:

<p align="center"></p>


emmiegit:
I would name it `logo.inkscape.svg` for now so at least we note we want to make distinctions. When we're more sure of the logo we can add a `logo.min.svg` version then.

I do think we should have a minified logo in the source so it's easy to access. It's not very big, and it's not like we'll be changing the logo very frequently (as opposed to committing minified bundles in the repo, which would change every build).

emmiegit (approved):

PR #60, opened 2020-11-19
Add docker-compose.yaml to separate services into containers

emmiegit:
I apologize in advance for the length of this PR. However it's made a lot of changes there are parallel efforts I don't want to interfere with.

In a nutshell, this PR:
* Replaces the singular `Dockerfile` with a `docker-compose.yaml` file with containers for memcached, postgres, and php-fpm.
* Removes old build files or places them in `legacy`
* Rename `installer -> install`

The primary reason I want to merge this is to preserve the postgres changes. The `scpwiki/postgres` image properly instantiates a postgresql container with all our seed data. Other changes include:

* Properly initializes the database outside of the `docker-entrypoint.sh` script
* Creates a system user `wikijump`, to allow for peer authentication
* Creates a database `wikijump` owned by postgres user `wikijump`
* Ensures all tables exist within the `wikijump` database (previously `postgres`)
* Decompresses the seed SQL file so we have actual diffs (until we have actual migrations anyways)

Since bluesoul is working on the web container, we can mostly copy that for the `docker-compose.yaml` here, since there are some implementation efforts that aren't worth replicating.


install/docker-compose.yaml
```diff
@@ -0,0 +1,25 @@
...
+      args:
+        FILES_DOMAIN: 'wjfiles.test'
+        POSTGRES_PASSWORD: 'wikijump-postgres'
+    restart: always
+  php-fpm:
+    build:
+      context: ./php-fpm
+      args:
+        MAIN_DOMAIN: 'wikijump.test'
+        FILES_DOMAIN: 'wjfiles.test'
```
danieltharp, line 19: I'm good with this in the interim, long term we're going to want to replace these with environment variables.

install/postgres/run.sh
```diff
@@ -0,0 +1,11 @@
...
+set -eu
+
+echo "Postgres version: $POSTGRES_VERSION"
+echo "Postgres cluster: $POSTGRES_CLUSTER"
+echo
+echo "Running database in foreground..."
+
+exec \
+	pg_ctlcluster "$POSTGRES_VERSION" "$POSTGRES_CLUSTER" \
+	start --foreground --stdlog
```
danieltharp, line 11: We'll want to verify the logging for everything is something that can be picked up with `docker logs` for an end-user.

danieltharp (approved):

rossjrw (changes requested):
I'd like to see the new installation process documented before I can approve for merge. (I made a commit to fix an `installer`->`install`, but that was a bit premature given that I hadn't realised the rest of the process was different, so apologies!)

`docker-compose up` seemed to be right right way to go on my end, but it got stuck during the `install php_fpm` stage - it looks like it's started up the DB per postgres/run.sh during installation rather than during deployment:

image

Once I can get this running on my end, which will probably be via documentation telling me what I need to do, it'll be thumbs up from me.

install/docker-compose.yaml
emmiegit: These are `ARG` directives, which from what I can tell are essentially the same as `ENV` to Docker. The idea here is they can be overridden within the `docker-compose.yaml` file.

But yeah, if there's a cleaner and less duplicative way of doing this, I'm all ears.

install/postgres/run.sh
emmiegit: If we want things in `docker logs` we should remove `--stdlog` and it'll appear in stdout, and thus `docker logs`.

emmiegit:
@rossjrw I added documentation, but I don't understand what this means:

> `docker-compose` up seemed to be right right way to go on my end, but it got stuck during the `install php_fpm` stage - it looks like it's started up the DB per postgres/run.sh during installation rather than during deployment:

The container adds the seed and setups up the database during the docker build. It just starts the service when run.

rossjrw (approved):

emmiegit:
thanks @pxdnbluesoul @rossjrw

PR #65, opened 2020-11-20
Fixes to docker-compose setup and insecure cookie fallback functions

danieltharp:
With this PR a `docker-compose build` and `docker-compose up` gives a working stack that can be connected to on www.wikijump.test and allows for login over HTTP in Firefox, Chromium based browsers still won't work. I want to put the logic for whether or not nginx listens on 443 behind some args because the AWS version won't need it, SSL stuff will be handled by the load balancers.

### DIFF
--- a/install/php-fpm/Dockerfile
+++ b/install/php-fpm/Dockerfile
@@ -98,6 +98,7 @@ RUN npm install && \
 
 # Cleanup
 RUN rm -rf /src
+RUN chown -R www-data:www-data .
 
 # Main process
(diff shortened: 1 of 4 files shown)


emmiegit:
Porting this from chat since it's relevant:
> Cookie “wikijump_token7” will be soon rejected because it has the “SameSite” attribute set to “None” or an invalid value, without the “secure” attribute. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite

emmiegit (approved):

PR #69, opened 2020-11-22
Laravel Integration First Steps

danieltharp:
This is enough to begin work from. The new nginx config is such that requests to paths at `/wikijump/*` will go to Laravel, and anything else will stay in the current Ozone-centric ecosystem. Obviously there needs to be a lot more done, the short-term goals beyond this will make sure that both sets of cookies are applied, and begin exploration of further namespacing for legacy/ozone stuff.


danieltharp:
Also, nice.

install/php-fpm/Dockerfile
```diff
@@ -100,8 +92,18 @@ RUN npm install && \
...
+
+RUN mv .env.example .env && php artisan key:generate
+
 # Main process
 # Let the upstream ENTRYPOINT handle running php-fpm
 ADD ./entrypoint.sh /usr/local/bin/
 CMD ["nginx", "-g", "daemon off;"]
-ENTRYPOINT ["sh","/usr/local/bin/entrypoint.sh"]
\ No newline at end of file
+ENTRYPOINT ["sh","/usr/local/bin/entrypoint.sh"]
```
emmiegit:
```suggestion
ENTRYPOINT ["sh", "/usr/local/bin/entrypoint.sh"]
```

install/php-fpm/Dockerfile
```diff
@@ -100,8 +92,18 @@ RUN npm install && \
...
 
+# Install nginx config files
+COPY etc/nginx/ /etc/nginx/
+
+# Enable wikijump site in nginx
+RUN mkdir /etc/nginx/sites-enabled && \
+    cp /etc/nginx/sites-available/wikijump /etc/nginx/sites-enabled/wikijump && \
+    rm -f /etc/nginx/sites-enabled/default
+
+RUN mv .env.example .env && php artisan key:generate
```
emmiegit: Would probably be more clear with `install` vs `mv`, but minor nitpick.

web/.editorconfig
```diff
@@ -0,0 +1,15 @@
+root = true
```
emmiegit: Should be `false`, it's at `/web` not `/`

web/.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
```
emmiegit, line 9: Normally I'm all for this option, but it's important to note that any wikitext sources will get munged by this (e.g. `> ` and the like)

web/.gitattributes
```diff
@@ -0,0 +1,5 @@
+* text=auto
+*.css linguist-vendored
+*.scss linguist-vendored
+*.js linguist-vendored
```
emmiegit, line 4: Huh, are all CSS and JS files in this subdirectory generated?

web/.gitignore
```diff
@@ -19,3 +19,15 @@ composer.phar
 # Lock Files
 composer.lock
 package-lock.json
+
+# Laravel suggestions not already included.
```
emmiegit, line 23: Not sure what this means

web/.gitignore
```diff
@@ -19,3 +19,15 @@ composer.phar
...
+/public/hot
+/public/storage
+/storage/*.key
+.env
```
emmiegit, line 27: Probably fits better in the `Configuration files` section

web/README.md
```diff
@@ -0,0 +1,61 @@
+<p align="center"><a href="https://laravel.com" target="_blank"><img src="https://raw.githubusercontent.com/laravel/art/master/logo-lockup/5%20SVG/2%20CMYK/1%20Full%20Color/laravel-logolockup-cmyk-red.svg" width="400"></a></p>
```
emmiegit: Is this all of Laravel's original README? Does this really need to be in our repo?

web/app/Console/Kernel.php
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Define the application's command schedule.
+     *
+     * @param  \Illuminate\Console\Scheduling\Schedule  $schedule
+     * @return void
+     */
+    protected function schedule(Schedule $schedule)
+    {
+        // $schedule->command('inspire')->hourly();
```
emmiegit, line 27: What does "inspire" do?

danieltharp:
I spotted some things in here that got munged during some renaming, I need to fix some things before anyone merges.

web/app/Http/Middleware/Authenticate.php
```diff
@@ -0,0 +1,21 @@
...
+     * Get the path the user should be redirected to when they are not authenticated.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return string|null
+     */
+    protected function redirectTo($request)
+    {
+        if (! $request->expectsJson()) {
+            return route('login');
+        }
```
emmiegit, line 19: I would advocate for explicit null return here

web/app/Http/Middleware/RedirectIfAuthenticated.php
```diff
@@ -0,0 +1,32 @@
...
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @param  string|null  ...$guards
+     * @return mixed
+     */
+    public function handle(Request $request, Closure $next, ...$guards)
+    {
+        $guards = empty($guards) ? [null] : $guards;
```
emmiegit, line 22: Maybe leave a comment why it needs to have `null` instead of being purely empty?

web/app/Http/Middleware/TrimStrings.php
```diff
@@ -0,0 +1,18 @@
...
+class TrimStrings extends Middleware
+{
+    /**
+     * The names of the attributes that should not be trimmed.
+     *
+     * @var array
+     */
+    protected $except = [
+        'password',
+        'password_confirmation',
```
emmiegit, line 16: Why would passwords (or password confirmation codes) have leading/trailing spaces?

web/app/Http/Middleware/TrustHosts.php
```diff
@@ -0,0 +1,20 @@
...
+{
+    /**
+     * Get the host patterns that should be trusted.
+     *
+     * @return array
+     */
+    public function hosts()
+    {
+        return [
+            $this->allSubdomainsOfApplicationUrl(),
```
emmiegit, line 17: Is this correct? If `allSubdomainsOfApplicationUrl()` returns a list it'll be returning `[ [ domain... ] ]`

web/app/Models/User.php
```diff
@@ -0,0 +1,43 @@
...
+     * @var array
+     */
+    protected $fillable = [
+        'name',
+        'email',
+        'password',
+    ];
+
+    /**
+     * The attributes that should be hidden for arrays.
```
emmiegit, line 26: What does "hidden for array" mean?

emmiegit:
Seems like a lot of this was copied or heavily adapted from a Laravel template or something? That's fine of course, but we should walk through or document basically where things are Wikijump-custom and which areas are inherited and will need to be changed.

emmiegit (approved):
nice

web/.editorconfig
danieltharp: Per the docs this is not indicating the root of any content, rather:

> root: special property that should be specified at the top of the file outside of any sections. Set to true to stop .editorconfig files search on current file.

web/.gitignore
danieltharp: This is a transitional list of stuff that's required by laravel but nothing else, so if I need to back all of this out I don't have to think all that hard on what needs removing.

web/app/Console/Kernel.php
danieltharp: It's an example provided by laravel for how the syntax works (in this case invoking a command from the `artisan` binary.)

web/README.md
danieltharp: No, removing.

web/app/Http/Middleware/TrimStrings.php
danieltharp: Not my code, but presumably because it's valid ASCII and a good way to increase the entropy of your password.

web/app/Http/Middleware/TrustHosts.php
danieltharp: `allSubdomainsOfApplicationUrl()` returns a string or null.

web/app/Models/User.php
danieltharp: When you work with a user object later, you can assume you don't want the password field or the remember token returned in the selection. This can be overridden if needed but it's a handy way to secure secrets for most use cases.

web/app/Http/Middleware/Authenticate.php
danieltharp: `null` is a valid return here but when we start writing code for the laravel side I'll show you how returning the route in the middleware is probably what you want.

web/.editorconfig
emmiegit: That line was a bit unclear to me, so I looked around. It seems based on https://github.com/editorconfig/editorconfig/issues/376 that it is in fact referring to a root directory? Since it inherits rules

web/app/Http/Middleware/Authenticate.php
emmiegit: Okay sounds fine

web/.editorconfig
danieltharp: I am a liar and a fool.

PR #72, opened 2020-11-23
Traefik integration

danieltharp:
This is a minimum viable product for local development using Traefik as an edge proxy. It includes mandatory connection upgrades to HTTPS and furnishes its own self-signed certificates as needed.

### DIFF
--- a/install/docker-compose.yaml
+++ b/install/docker-compose.yaml
@@ -46,3 +46,4 @@ services:
     volumes:
       # So that Traefik can listen to the Docker events
       - /var/run/docker.sock:/var/run/docker.sock:ro
+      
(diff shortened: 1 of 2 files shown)


install/docker-compose.yaml
```diff
@@ -16,12 +16,33 @@ services:
...
       - cache
       - database
     restart: always
     depends_on:
       - cache
       - database
+    labels:
+      - traefik.enable=true
+      - traefik.http.routers.php-fpm.tls=true
+      - traefik.http.routers.php-fpm.rule=Method(`GET`,`POST`,`PUT`,`PATCH`,`DELETE`)  # Poor man's default route
```
emmiegit: Missing `HEAD`

emmiegit (approved):

PR #74, opened 2020-11-24
Remove redundant ListResolver.php file per WJ-189

danieltharp:
This should get Composer to shut up.

This file is not actually used anywhere in the codebase, the calls for "ListResolver" are all from a smarty template that assumes a different path than the one this file is in, but this is the one being loaded by Composer due to both files occupying a null namespace.


emmiegit:
https://scuttle.atlassian.net/browse/WJ-189

emmiegit (approved):

PR #76, opened 2020-11-25
Decompress seed.sql

emmiegit:
Now changes to the seed will have actual diffs, instead of an opaque "changed binary" each time.


rossjrw (approved):

emmiegit:
thanks @rossjrw

PR #78, opened 2020-11-25
[WJ-250] Convert modules to TypeScript

rossjrw:
[WJ-250](https://scuttle.atlassian.net/browse/WJ-250)

Converting modules from Javascript to Typescript was initially scheduled for #34, but I considered it more important to merge the core TS conversions into the main branch once they were in a stable state. Progress on the backend has been advancing much faster than I expected and it's now time to get these modules operational.

- [ ] Type each module
- [ ] Attach each module to a `modules` object
- [x] Attach the `modules` object to the main `Wikijump` object
- [ ] Investigate modules that are not attached to the `modules` object (#34 has a comment about this)
- [x] Bundle typed modules
- [ ] Update PHP module handler not to fetch modules JS
- [ ] Update JS module handler (`OZONE.ajax.requestModule`) not to fetch modules JS
- [ ] Remove all the stuff that's attached to `window`
- [ ] Standardise or otherwise sort out init methods - they need to be called when the module is activated, not when the JS appears, as it will be bundled regardless of whether it's being used
- [ ] Check whether any `expect-error` declarations should be removed
- [ ] Rename all "ManagerSite" to "ManageSite"

Noting that in a future PR I will be implementing code splitting, which will split rarely-used modules into their own bundles. However, I'll be doing this in a more intelligent way than Wikidot's naïve "every module is its own bundle" approach. For this PR, a single bundle is fine. 

Ideas for the future:
* There are a few cases where stuff from the database is sent directly to the JS, e.g. category metadata. It would be useful to have a file solely for database types.
* There is some code that specifically converts https to http - get rid of this!


rossjrw:
Old modules have free reign to communicate via the `window` object. New modules will be held to higher standards. Noting the following places where modules implicitly attach stuff to the `window` object (incomplete) (JS unless otherwise specified):

Module | Attaches | Used in | Fix
--- | --- | --- | ---
WikiLayout template | `HTTP_SCHEMA`, `URL_HOST`, `URL_DOMAIN` | A lot of places
`Wikijump.page.callbacks.editClick` | `editMode` | `edit/PageEditModule`
`account/AccountMessagesModule` | `inboxPage()` | `account/AccountMessagesModule`, `account/pm/PMInboxModule` (JS & template), `account/pm/PMComposeModule` | Exported
`account/AccountMessagesModule` | `sentPage()` | `account/pm/PMSentModule` (JS & template) | Exported
`account/AccountMessagesModule` | `draftsPage()` | `account/pm/PMDraftsModule` (JS & template) | Exported
`account/AccountModule` (template) | `composeTo`, `inboxMessage` | `account/AccountMessagesModule`
`account/AccountModule` (template) | `accountStartPage` | `account/AccountModule`
`forum/ForumViewThreadModule` (template), `forum/ForumCommentsListModule` (template), `forum/ForumCommentsModule` | `Wikijump.forumThreadId` | `forum/ForumViewThreadModule`, `forum/sub/ForumThreadMoveModule`, `forum/sub/ForumEditThreadBlockModule`, `forum/sub/ForumEditThreadStickinessModule` | [Retained](https://github.com/scpwiki/wikijump/pull/78#issuecomment-736901677)
`forum/ForumNewThreadModule` (template) | `Wikijump.cancelUrl` | `forum/ForumNewThreadModule` | Modularised
`forum/ForumViewThreadModule` | `fDiv` (`foldedPost`) | `forum/ForumViewThreadModule` | Scoped
`forum/ForumViewThreadModule` | `togglePostOptions()` | `templates/macros/Forum.tpl`
`forum/ForumViewThreadModule` | `togglePostFold()` | `templates/macros/Forum.tpl`
`forum/ForumViewThreadModule` | `postReply()` | `templates/macros/Forum.tpl`
`managesite/ManageSiteAdminsModule` | `removeAdmin()` | `managesite/ManageSiteAdminsModule` (template) | Modularised
`managesite/ManageSiteModule` | `moduleName` | | Scoped
`managesite/ManageSiteModule` (template) | `smStartPage` | `managesite/ManageSiteModule`
`login/LoginModule` | `getQueryString()` | `login/LoginModule` | No change
`login/LoginModule` | `setCookie()` | `login/LoginModule` | No change


Fix | What it means
--- | ---
Exported | The global variable has been moved to the ES6 import/export system.
Scoped | The global variable was changed to a local variable.
Modularised | The global variable was attached to the module.
Retained | The global variable was kept, practically unchanged.
No change | The global variable is now scoped by default because it is in a JS module and not globally exported, which for this particular case is fine.
<empty> | The global variable still exists. It has either been declared and typed with a `declare` statement, or a compiler comment has been added that ignores the error (`@ts-expect-error`). These are both temporary solutions.

Where a function is defined globally and then used only in the JS, it can be left as-is and exported. I wonder how many other methods could or should be rewritten like this.

Noting that a bunch of templates will probably have JS functions embedded in `onclick`s and stuff.

Noting that `php/smarty_plugins/function.pager.php` defines a Smarty function that accepts a JS function to be executed that powers a pager. The function is attached to the pager element's `onclick`. Other plugins might do something similar.

rossjrw:
These modules are all designed to be activated the instant they're loaded, so I expect many of them to immediately fire an initialisation function. I'll need to find a way around this.

In fact, I counted the number of modules that define and call their own `init` method and found 63 - that's a lot less than the ~146 total modules, and less even than the 72 registered modules. I'll have to check to see what means a module has `init` and what doesn't.

Assuming that investigation doesn't reveal anything problematic, what I'll probably do is keep the `init` methods and change the module handlers to simply fire that method instead of AJAXing the module out of oblivion. Of course that approach assumes that all modules have an `init`, which isn't the case.

One thing to note is that whenever a module defines its own `init`, it always calls it immediately, and the method is never used anywhere else. Exceptions:
* `edit/PageEditModule`: called with a 10ms delay
* `login/LoginModule`: called with a 100ms delay
* `sitesactivity/RecentWPageRevisionsModule`: call is commented out
* `account/pm/PMDraftsModule`: is never called
* `account/pm/PMSentModule`: is never called
* `account/pm/PMInboxModule`: is never called

During typing, I've removed several empty `init` methods. They're typically used for adding stuff like event listeners that require the DOM to be ready, though as the method is always called immediately, there's not much point in having that stuff be its own method, and I commend the Wikidot developers for doing it anyway. 

However, a lot of modules attach functionality directly to the template with `onclick` methods. In some cases this bypasses the need for an init method. I'm not sure whether I like this - some discussion is needed and I'll be making an issue for it.

rossjrw:
`Wikijump.forumThreadId` is an interesting case.

It's set by: 
* `forum/ForumCommentsModule`: top-level as `[[module Comments]]`
* `forum/ForumViewThreadModule` (template): top-level as `[[module ForumThread]]`, JS requested by `forum/ForumCommentsListModule` (php)
* `forum/ForumCommentsListModule` (template): requested by `forum/ForumCommentsModule`

It's used by:
* `forum/ForumViewThreadModule`: ^
* `forum/sub/ForumThreadMoveModule`: requested by `forum/ForumViewThreadModule`
* `forum/sub/ForumEditThreadBlockModule`: requested by `forum/ForumViewThreadModule`
* `forum/sub/ForumEditThreadStickinessModule`: requested by `forum/ForumViewThreadModule`

So it looks like there's two entry points: `forum/ForumCommentsModule` and `forum/ForumViewThreadModule`. `forum/ForumCommentsModule` has no `init`, but requests `forum/ForumCommentsListModule` when <_show comments_> is clicked. `forum/ForumCommentsListModule` has no JS, but the PHP makes a request for the JS of `forum/ForumViewThreadModule`, indicating that they are effectively the same module in terms of JS. The template confirms this, mostly referring to `forum/ForumViewThreadModule`; the PHP of the two modules are very different, at least. Because only the JS of `forum/ForumViewThreadModule` is requested and not the template, the assignment is skipped, resulting in `forum/ForumCommentsModule` being the only source of `Wikijump.forumThreadId` for this chain.

_(`forum/ForumCommentsListModule` requests the JS for `forum/ForumViewThreadModule` via an `extraJs` parameter on the module internally. There are only 7 modules that use this feature. I will have to investigate if it is solely used for code deduplication or if it has other purposes.)_

But take `forum/sub/ForumThreadMoveModule` or either of its siblings. It needs a thread ID, and it doesn't know or care whether it comes from `forum/ForumCommentsModule` or `forum/ForumViewThreadModule`. Variable access via `Wikijump.modules.*.vars.forumThreadId` won't cut it because we won't know which module the ID came from.

`Wikijump.forumThreadId` works because it's module-agnostic, but it's not type-safe, and I don't want to make it type-safe and encourage randomly slapping shit onto the global object whenever we feel like it. There's `Wikijump.vars` which stores a couple of login-related things, but those are always necessary, and a thread ID is only necessary when the user is looking at a thread so I don't feel like touching it.

I don't see any easy way out. For now, I think I'm going to just do the `vars` thing, and add `forumThreadId` as a nullable property on the `Wikijump.vars` object. The best way, eventually, I think, would be to combine `forum/ForumCommentsModule` and `forum/ForumViewThreadModule` into one, and then the module chain will resolve itself.

emmiegit:
That is very convuluted and unfortunate.

rossjrw:
`login/LoginModule2` doesn't actually define a module - it's just an init function. When I eventually get around to sorting out how init functions work (which will be after the rest of the modules have been typed), I'll have to pay special attention here.

`createaccount/CreateAccount2Module` requests `createaccount/CreateAccount3Module`, which only exists in PHP (though I suspect the entirety of `createaccount/` is a deprecated version of `createaccount2/` so it doesn't really matter)

`createaccount2/CreateAccount3Module` is just an init function (not even that - at the time of writing it's just an unwrapped function call), so that implies to me that CreateAccount3 does not have a JS component.

rossjrw:
Note to self for later: by adding ESLint plugin `plugin:@typescript-eslint/recommended-requiring-type-checking`, I can automate **a lot** more fixes which will hopefully speed things up. Just wish I'd found out about this sooner.

rossjrw:
Noting that as of 2b630375207f3a0fd9821dbbd22bcd48e6aab8a4 this PR partially implements [WJ-370](https://scuttle.atlassian.net/browse/WJ-370) by removing the frontend code for the `pagepath` data form field type.

emmiegit:
@rossjrw This PR has been on hold for a bit, how much of this can we merge to `develop` as-is? I don't want to have wasted effort here since there's a lot of work, but this entire task is a very tall order.

rossjrw:
@ammongit There is a lot more work to do. There are about one third of the modules remaining that still need to be typed, and the module loader on both the front and back ends needs to be adjusted to do nothing but call a module's init (as opposed to fetching the source and adding it to the document). There's probably a bunch of other stuff that will come up, too. 

If we're redoing the frontend from scratch, all that work, and all the previous work in typing the Wikijump/OZONE core, will be wasted. 

Not that that's a bad thing - the code here is not great and there's a *lot* of it. This PR and the other one was just an initial typing pass. The tsconfig doesn't even have strict mode on, and I've made no attempt to restructure anything. It would be a series of mammoth tasks to get this to a sensible state, each one tedious and mind-numbing. So yeah, it's a lot of work, but it's work spent digging Wikidot's hole deeper rather than climbing out of it. 

I will need to look over this PR to see if there's anything that can be salvaged; once I've done that, I'll close it. I think I'd prefer to persist the branch until work on the new frontend is at least started.

PR #79, opened 2020-11-25
[WJ-250] Fix editMode + formatDates

rossjrw:
This PR fixes a couple of bugs introduced in #34.

* `editMode` is now reattached to the `window` object (will be super-fixed in #78) (this specifically was blocking page editing, which I need to test modules)
* Modules that wanted the deprecated `Wikijump.utils.formatDates` have been told to ask for `OZONE.utils.formatDates` instead

This probably works but I'm not going to merge it until I work out how to bind mounts with docker compose.

### DIFF
--- a/web/web/files--common/modules/js/edit/PageEditModule.js
+++ b/web/web/files--common/modules/js/edit/PageEditModule.js
@@ -569,7 +569,7 @@ Wikijump.modules.PageEditModule.utils = {
 
 Wikijump.modules.PageEditModule.init = function(){
 	if(Wikijump.page.vars.locked == true){
-		Wikijump.utils.formatDates();
+		OZONE.utils.formatDates();
 
 	} else {
 
--- a/web/web/files--common/modules/js/files/PageFilesModule.js
+++ b/web/web/files--common/modules/js/files/PageFilesModule.js
@@ -246,4 +246,4 @@ toggleFileOptions = function(fileId){
 }
 
 YAHOO.util.Event.addListener("show-upload-button", "click", Wikijump.modules.PageFilesModule.listeners.showUploadClick);
-Wikijump.utils.formatDates("action-area");
+OZONE.utils.formatDates("action-area");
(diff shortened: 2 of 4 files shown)


web/web/files--common/modules/js/edit/PageEditModule.js
```diff
@@ -569,7 +569,7 @@ Wikijump.modules.PageEditModule.utils = {
 
 Wikijump.modules.PageEditModule.init = function(){
 	if(Wikijump.page.vars.locked == true){
-		Wikijump.utils.formatDates();
+		OZONE.utils.formatDates();
```
danieltharp, line 572: Just verifying, we need this to reference an `OZONE` object?

web/web/files--common/modules/js/files/PageFilesModule.js
```diff
@@ -246,4 +246,4 @@ toggleFileOptions = function(fileId){
 }
 
 YAHOO.util.Event.addListener("show-upload-button", "click", Wikijump.modules.PageFilesModule.listeners.showUploadClick);
-Wikijump.utils.formatDates("action-area");
+OZONE.utils.formatDates("action-area");
```
danieltharp, line 249: Same question here.

danieltharp (changes requested):

web/web/files--common/modules/js/edit/PageEditModule.js
rossjrw: Yep, that's where the non-deprecated `formatDates` is: https://github.com/scpwiki/wikijump/blob/3f2efdaf889a03e797cc14e6cc875c08ab35d4f2/web/web/files--common/javascript/OZONE/utils.ts#L231

...and on seeing that snippet I realise that the method needs to be adjusted to account for an optional argument.

Nothing that I wouldn't have caught once I work out how to test this 😛

danieltharp (approved):

emmiegit (approved):

rossjrw:
With the changes made in #80 I'm able to verify that my fixes work, and I'm able to edit and save a page after the changes in #81, however, any edit to a page currently results in all its content being deleted. Not sure where the issue is coming from.

PR #90, opened 2021-01-03
Use alternate name for DATABASE_PORT var

danieltharp:


danieltharp:
This also renames the github actions file to include `php` as there will be another build one for postgres dev. This PR should be ready to squash and merge.

.github/workflows/docker-build-push-php.dev.yml
```diff
@@ -31,7 +32,7 @@ jobs:
             IMAGE_TAG: ${{ github.sha }}
           run: |
             set -ex
-            docker build --build-arg MAIN_DOMAIN=wikijump.dev --build-arg FILES_DOMAIN=wjfiles.dev -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG .
+            docker build --build-arg WIKIJUMP_REPO_BRANCH=${GITHUB_REF##*/} --build-arg MAIN_DOMAIN=wikijump.dev --build-arg FILES_DOMAIN=wjfiles.dev -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG .
```
rossjrw, line 35: Is `${GITHUB_REF##*/}` a shell variable? If so:

1. what does `##*/` mean?
1. I don't see where `GITHUB_REF` is being set - should this be an actions expression, e.g. [`${{ github.ref }}`](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#github-context)?

danieltharp: `GITHUB_REF` is an environment variable set by the GitHub Actions runner that contains the branch along with some other stuff, the `##*/` is a substring search in bash: https://tldp.org/LDP/abs/html/parameter-substitution.html

rossjrw: > ${var##Pattern} Remove from $var the longest part of $Pattern that matches the front end of $var. 

I assume it takes `refs/heads/whatever`, globs `*/` against `refs/heads/` and trims it? If so, works for me

rossjrw (approved):

PR #92, opened 2021-01-07
Move to single install folder

danieltharp:
This completes the refactor of the `infra` and `install` folders to a single location, it breaks out local vs. aws deployments and allows for 3 tiers of development.


emmiegit (approved):
I brought up my concerns regarding code drift in chat, but given the configurations will be different, I'm fine with the PR :pineapple:

PR #95, opened 2021-01-12
Update README, Wikijump is separate from EN Tech

emmiegit:
While there is significant, the Wikijump team is separate from the EN Tech Team, and going forward this will diverge further. Additionally, Wikijump as an entity will be incorporated separately from the SCP Wiki.

This PR intends to update the README to reflect this.


rossjrw (approved):

emmiegit:
thanks @rossjrw

PR #100, opened 2021-01-25
[WJ-153] Allow inline use of newline blocks

emmiegit:
To resolve [WJ-153](https://scuttle.atlassian.net/browse/WJ-153), this issue lifts the requirement that newline-separated blocks must have newlines, permitting them to be declared inline.

However they still maintain the newline-consuming behavior, so that existing uses of the syntax does not break.

Newly-accepted syntax:
```
[[div]]Alpha beta gamma[[/div]]
```
Which is equivalent to:
```
[[div]]
Alpha beta gamma
[[/div]]
```

### DIFF
--- a/ftml/src/parsing/rule/impls/block/blocks/include.rs
+++ b/ftml/src/parsing/rule/impls/block/blocks/include.rs
@@ -33,6 +33,7 @@ pub const BLOCK_INCLUDE: BlockRule = BlockRule {
     name: "block-include",
     accepts_names: &["include"],
     accepts_special: false,
+    newline_separator: true,
     parse_fn,
 };
 
--- a/ftml/src/parsing/rule/impls/block/blocks/include.rs
+++ b/ftml/src/parsing/rule/impls/block/blocks/include.rs
@@ -33,7 +33,7 @@ pub const BLOCK_INCLUDE: BlockRule = BlockRule {
     name: "block-include",
     accepts_names: &["include"],
     accepts_special: false,
-    newline_separator: true,
+    accepts_newlines: true,
     parse_fn,
 };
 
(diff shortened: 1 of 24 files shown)


ftml/src/parsing/rule/impls/block/blocks/include.rs
```diff
@@ -33,7 +33,7 @@ pub const BLOCK_INCLUDE: BlockRule = BlockRule {
     name: "block-include",
     accepts_names: &["include"],
     accepts_special: false,
-    newline_separator: true,
+    accepts_newlines: true,
```
rossjrw, line 36: Just noting that I'm curious how this will work for blocks such as include, which to my knowledge doesn't have a concept of a head or tail, which is what I assumed `accepts_newlines` is for - it just has a single tag

ftml/src/parsing/rule/impls/block/mod.rs
```diff
@@ -61,8 +61,9 @@ pub struct BlockRule {
     /// `[[user aismallard]]` and `[[*user aismallard]]`.
     accepts_special: bool,
 
-    /// Whether this block wants its head and tail to be separated by newlines.
-    newline_separator: bool,
+    /// Whether this block allows its head and tail to be separated by newlines.
```
rossjrw: While "allows" is probably clear enough, I'd prefer to specify explicitly that the newlines are optional (assuming this is the case)

rossjrw (approved):

ftml/src/parsing/rule/impls/block/blocks/include.rs
emmiegit: Do you have any examples other than `[[include]]`? Includes are a special case that are replaced before the preprocessor step.

But if you mean head-only blocks like `[[module Rate]]`, it will try to consume a newline after, then stop because there's no body to fetch.

rossjrw: That's a good point regarding include. The rating module is also the only example I had in mind, I guess there's also [[newlines n]].

I think that default response - consuming a newline if possible and stopping there - is fine for cases like these.

emmiegit:
thanks @rossjrw

PR #101, opened 2021-01-25
[FTML-52] Add html and iframe blocks

emmiegit:
This PR adds `[[html]]` and `[[iframe]]` to the parser.

* https://scuttle.atlassian.net/browse/FTML-52
* https://scuttle.atlassian.net/browse/WJ-99
* https://scuttle.atlassian.net/browse/WJ-101

PR #102, opened 2021-01-26
[FTML-69] Allow arbitrary HTML attributes

emmiegit:
This PR accomplishes two things:
* Permits dashes and underscores in block argument keys
* Modifies styled elements to accept an arbitrary map of HTML attributes instead

Wikidot is rather restrictive, sometimes arbitrarily so, about which arguments it will pass on to the final HTML tag being created. Instead of this, it was proposed in [FTML-69](https://scuttle.atlassian.net/browse/FTML-69) to allow any other arguments to be passed into the final `Element` being constructed.

This blocks #101

### DIFF
--- a/ftml/src/parsing/rule/impls/block/parser.rs
+++ b/ftml/src/parsing/rule/impls/block/parser.rs
@@ -385,7 +385,7 @@ where
 
                     // Gather argument key string slice
                     let end = self.current();
-                    self.full_text().slice(&self.log(), start, end)
+                    self.full_text().slice_partial(&self.log(), start, end)
                 };
 
                 // Equal sign
--- a/ftml/src/parsing/rule/impls/block/parser.rs
+++ b/ftml/src/parsing/rule/impls/block/parser.rs
@@ -358,7 +358,7 @@ where
                             // End parsing block head
                             Token::RightBlock => {
                                 args_finished = true;
-                                break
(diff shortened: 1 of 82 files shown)


ftml/src/parsing/rule/impls/block/parser.rs
```diff
@@ -334,18 +339,54 @@ where
...
+                        let current = self.current();
+                        match current.token {
+                            // End parsing block head
+                            Token::RightBlock => {
+                                args_finished = true;
+                                break;
+                            }
+
+                            // End parsing argument key
+                            Token::Whitespace | Token::Equals => break,
```
rossjrw, line 365: Would the distinction between this and end-of-block allow for boolean attributes like a `[[video autoplay]]` that maps to `<video autoplay>`? Not that it matters, because `<video autoplay="true">` is equivalent

rossjrw (approved):

ftml/src/parsing/rule/impls/block/parser.rs
emmiegit: No, this exists because we consume an optional whitespace afterwards, and the key ends either when it hits ` ` or `=`.

Presently all arguments require an associated value.

emmiegit:
thanks @rossjrw

PR #104, opened 2021-01-28
Add new miscellaneous blocks to parser

emmiegit:
This implements the following issues:

* [FTML-13](https://scuttle.atlassian.net/browse/FTML-13)
* [FTML-14](https://scuttle.atlassian.net/browse/FTML-14)
* [FTML-47](https://scuttle.atlassian.net/browse/FTML-47)

Which adds the following new blocks:
* `[[radio]]` - produces a radio button
* `[[checkbox]]` - produces a checkbox
* `[[hidden]]` - produces a hidden span (equivalent to `<span style="display: none;">`)
* `[[invisible]]` - produces an invisible span (equivalent to `<span style="color: transparent;">`)
and abstract syntax tree JSON tests for each.


ftml/src/parsing/rule/impls/block/blocks/checkbox.rs
```diff
@@ -0,0 +1,56 @@
...
+    in_head: bool,
+) -> ParseResult<'r, 't, Element<'t>> {
+    debug!(
+        log,
+        "Parsing checkbox block";
+        "in-head" => in_head,
+        "name" => name,
+    );
+
+    assert_eq!(special, false, "Checkboxes don't allow special variant");
```
rossjrw: A special variant could start pre-checked (or some other syntax for that)

ftml/src/parsing/rule/impls/block/blocks/hidden.rs
```diff
@@ -0,0 +1,60 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+use super::prelude::*;
+
+pub const BLOCK_HIDDEN: BlockRule = BlockRule {
+    name: "block-hidden",
+    accepts_names: &["hidden"],
+    accepts_special: false,
+    accepts_newlines: false,
```
rossjrw: This means that it only works inline, right? What if the user wants to hide a couple of paragraphs, for whatever misguided reason?

ftml/src/parsing/rule/impls/block/blocks/hidden.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    let arguments = parser.get_head_map(&BLOCK_HIDDEN, in_head)?;
+
+    // Get body content, without paragraphs
+    let (elements, exceptions) = parser.get_body_elements(&BLOCK_HIDDEN, false)?.into();
+
+    let element = Element::StyledContainer(StyledContainer::new(
+        StyledContainerType::Hidden,
+        elements,
+        arguments.to_hash_map(),
```
rossjrw, line 56: Does [[hidden]] have any arguments? If not, does this just pass any arguments through to whatever html is produced?

rossjrw (approved):

ftml/src/parsing/rule/impls/block/blocks/hidden.rs
emmiegit: Oh right this is a good catch, I forgot to set these.

ftml/src/parsing/rule/impls/block/blocks/hidden.rs
emmiegit: Yeah, it just passes through arguments like it were a `[[span]]`.

ftml/src/parsing/rule/impls/block/blocks/checkbox.rs
ammongit: I'm not sure what you mean. The checks here are asserts (meaning the program will crash) as safety, since the block rule should've already checked if special blocks are accepted.

emmiegit: Oh that's a good idea. I'll also add this for radio buttons too.

emmiegit:
thanks @rossjrw

PR #105, opened 2021-01-29
Add two missing fields to package.json

emmiegit:
npm warns about those two fields not existing, so here's a PR to add them. Minor change.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -1,5 +1,7 @@
 {
     "name": "wikijump",
+    "description": "SCP Wiki's fork of the unmaintained Wikidot",
+    "license": "agpl-3.0-or-later",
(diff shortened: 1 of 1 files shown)


rossjrw (approved):
I actually removed these fields intentionally. `web/` isn't intended to be an NPM package, so this sort of information is irrelevant as far as `package.json` is concerned - it's only to track dependencies. If it must have a description, I'd prefer to keep that description as vague as possible, so that if we ever change the description of this project, we can safely forget about updating this file.

emmiegit:
Sounds good, edited the description to be more vague and clear that it's not meant to be a node thing.

emmiegit:
thanks @rossjrw

PR #106, opened 2021-01-29
Update channel before pulling igbinary

emmiegit:
A user was having issues creating a local Wikijump instance via docker-compose, because this source repository needed updating. It seems to be inconsistent across users, but we may as well ensure it's up-to-date in the install script.


rossjrw (approved):

emmiegit:
thanks @rossjrw

PR #108, opened 2021-02-03
[FTML-76] Add lists to parser

emmiegit:
This PR adds bullet and numbered lists to the parser.

It includes their ability to be nested at arbitrary depths, the conversion process into an element tree, and heterogenous mixes of bullet and numbered list items. Several syntax tree tests are added to check these cases.

----

A number of internal implementation changes are introduced here:

**Addition of `Token::InputStart`:**
At the beginning of any token list, the first will be `Token::InputStart`, which mirrors how `Token::InputEnd` works. It always has a slice of `""` and a range of `0..0`, and represents the pseudo-token that matches the beginning of input. This was added to allow lists to begin at the start of a page, instead of only at newline termination. Tests are updated to reflect the new token.

**Move and rename of `ListStyle`:**
Previously kept at `ftml::enums::ListStyle`, the enum was moved to `ftml::tree::list::ListType`. Additionally it gained a new variant, `ListType::Generic`, to reflect lists that only contain sub-lists. This was necessary as part of how depth lists are converted into their tree form.

**Macro improvements:**
Macros can use `$(,)?` to allow for an optional trailing comma. Before a duplicate rule was added that had an extra `,` at the end.

**Creation of `NonEmptyVec<T>`:**
A new generic container called `NonEmptyVec` has been introduced. It is essentially what it says on the tin, a `Vec` which always has at least one element. Not all of the typical container methods and traits are implemented, only those that are actually used are. Additionally it does not store all the elements continguously, but rather uses the type system to assert that at least one item is present. It is utilized during the depth list to tree conversion process.

**Rules now return `Elements<'t>`:**
A new enum is introduced called `Elements`. It allows rules to return no, one, or many elements. Previously rules had to produce exactly one element, and in order to produce "nothing", `Element::Null` was returned. This variant has been removed, and `Elements::None` is returned instead.

The new variant, `Elements::Multiple`, stores a `Vec<Element>`, currently only used for lists. The existing behavior (returning one element) exists in `Elements::Single`, to avoid an allocation on every single rule execution if it's not needed.

This new enum also has an iterator implementation (specifically `IntoIterator` for owned `Element`s being returned) called `ElementsIterator`. This ensures that any variant can be trivially handled by treating it as a list of zero or more `Elements` being yielded. There are also tests to ensure it operates properly.

**Module rules now return `Option<Module<'t>>`:**
Similar to `Elements`, `Module::Null` was removed, and now all modules return an option.

**Allow `collect()` to accept end-of-input:**
Previously `collect()` and its variants would just fail with `ParseWarningKind::EndOfInput` when it hit the end. With lists, we instead want end of input to yield collection success, and terminate iteration there. So there is now new logic that allows accepting `Token::InputEnd` if one of the `ParseCondition`s accepts it. (Specifically, the end of input change is added after the close condition checks, allowing the success to go through if it exists).

**Depth logic:**
As lists can be nested at arbitrary depths, we need logic to properly build lists and sublists to reflect what the user is specifying. This implementation is more generous than Wikidot's, which fails in certain (albeit uncommon) cases. This necessitated the creation of `NonEmptyVec` (see above) and introduces `DepthStack`, which is a stack that is pushed and popped to build up the tree. It is generic, allowing any type `T` as its value, and allows for a "list type" `L`, which  is used to permit heterogenous mixing. This ensures that lists that have both numbered or bullet lines in them are nested properly (see [markdown test](https://gist.github.com/ammongit/1b46c27b615aa43065a9ca67eb71ca26)). For types where this property does not matter, simply set `L = ()`. There are tests to ensure the depth logic works, both with and without list types.

This logic will be reused when we get to implementing blockquotes.

**Tokenizer:**
This adds tokens `Token::BulletItem` and `Token::NumberedItem`, which simply correspond to `*` and `#`.

**List rule:**
Obviously, there must be a parse rule for lists.

This is an unusual rule, as rather than looking for `Token::BulletItem` or `Token::NumberedItem` directly, it instead must match against the start of a line. Thus the rule begins at either `Token::InputStart` or `Token::LineBreak`, and then looks for optional whitespace (used for determining depth) and then the list items. If they are found, then the depth list is constructed. If not, then the rule passes on. After consuming each list of elements at their corresponding depth (e.g. `[(0, Token::BulletItem, Element::Text("hi")), (1, Token::BulletItem, Element::Text("bye"))]`), it then runs `process_depths()`, which utilizes the `DepthStack` to convert this depth list into a tree. This list tree is then mapped into its corresponding `Element::List` using a recursive function. A maximum depth limit of 20 has been added as a reasonable upper limit, with a test for this, to avoid user-instigated stack overflow.

The actual tokens `Token::BulletItem` and `Token::NumberedItem` simply turn into text, since without the proper line context, they're just `*` or `#`.


ftml/src/parsing/depth.rs
```diff
@@ -0,0 +1,337 @@
...
+        if self.stack.is_single() {
+            // This is the last layer, so the pop/push trick doesn't work.
+            //
+            // Instead, output this entire thing as a finished list tree,
+            // then create a new one for the process to continue.
+            self.finish_depth_list(Some(ltype));
+        } else {
+            // We can just decrease and increase to make a new list
+            self.decrease_depth();
+            self.increase_depth(ltype);
```
rossjrw, line 69: lmao nice

ftml/src/parsing/rule/impls/list.rs
```diff
@@ -0,0 +1,189 @@
...
+
+                break;
+            }
+        };
+
+        // Check that the depth isn't obscenely deep, to avoid DOS attacks via stack overflow.
+        if depth > MAX_LIST_DEPTH {
+            info!(
+                log,
+                "List item has a deep greater than the maximum! Failing";
```
rossjrw:
```suggestion
                "List item has a depth greater than the maximum! Failing";
```

ftml/src/parsing/token/mod.rs
```diff
@@ -146,7 +153,17 @@ impl Token {
...
+                // Pest already adds a Token::InputEnd at the end
+                let start = ExtractedToken {
+                    token: Token::InputStart,
+                    slice: "",
+                    span: 0..0,
+                };
+
+                let mut tokens = vec![start];
+                tokens.extend(pairs.map(|pair| Token::convert_pair(log, pair)));
+                tokens
```
rossjrw, line 166: What's this?

ftml/src/tree/element.rs
```diff
@@ -195,3 +196,126 @@ impl slog::Value for Element<'_> {
...
+            Elements::None => true,
+        }
+    }
+
+    #[inline]
+    pub fn len(&self) -> usize {
+        match self {
+            Elements::Multiple(elements) => elements.len(),
+            Elements::Single(_) => 1,
+            Elements::None => 0,
```
rossjrw, line 222: So `Elements` is a wrapper for either a list of elements, a single element (but not in a list) or nothing? Why this instead of  compositing a list that can be empty?

rossjrw (commented):
No huge, glaring errors that jump out at me here - LGTM

rossjrw (approved):

ftml/src/parsing/token/mod.rs
emmiegit: The original code is
```rust
pairs.map(|pair| Token::convert_pair(log, pair)).collect()
```
Which runs `Token::convert_pair()` on each item in an iterator of `pest::Pair`s, in a 1:1 mapping.

We want to do the same, but add a new token (`Token::InputStart`) to the beginning. So we create a `Vec` with that as the first element (`vec![start]`), then run `.extend()` on the same iterator map to add all `Token::convert_pair()`-ed items after.

ftml/src/tree/element.rs
emmiegit: You are correct that a `Vec<Element<'t>>` could serve the same purpose logically in the code.

However `Vec` allocates space on the heap (except if it's empty). Given that the most common result for a rule is returning a single element, it's a bit wasteful to always allocate when we could just pass the object on the stack. So with this enum, we only allocate if it's a rule that actually returns more than one element, and needs to have dynamic sizing.

ftml/src/parsing/token/mod.rs
rossjrw: Sorry, that's GitHub helpfully providing the context to my selection - I was referring to the `tokens` on a line all by itself

emmiegit: Rust uses value expressions, so the evaluation result of a given block is its value. This means that the last thing referenced in it is the overall value, without needing an explicit `return` for every function.

For instance, for inline expressions you can just do:
```rust
fn plus_one(n: i32) -> i32 {
    n + 1
}
```

And this also applies to scopes, allowing you to tightly define bounds for bindings (variables) without needing to make everything a helper function:

```rust
let weighted_score = {
   let sigma_thing_blah = scores.iter().map(calc_sigma).sum();
   if sigma_thing_blah > threshold(scores.len()) {
       do_thing();
   }

   meta_score * calibrated(sigma_thing_blah)
};
```

The internal implementation bindings like `sigma_thing_blah` aren't available later in that function, as they were dropped at the end of the scope. All that happened was that `weighted_score` was assigned to some value. It's a nice way to group and modularize things and avoid code reaching into places where it logically should not.

----

So here, we have a local scope where we perform computations, and then the last item of the scope is the thing that's actually returned from match arm `Ok(pairs)`.

You'll notice that this function (`Token::extract_all`) actually has two layers of value expressions:

1. The method overall returns a `match` block, so the return value is whatever is returned on each of its match arms.
2. Within the `Ok(pairs)` arm, it returns the `tokens` list it was building in the code block.

emmiegit:
thanks @rossjrw

emmiegit:
https://scuttle.atlassian.net/browse/FTML-76

ftml/src/parsing/token/mod.rs
rossjrw: Ahhhhh! I see. Thank you!

PR #110, opened 2021-02-08
Add blockquotes

emmiegit:
This PR adds blockquote blocks (`[[blockquote]]` or `[[quote]]`) and classic Wikidot blockquotes (`> `).

Due to a regression involving typography substitutions (specifically `>>` into `»`), the token `Token::Quote` has been modified, and a parse rule has been added for those substitutions, similar to em dashes.


ftml/src/parsing/rule/impls/blockquote.rs
```diff
@@ -0,0 +1,161 @@
...
+            }
+        };
+        parser.step()?;
+        parser.get_optional_space()?; // allow whitespace after ">"
+
+        // Check that the depth isn't obscenely deep, to avoid DOS attacks via stack overflow.
+        if depth > MAX_BLOCKQUOTE_DEPTH {
+            info!(
+                log,
+                "Native blockquote has a depth greater than the maximum! Failing";
```
rossjrw, line 76: This is a new feature - what should happen when the limit is exceeded?

How feasible is such an attack? Is this an entirely pre-emptive measure or is there data to support it?

ftml/src/parsing/rule/impls/blockquote.rs
```diff
@@ -0,0 +1,161 @@
...
+
+    // Produce a depth list with elements
+    let mut depths = Vec::new();
+    let mut exceptions = Vec::new();
+
+    loop {
+        let current = parser.current();
+        let depth = match current.token {
+            // 1 or more ">"s in one token. Return ASCII length.
+            Token::Quote => current.slice.len(),
```
rossjrw, line 54: A nested blockquote can also be constructed with a bunch of spaces before the >, right?

ftml/src/parsing/rule/impls/blockquote.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+
+fn build_blockquote_element(list: DepthList<(), Vec<Element>>) -> Element {
+    let mut all_elements = Vec::new();
+
+    // Remove this trailing line break, these should
+    // only be between lines in the blockquote.
+    macro_rules! remove_trailing_line_break {
+        () => {
+            if let Some(Element::LineBreak) = all_elements.last() {
```
rossjrw, line 131: I'm always a bit thrown when I see an assignment in a conditional, but here I'm not sure what `Some(Element::Linebreak)` is or what it does - could you explain this?

ftml/src/parsing/rule/impls/double_angle.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    match current.token {
+        // « - LEFT-POINTING DOUBLE ANGLE QUOTATION MARK
+        Token::LeftDoubleAngle => ok!(text!("\u{0ab}")),
+
+        // » - RIGHT-POINTING DOUBLE ANGLE QUOTATION MARK
+        Token::Quote if current.slice == ">>" => ok!(text!("\u{0bb}")),
+
+        // Some other series of ">"s in a line
+        Token::Quote => Err(parser.make_warn(ParseWarningKind::RuleFailed)),
```
rossjrw, line 48: Could a string like `>>>test` or `test>>>test` reach this rule?

rossjrw (approved):

ftml/src/parsing/rule/impls/blockquote.rs
emmiegit: The "depth" of a blockquote is just how many `>`s there are. So with a maximum of 30 here, it means that anything like
```
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> Super deep blockquote
```
won't be processed as a blockquote. I can increase the depth if it's desired, it's not a big deal, it's just we need to have some kind of limit to avoid 200,000 `>`s being used to induce a stack overflow crash in the parser.

ftml/src/parsing/rule/impls/blockquote.rs
emmiegit: This does not appear to be the case


(Unless I'm misunderstanding something)

ftml/src/parsing/rule/impls/blockquote.rs
emmiegit: This is saying "if the last element is `Element::LineBreak`, then do ..."

The `Some(_)` here is because `.last()` returns an option -- if the collection is empty, then it returns `None`.

ftml/src/parsing/rule/impls/double_angle.rs
emmiegit: Yes, that is the intent. This rule will fail, then the next rule in the mapping is `RULE_TEXT`, which will just output it as text.

ftml/src/parsing/rule/impls/blockquote.rs
emmiegit: I will also note that a depth limit also exists in `RULE_LIST`, for the same reason.

ftml/src/parsing/rule/impls/blockquote.rs
rossjrw: Huh - I could've sworn nesting worked the same way for blockquotes as it does for lists. I guess not!

image

image

ftml/src/parsing/rule/impls/blockquote.rs
rossjrw: Aha - I thought `Some` was some function defined elsewhere. I'm reading up on the Option type now. I'll understand this line someday 😛

ftml/src/parsing/rule/impls/blockquote.rs
rossjrw: 30 is probably a pretty reasonable limit, though I think it's *almost* feasible for good-faith usage. If 30 isn't at any risk of causing an issue, I'd prefer to set this limit safely in the boundary between good-faith and bad-faith at something like 100.

ftml/src/parsing/rule/impls/blockquote.rs
emmiegit: Yeah it's really nice. The way I would think about it is that all types are non-nullable, unless made explicitly nullable by wrapping the `T` in `Option<T>`.

This way Java's billion-dollar mistake of `NullPointerException` is impossible -- if something is `Option` then the compiler enforces that you handle the `None` case, and if it's just `T`, then it's impossible for it to be "null".

emmiegit:
thanks @rossjrw

PR #112, opened 2021-02-11
[FTML-77] Add the [[size]] block

emmiegit:
This adds the `[[size]]` block, which takes its parameter as a CSS font size, and is passed into element's attributes.

There are also changes to resolve new lints that appear in the latest version of clippy.


emmiegit:
https://scuttle.atlassian.net/browse/FTML-77

rossjrw (approved):

emmiegit:
thanks @rossjrw

PR #113, opened 2021-02-12
[WJ-386] Add a CODEOWNERS file

emmiegit:
Adds a [CODEOWNERS file](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners) to document who owns which parts of the code, and automatically request reviews from relevant contributors.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,12 @@
+# This is a CODEOWNERS file, describing who is responsible for
+# maintenance of, or has experience with various parts of this
+# repository.
+
+# The ftml (Foundation Text Markup Language) service provides
+# parsing and rendering of wikitext.
+ftml                                @ammongit
+
+# Frontend files related to browser scripting or styling.
+web/web/files--common/dialogs       @rossjrw
+web/web/files--common/javascript    @rossjrw
+web/web/files--common/modules       @rossjrw


rossjrw (changes requested):
Marking yourself as owning FTML makes sense, but I'm not sure I'm comfortable marking myself as owning the frontend. Right now there's no difference between me and a hypothetical other contributor other than my work on the TypeScript conversion. I worry that marking myself as owning this - when I've not really even written any new code - is, in name at least, a deterrent for other contributors.

Once we've reached an internal system where we know for sure who's doing what, I think whoever is managing frontend development should be marked as the owner (which could be me). Right now I feel like my role on the frontend is because I am the only person available to do it at this stage in the project.

This all depends on the scope of the CODEOWNERS file, of course, but this is actually my first time using one so I'm unsure what standard practice is. "describing who is responsible for maintenance of" parts of the codebase certainly describes my role on the frontend right now, but I feel it's a little premature to define that right now.

rossjrw (approved):
I'm actually going to change my review action to an approval, because you're likely more familiar than me with this, but my comments above still apply.

emmiegit:
Well we can always edit the CODEOWNERS file later. If someone else comes on, and they end up taking over frontend maintenance, and you go work on something else, we can just add a PR to change it to reflect the new state of things.

It's meant to be a living document, and descriptive rather than prescriptive, in my view.

emmiegit:
thanks @rossjrw

PR #114, opened 2021-02-13
[FTML-79] Strip off category in triple-bracket links

emmiegit:
As demonstrated [here](http://scptestwiki.wikidot.com/triple-link-categories), triple-bracket links remove the category to make it more convenient to specify pages outside of `_default`:
image

This PR adds this functionality to ftml, changing the type of `LinkLabel::Url` from unit to `Option<Cow<str>>` (the null case being the original use, if the URL is the same as the label), and a helper function `strip_category()` which produces a slice without the category.

This is verified with a test for `strip_category()` and AST tests for the new behavior.

### DIFF
--- a/ftml/src/parsing/rule/impls/link_triple.rs
+++ b/ftml/src/parsing/rule/impls/link_triple.rs
@@ -218,7 +218,7 @@ fn test_strip_category() {
 
             assert_eq!(
                 actual, $expected,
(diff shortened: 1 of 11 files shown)


rossjrw (approved):

emmiegit:
thanks @rossjrw

PR #116, opened 2021-02-14
Bump ftml version to v0.5.0

emmiegit:
It's been a while since v0.4.0, we should increment the version number. Also fixes the repository link.


rossjrw (approved):

emmiegit:
thanks @rossjrw

PR #117, opened 2021-02-14
Move CodeCov badge

emmiegit:
It seemed visually inconsistent to have the CodeCov badge in the middle of two CI test badges, so I have moved it to be the first item.

Additionally I removed an extraneous `<a>` line that did not match with any `</a>`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,15 +5,14 @@
 </p>
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


rossjrw (approved):

emmiegit:
thanks @rossjrw

PR #118, opened 2021-02-15
Fix login issue

emmiegit:
Currently on [wikijump.dev](https://www.wikijump.dev/auth:login) attempting to login produces the following error:

```html
<br />
<b>Fatal error</b>:  Uncaught Error: Call to a member function getPassword() on null in /var/www/wikijump/web/lib/ozoneframework/php/core/SecurityManager.php:79
Stack trace:
#0 /var/www/wikijump/web/php/actions/Login2Action.php(36): SecurityManager-&gt;authenticateUser('aismallard', 'JKOvkOwL2L6ar1R...')
#1 /var/www/wikijump/web/php/utils/AjaxModuleWikiFlowController.php(227): Login2Action-&gt;loginEvent(Object(RunData))
#2 /var/www/wikijump/web/web/ajax-module-connector.php(11): AjaxModuleWikiFlowController-&gt;process()
#3 {main}
  thrown in <b>/var/www/wikijump/web/lib/ozoneframework/php/core/SecurityManager.php</b> on line <b>79</b><br />
```

<details>
<summary>Screenshot of the above error</summary>

screenshot

It returns 200 OK, as all error responses should... 🙄

</details>

----

This is a result of `$user` being `null` (as a user was not found), and then attempting to get this nonexistent user's password.

This PR makes the following changes:
* Adds a null check after attempting to retrieve the user.
* Performs some refactoring to make this section of code more readable.
* Adds clarifying documentation, such as the need to rebuild your container and build your JS bundle.
* Makes workflow changes that were necessary during the investigation of this bug.
* Specifically, it makes `lib/` available to the container when running in development mode.


rossjrw (approved):

emmiegit:
thanks @rossjrw

PR #119, opened 2021-02-15
[WJ-387] Disallow logging in via user ID

emmiegit:
Wikijump apparently allowed people to log in via user ID, a value most people do not know, and a behavior not reflected in mainline Wikidot. As such, it has been removed.

I am not adding a DEADCODE entry because it just used existing code mechanisms to implement this.


emmiegit:
https://scuttle.atlassian.net/browse/WJ-387

rossjrw (approved):

emmiegit:
thanks @rossjrw

PR #120, opened 2021-02-26
Move the glossary to the repo

emmiegit:
This creates a document here which describes common terms and their official name, a previous form of which existed in Confluence instead. It is being moved here as to be public in viewing and editing (via pull requests).

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -14,7 +14,8 @@ Definitions are not intended to be exhaustive, but rather quickly summarize what
 | Page category  |              |             | A particular namespace a page may exist in. Denoted by the part of a slug before the `:`. If a colon is absent, then the page is in category `_default`. |
 | Wikitext       | Source       | Wikidot code, ftml | The textual representation of the contents of a page. |
 | Vote           |              |             | An individual vote cast on a page, along with its timestamp and user information. |
-| Score          | Rating       |             | The calculated result of all the votes presently active on a page. This is **not** necessarily the same as `ups - downs`. |
+| Score          | Rating       |             | The calculated result of all the votes presently active on a page, per the site's chosen scoring algorithm. |
+| Rating         | Rating       |             | Wikidot's simple scoring algorithm, equivalent to the sum of `upvotes - downvotes`. |
 | File           |              | Attachment, upload | A blob attached to a page, with an associated filename and MIME type. |
 | Revision       |              | Edit        | A particular set of changes on a page at a point in time. It has an associated timestamp, user, and revision message attached. |
 | User           |              | Account     | A user account on Wikijump, with its own username, avatar, and credentials. |
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -1,0 +1,25 @@
+## Wikijump Glossary
+
+There are a number of concepts within Wikidot, which can take on a number of different names in casual usage.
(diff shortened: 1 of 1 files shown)


rossjrw (approved):
This is fine for now but at some point I'd like to clear up what "official term" means - does it mean that's the term we use in documentation / the code? Should it be forced into being different from what's currently generally used now?

emmiegit:
For casual conversations the name doesn't matter too much, but when writing new code, using the API, or discussing internals it's important to be precise and avoid confusion.

So what I mean by "official" is "the term you should use to avoid ambiguity". If you have a suggestion for a better way of describing this I'm open.

rossjrw:
I'm thinking of the way Discord distinguishes between "server" and "guild". If we have one term used in user-oriented documentation and another for developer-oriented documentation and code, that's fine by me, but I want to be specific about what's what and I want the one that's already in common use to be the one that ends up in user documentation. "Official" is ambiguous here

E.g. I don't know off the top of my head how the codebase refers to rating internally, but nobody calls it "score"

emmiegit:
Yeah I was thinking of the issue with discord server (guild) and physical server too. I think you're right changing "official" to a descriptor that better relays its use regarding code/API makes sense.

Regarding an individual case, I'm fine changing "score", that comes from the fact that there will be differing means of scoring pages, which seems to be different from the idea of a rating? Do you have thoughts on this?

rossjrw:
Could you clarify what you mean by different means of scoring pages? Are we talking eventual support for e.g. Wilson scoring?

emmiegit:
Yes, where the rating for a page is different than `ups - down` and uses some scoring algorithm. I suggested "score" because I thought it would reflect this better than "rating".

rossjrw:
That sounds good, then, with rating being a naive scoring algorithm and therefore being a type of score?

docs/glossary.md
```diff
@@ -0,0 +1,25 @@
...
+| Page           |              | Article     | The main data unit of Wikidot, containing wikitext and metadata, and existing at exactly one particular URL at a time. |
+| Slug           | UNIX name    | Fullname    | The "file" portion of a page's URL, always in [Wikidot normal form](https://scuttle.atlassian.net/wiki/spaces/WD/pages/541655041/Wikidot+Normal+Form). |
+| Normalization  | `toUnixName`, `unixify` |  | The process of converting a string into Wikidot normal form. |
+| Page category  |              |             | A particular namespace a page may exist in. Denoted by the part of a slug before the `:`. If a colon is absent, then the page is in category `_default`. |
+| Wikitext       | Source       | Wikidot code, ftml | The textual representation of the contents of a page. |
+| Vote           |              |             | An individual vote cast on a page, along with its timestamp and user information. |
+| Score          | Rating       |             | The calculated result of all the votes presently active on a page. This is **not** necessarily the same as `ups - downs`. |
```
rossjrw:
```suggestion
| Score          | Rating       |             | The calculated result of all the votes presently active on a page, per the site's chosen scoring algorithm. |
| Rating         | Rating       |             | Wikidot's simple scoring algorithm, equivalent to the sum of `upvotes - downvotes`. |
```

emmiegit:
Yeah, I think that makes sense.

emmiegit:
thanks @rossjrw @stormbreath

PR #123, opened 2021-03-09
Add support for trimmed anchor blocks

emmiegit:
Adds `[[a_]]` / `[[anchor_]]` which exhibits the behavior of `[[span_]]` but for the anchor block. Also adds AST tests to verify this behavior. and updates Cargo dependencies.


ftml/src/parsing/rule/impls/block/blocks/anchor.rs
```diff
@@ -23,7 +23,7 @@ use crate::tree::AnchorTarget;
 
 pub const BLOCK_ANCHOR: BlockRule = BlockRule {
     name: "block-anchor",
-    accepts_names: &["a", "anchor"],
+    accepts_names: &["a", "a_", "anchor", "anchor_"],
     accepts_special: true,
```
rossjrw, line 27: Would it be a good idea / more consistent to, in the future, either:

1. have trailing-underscore tags be some sort of static configuration option e.g. `accepts_modifier`?
2. have leading-asterisk tags be generically declared in `accepts_names`?

If the former the two should probably be named either to make the most sense in their individual functions or to describe what effect they have on the syntax

ftml/test/anchor2-alias-special.json
```diff
@@ -0,0 +1,43 @@
...
+                    "type": "paragraph",
+                    "elements": [
+                        {
+                            "element": "anchor",
+                            "data": {
+                                "attributes": {
+                                    "href": "http://example.com",
+                                    "style": "color: green;"
+                                },
+                                "target": "new-tab",
```
rossjrw, line 17: (Out of scope for this PR) Should this be `_blank` or is that implementation left to the API consumer?

ftml/test/anchor2-alias-special.json
```diff
@@ -0,0 +1,43 @@
+{
+    "input": "[[*anchor_ href=\"http://example.com\" style=\"color: green;\"]]My link[[/anchor]]",
+    "tree": {
+        "elements": [
+            {
+                "element": "container",
+                "data": {
+                    "type": "paragraph",
```
rossjrw, line 8: Noting that Wikidot does not create a wrapping paragraph here, is the divergence intentional?

It would be good to have a test containing two consecutive trailing-underscore links to verify that their combined behaviour is desirable - when separated by 2 newlines, are they placed in the same paragraph or in separate paragraphs?

ftml/test/anchor2-end.json
```diff
@@ -0,0 +1,42 @@
+{
+    "input": "[[a_ href=\"/some-page\"]]My link[[/a_]]",
```
rossjrw: I recall agreeing that we wouldn't implement this (#9)

ftml/test/anchor2-alias-wrap.json
```diff
@@ -0,0 +1,48 @@
...
+            {
+                "element": "container",
+                "data": {
+                    "type": "paragraph",
+                    "elements": [
+                        {
+                            "element": "anchor",
+                            "data": {
+                                "attributes": {},
+                                "target": "same",
```
rossjrw, line 14: (Out of scope for this PR) Does FTML distinguish between "same as specified by the syntax in this link" and "same by default"? If not, how would the API consumer know which links should override e.g. a [`<base target="_blank">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) on the webpage?

rossjrw (approved):
There should be a test with 2 consecutive newlines, to verify that that parser does not (or does?) create internal paragraphs.

On Wikidot:
* `[[a]]\nOne\n\nTwo\n[[/a]]` &rarr; `<p><a href=""><br>One</a></p><p><a href="">Two</a></p>`
* `[[a_]]\nOne\n\nTwo\n[[/a]]` &rarr; `<a href="">OneTwo</a>`

ftml/src/parsing/rule/impls/block/blocks/anchor.rs
emmiegit: 1. Hmm, I suppose that's possible. The way it works for "special" is that `[[*` is a distinct token from `[[`, so it's just a field when block checking happens. So I guess this means the field does the underscore checking itself, before handing off to the block rule. I will think on this.
2. As mentioned above, this would require some changes to how rules are parsed. Additionally it causes issues if there's a space after `[[*` and before the name (though apparently Wikidot itself does not support this behavior).

ftml/test/anchor2-alias-special.json
emmiegit: `AnchorTarget` is defined at https://github.com/scpwiki/wikijump/blob/develop/ftml/src/tree/anchor.rs#L24-L45. The HTML-equivalent fields are listed in the documentation there, however for the purposes of an AST we should be descriptive and avoid HTML-specifics when possible.

This field in the JSON is a straight mapping to the enum, it's just saying `AnchorTarget::NewTab`.

ftml/test/anchor2-alias-special.json
emmiegit: Yeah...

So the paragraph logic is _closer_ now than it's ever been, but there are still several deficiences. The issue [FTML-73](https://scuttle.atlassian.net/browse/FTML-73) exists to ensure we iron out these issues.

For now, as long as the output is consistent with other rules, we'll treat it as valid.

ftml/test/anchor2-end.json
emmiegit: PR #9 was about adding it to the Wikidot regex, discussing about its unnecessary complexity. Also the issue [WJ-147](https://scuttle.atlassian.net/browse/WJ-147) ("Multiple ways of typing something") was created addressing a similar issue.

This test only verifies existing behavior. Block rules permit any valid name to terminate the ending block. If we implement the first suggestion from [your reply here](https://github.com/scpwiki/wikijump/pull/123#discussion_r589998768) then we could definitely make it so that underscores like this are not accepted.

ftml/test/anchor2-alias-wrap.json
emmiegit: I was not aware that `<base>` was a tag before you mentioned this. I'm not really sure what you mean by

> "same as specified by the syntax in this link" and "same by default"

So I'm not exactly certain what the problem being identified here is.

rossjrw: On a page with `<base target="_blank">`, I would expect all links to, by default, open in a new tab. (This could happen with a browser extension that makes the default link target _blank, possibly)

If I take my Wikitext for that page's content, send it to FTML, and FTML sends back the AST for me to render, and a given link in the tree has `target: "same"`, I don't know whether or not to ignore that (because it's the default, and all links on this page should open in a new tab), or obey that (because this link clearly wants to override the base behaviour).

ftml/test/anchor2-end.json
rossjrw: Fair enough. If it's too much effort to implement, I'd be okay with this provided that it's not documented anywhere. I'm never certain what the best balance of flexibility vs strictness is. I'd prefer to err on the strict side but this is pretty harmless.

emmiegit: Okay makes sense. Yeah it was less of an intentional addition and more about permissiveness. I think I'll keep the behavior but only have the one test for it, and not really document it.

ftml/test/anchor2-alias-wrap.json
emmiegit: Interesting, this changes my understanding of links.

Now that there's a very clear case for `Option<AnchorTarget>` as the field type, I will be adjusting how links handle this and returning `None` (JSON `null`) when it's not explicitly set.

emmiegit:
thanks @rossjrw

PR #124, opened 2021-03-10
Set AnchorTarget optionally if it's not overwriting

emmiegit:
As [discussed here](https://github.com/scpwiki/wikijump/pull/123#discussion_r590018307), we want a renderer to know whether to override an anchor target or just leave it to carry the default. We accomplish this by making the field `Option<AnchorTarget>` instead of `AnchorTarget`. This PR also edits all AST tests to comply with the change.


rossjrw (approved):

emmiegit:
Thanks @rossjrw

PR #128, opened 2021-03-16
Add component and theme to glossary

emmiegit:
Thanks to @rossjrw, apparently Wikidot conflates components and themes and instead has a strange name for them:

> The CSS module lets you insert CSS code into a wiki page. This is particularly useful for cross-site include (CSI) packages that need to use custom styling for their code. When you use the CSS module in a CSI, that CSS code will be included in all pages that use the CSI.

We won't be using this terminology in Wikijump, but it's good to know what Wikidot calls them.


rossjrw (approved):
Would be good to also define what "included" means in this context

emmiegit:
@rossjrw I've intended for these definitions to be basically one line. I feel like "include" in this context is clear (i.e. what happens when you use `[[include]]`).
Unless I'm misunderstanding what you mean here?

rossjrw:
Aye, that's what I mean. If you feel "include" is unambiguous, I'm cool with that. It is in the syntax after all, rather than a makeshift term.

emmiegit:
I think it's fine for now. We could add "include" as its own entry to this glossary though.

emmiegit:
thanks @rossjrw

PR #130, opened 2021-03-20
[FTML-85] Filter attributes based on safety.

emmiegit:
Those with Confluence access can see https://scuttle.atlassian.net/wiki/spaces/WD/pages/1030782977/Allowed+Attributes+in+Wikitext, which describes which attributes are permitted and which are filtered out. The attribute list was taken from https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes.

The goal is to prevent security issues through use of attributes that should not be set by end users.


rossjrw (approved):
This seems fine.

I wonder if it would be possible to extract the list of safe attributes into a configuration file?

Is there a test to verify that data-* attributes are passed through?

emmiegit:
@rossjrw yeah `span-style.json`, `span2-style.json`, and some others test `data-xxx`.

As for a configuration file, yeah that's possible, but it would require changing how ftml loads to involving an initialization step where constants like the attribute list are passed in. So I think until there's a direct need for dynamic attribute filtering, I'll leave it as-is.

emmiegit:
thanks @rossjrw

PR #131, opened 2021-03-21
[WJ-96] Remove outdated magic files

emmiegit:
Wikidot uses these magic files pulled from an old version of `file(1)` for some reason. These old files prevent recognition of all SVGs, whereas modern versions of `file(1)` have no issue.


rossjrw:
I'd like to mark this as draft until we know for sure that these files aren't used anywhere

emmiegit:
Sure.

emmiegit:
Note that this is on ice until Wikijump is able to upload files.

emmiegit:
I'm reopening this.

Given the file [`web/php/Utils/FileMime.php`](https://github.com/scpwiki/wikijump/blob/develop/web/php/Utils/FileMime.php#L16-L19), we can see that Wikidot gets the MIME type by just invoking `file`. It packages these old magic files for "completeness" I guess, but the issue with [WJ-96](https://scuttle.atlassian.net/browse/WJ-96) is fundamentally that these magic files are extremely out-of-date. We have already confirmed that a more recent version of the `file` utility will recognize SVGs just fine.

Since these files are not directly used by Wikijump, and are outdated as mentioned, it is thus safe to remove them entirely.

rossjrw (approved):

emmiegit:
thanks @rossjrw

PR #137, opened 2021-03-24
Thanks Windows. Folders now have correct capitalization.

danieltharp:


emmiegit:
It's cases like this where I wish git could coalesce multiple file moves into a single directory move.

emmiegit (approved):

PR #145, opened 2021-03-25
Copy deployment fixes from dev to local

emmiegit:
Bluesoul has made a number of fixes to our dev deployment. Since we cannot use symlinks for a lot of our per-environment deployment files, this PR copies those changes to local.

Here is the local environment running on my machine:
image

This also adds a small README file to `install/legacy` briefly explaining what it is.


danieltharp (approved):

emmiegit:
thanks @pxdnbluesoul

PR #147, opened 2021-03-25
Use getNamespacedClassFromPath() everywhere

danieltharp:
This PR audits all uses of the string `new $class` as there's a lot of repetitive code in the base that uses that. It replaces a call to various `getThingClassName();` calls that assume no namespacing with the new `LegacyTools::getNamespacedClassFromPath()` method call. It also fixes a call to a nonexistent property that should've been a method call in `AjaxModuleWebFlowController.php`

### DIFF
--- a/web/lib/ozoneframework/php/core/Form.php
+++ b/web/lib/ozoneframework/php/core/Form.php
@@ -176,7 +176,8 @@ class Form {
 			if($this->validatorName == null) {
 				$this -> validatorName = "BaseFormValidator";
 			}
-
+			// As far as I can tell there's only the BaseFormValidator but I might be wrong.
+            $this->validatorName = "Ozone\Framework\\".$this->validatorName;
 			foreach($chain as $rule){
 				$this->isValidArray["$fieldName"]=true;
 				$ruleName = $rule['name'];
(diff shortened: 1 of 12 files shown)


danieltharp:
This audit has been expanded to all uses of `new $` which caught a few additional things. Some libraries like text_wiki and smarty still live in the root namespace and I haven't extended this to them yet.

web/lib/ozoneframework/php/core/Form.php
```diff
@@ -176,7 +176,8 @@ public function validate($fieldName = null){
 			if($this->validatorName == null) {
 				$this -> validatorName = "BaseFormValidator";
 			}
-
+			// As far as I can tell there's only the BaseFormValidator but I might be wrong.
+            $this->validatorName = "Ozone\Framework\\".$this->validatorName;
```
emmiegit, line 180: Is `BaseFormValidator` something essential we should know about / keep an eye on?

emmiegit (approved):
:+1:

web/lib/ozoneframework/php/core/Form.php
danieltharp: This looks like it's only relevant with regards to Data Forms. We may have to explore what validation, if any, is missing. But I'd be far more likely to let Laravel's robust validation engine do anything else we need.

emmiegit: Sounds good to me. And data forms, while they should be carried over, are not high on my list of priorities.

PR #148, opened 2021-03-26
Remove all references to Wikidot\DB\OzoneSession

danieltharp:
Since both `OzoneSession` classes extend the same base class, I'm going to be removing `Wikidot\DB\OzoneSession` in a future version. This PR should eliminate any remaining references to it in the code.


emmiegit (approved):
🗑️

PR #149, opened 2021-03-26
Pass along username in login form when previously logged in

danieltharp:
This fixes that bug where the username is visible but returns invalid username or password when trying to log in.


web/templates/modules/Login/LoginModule.tpl
```diff
@@ -5,6 +5,7 @@
 
 	{if $user}
 		{t}Hello{/t}, <span style="font-size:130%; font-weight: bold">{$user->getNickName()|escape}</span>
+        <input type="hidden" name="name" value="{$user->getNickName()|escape}" id="login-form-name"/>
```
ammongit, line 8: The original version didn't have an input field for the username?

emmiegit (approved):

PR #151, opened 2021-03-27
[FTML-86] Add documentation for modules.

emmiegit:
Finishes [FTML-86](https://scuttle.atlassian.net/browse/FTML-86) by adding documentation for each currently-implemented module.-


rossjrw (approved):

rossjrw:
How feasible do you think it would be in the long-term to derive the information for modules from the PHP?

emmiegit:
Oh, good point. I guess we could think of an OpenAPI or similar spec both services can derive from?

emmiegit:
thanks @rossjrw

PR #152, opened 2021-03-28
Minor fixes while investigating bottom bar issues

emmiegit:
This renames some of the missed modules in the database seeds, and removes a dead if branch.


rossjrw (approved):

emmiegit:
thanks @rossjrw

PR #154, opened 2021-03-29
Small fix and QOL improvement

danieltharp:
This PR has two little things outside the scope of my Laravel work. One is having Wikidot's OzoneSession inherit the Ozone Framework class of the same name to work around some limitations in how namespacing has been introduced to the legacy codebase. Without this, long-lived sessions will cause the user to only see PHP fatal errors rather than the site.

It also changes the local docker-compose environment to allow for connections to the postgres container from your client or IDE of choice.


emmiegit (approved):
🐘

PR #155, opened 2021-03-30
[WJ-3] Change password length limits to min 8, max 1024.

emmiegit:
As detailed in [WJ-3](https://scuttle.atlassian.net/browse/WJ-3), a maximum of 20 is arbitrarily short and limits password effectiveness. This PR raises the upper bound to 1024 (not so large as to enable denial-of-service attacks, but more than long enough for very secure passwords), and raises the lower bound to 8. This will bring Wikijump in line with FIPS recommendations regarding password lengths.

This PR also fixes the various inconsistent password input field maximum lengths set throughout the codebase, standardizing on the maximum length of 1024.

### DIFF
--- a/web/php/Actions/CreateAccountAction.php
+++ b/web/php/Actions/CreateAccountAction.php
@@ -143,8 +143,8 @@ class CreateAccountAction extends SmartyAction
         // check password
         if (strlen8($password)<8) {
             $errors['password'] = _("Please provide a password at least 8 characters long.");
-        } elseif (strlen8($password)>20) {
-            $errors['password'] = _("Password should not be longer than 20 characters.");
+        } elseif (strlen8($password)>1024) {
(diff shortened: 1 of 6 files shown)


danieltharp (approved):
Works for me. For those playing at home, the actual string that's stored in the DB is a fixed length, so while a 1024-character password does increase entropy (assuming it's sufficiently random), the security gains are marginal over, say, a 32-character password. Collision is your biggest risk in either case.

emmiegit:
The actual string being the password hash, right? I didn't realize the chance of collision was relevant for passwords longer than ~30 chars.

danieltharp:
The risk is astronomically low, but not zero, and the most likely method of cracking a sufficiently random `bcrypt`'ed password.

emmiegit:
thanks @pxdnbluesoul

PR #157, opened 2021-03-30
[WJ-428] Remove the Zend framework

emmiegit:
The entire vendored framework is present in code, exclusively for search (which we're replacing) and pingbacks (shitty and unused).

This PR removes it.

PR #158, opened 2021-03-30
[WJ-431] Remove pingbacks

emmiegit:
This PR is a follow-up to #157 which removed the Zend framework. We're not using pingbacks, so this removes the associated code from the repository:

* The pingback settings columns in the database
* The code that gets/sets those columns
* Some miscellaneous files related to pingbacks


danieltharp (approved):

emmiegit:
thanks @pxdnbluesoul

PR #159, opened 2021-03-30
Have OzoneSessionPeer shadow Wikidot class

danieltharp:
This is a dumb but seemingly necessary change to deal with an `OzoneSessionPeer` instance being dynamically created in the wrong namespace.


emmiegit:
Incredible.

emmiegit (approved):

PR #161, opened 2021-03-31
Forbid api domains from registration

danieltharp:
We may or may not use the e.g., `api` or `api01` subdomain but there's very little legitimate reason to allow it to be registered by a user.

### DIFF
--- a/web/conf/forbidden_site_names.conf
+++ b/web/conf/forbidden_site_names.conf
@@ -44,3 +44,4 @@
 /^wdupload$/
 /^wjupload$/
 /^file[s]?$/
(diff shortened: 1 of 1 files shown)


emmiegit:
Discussing it, I think we mostly considered having `slug.wikijump.com/api--v0/method` as the scheme, similar to `files--common` and the like.

But I agree, there's no reason somebody should be registering `api.wikijump.com`. It shouldn't even really be its own site, we can just have an `api:` category on the `www` / main wikijump site.

emmiegit (approved):

PR #164, opened 2021-04-02
Add later block.

emmiegit:
Adds an extremely important block to ftml. This block, `[[later]]`, will allow writers to designate that this section of text is incomplete and will be completed at a later date. Of course, for its efficacy as a special-use block to make sense, it must be undocumented within the formal list of blocks and their arguments.


danieltharp (approved):

emmiegit:
thanks @pxdnbluesoul

PR #165, opened 2021-04-02
Move forbidden names from configuration files

emmiegit:
Presently, we have both inline default configuration lists and the list in `*.conf` files. This is confusing and duplicates some code.

This PR resolves this by moving the configured list into separate namespaced PHP files to be read from when those lists are needed.


danieltharp:
Honestly I'd rather do the opposite, keep the static arrays in namespaced php code and remove the .conf files and make sure the references in the codebase are to that method rather than these loose files.

emmiegit:
Huh interesting. In general I've preferred static configuration files over code. Do you have any special uses (or issues) in mind, or is this how you prefer to structure things?

emmiegit:
Okay going to move the conf files as namespaced PHP. I don't think `Wikidot\Utils` makes sense, going to move to `Wikidot\Config` unless you are opposed.

emmiegit:
@pxdnbluesoul Moved to a separate, namespaced PHP file, which the other usages here read from. I ran through the flow locally and made sure that an invalid name was failing where I expected it to.

danieltharp:
> 
> 
> Huh interesting. In general I've preferred static configuration files over code. Do you have any special uses (or issues) in mind, or is this how you prefer to structure things?

With PHP-FPM those arrays will persist in RAM, `file_get_contents()` will hit the disk every time, which is actually across the network, these servers have no onboard storage. It also has to do the translation to an array, just higher overhead in a number of ways.

danieltharp (approved):

emmiegit:
Oh I see, yeah. Loading that every time is certainly inefficient.

I guess I'm so used to Wikidot I didn't blink when I saw that and was thinking of other issues.

emmiegit:
thanks @pxdnbluesoul

PR #163, opened 2021-04-01
[WJ-429] Rename CreateAccount2 → CreateAccount

emmiegit:
With the `CreateAccount/` directory removed (see #162), we can rename `CreateAccount2` to remove the unnecessary 2.

PR #166, opened 2021-04-02
[WJ-429] Rename CreateAccount modules to have StepX.

emmiegit:
Follow-up to #163, adding `Step1`, `Step2`, etc to modules to make it clear what the numbering is for.


danieltharp (approved):

emmiegit:
thanks @pxdnbluesoul

PR #167, opened 2021-04-02
Nginx & php-fpm improvements

danieltharp:
This PR does a couple of things to help get us closer to using Laravel.

- Adds the needed `pdo_pgsql` extension to the php-fpm container.
- Adds a note to the nginx config to allow larger buffers, useful for dev-level stack traces to come through the web frontend on errors.
- Updates the env.example file (which is used on dev deploys) to correctly reference the database and memcached containers.
- Updates the local/dev dockerfile for improved functionality with WSL2.


install/aws/dev/docker/php-fpm/etc/nginx/sites-available/wikijump
```diff
@@ -19,42 +19,39 @@
...
+            try_files $uri $uri/ /index.php?$args;
+            # Fallback Route
+        }
+
+        location ~ \.php$ {
+                    include snippets/fastcgi-php.conf;
+                    fastcgi_pass 127.0.0.1:9000;
+        }
+
+        rewrite ^/common--(.+)$ /files--common/$1 break;
```
emmiegit, line 54: What is the origin of this rule? Aren't there kinds of `common--xxx` paths other than files?

web/.env.example
```diff
@@ -1,28 +1,27 @@
-APP_NAME=Laravel
+APP_NAME=Wikijump
 APP_ENV=local
 APP_KEY=
 APP_DEBUG=true
-APP_URL=http://localhost
+APP_URL=https://www.wikijump.dev/wikijump
+ASSET_URL=https://www.wikijump.dev/wikijump/assets
```
emmiegit, line 6: Noting this prohibits the effective creation of a page with slug `wikijump`. We will need to create a psuedo-page in that location similar to what will be needed for `_admin`.

emmiegit (approved):

install/aws/dev/docker/php-fpm/etc/nginx/sites-available/wikijump
danieltharp: They're all in the `web/files--common` folder, under various subfolders. This is not modified from the original, only moved around.

web/.env.example
danieltharp: This is not intended to be a long-term solution in all honesty, but I haven't decided on which way to take it. But this way does at least allow for separation of the two codebases for the time being.

emmiegit: I mostly bring it up since generally special routes get some kind of `--` in it (e.g. `files--common`, `api--v1`). This is fine for now though.

danieltharp: If we wish to move it around later, only this file and the `install/*docker/php-fpm/etc/nginx/sites-available/wikijump` need to be changed, and I can't see any reason why adding `--` to the path would be problematic, but I'm also not sure if I want to standardize on that as the design pattern we want to use.

emmiegit: I'm definitely open to alternatives to `--` being used, I've just used it for now since it's what Wikidot uses, and honestly it's a relatively elegant solution to avoid path collision with arbitrary page names.

danieltharp: > 
> 
> Noting this prohibits the effective creation of a page with slug `wikijump`. We will need to create a psuedo-page in that location similar to what will be needed for `_admin`.

Also just to clarify, even if a page was created at `wikijump`, nginx is sending it to Laravel before the Wikidot code is even aware there's a request. So you'd have essentially an unreachable page for now.

emmiegit: Right, but this has some side effects such as any local links pointing to the page appearing as orange, even though there is a valid path there. This phenomenon is documented in [WJ-331](https://scuttle.atlassian.net/browse/WJ-331).

PR #170, opened 2021-04-07
Cleanup of many unused and broken classes and methods

danieltharp:
This PR is mostly centered around removing class files that extend classes that don't actually exist in the codebase, a surprising amount of files. There are also a number of updates to various PHPDoc blocks to fix code inspectors yelling about mismatched data types. There are a few missing libraries I was able to track down that are not available in Composer and included them. Lastly, there are a few minor bugfixes, mostly around adding missing use statements or fixing small typos and the like.


web/DEADCODE.md
```diff
@@ -129,3 +129,10 @@ If possible, add to this log in the same commit in which the code is removed.
...
+* Where it was: As listed above.
```
emmiegit: So there's a link to the version of the repo when these files still existed:
```suggestion
* Where it was: As listed above, see [web/](https://github.com/scpwiki/wikijump/tree/64b94cda1ff4e941da45621f0255bba19adae4ee/web) directory.
```

web/lib/date/Date.php
```diff
@@ -0,0 +1,1465 @@
+<?php
```
emmiegit, line 1: General question: when will be able to break vendored code like this out of the repo?

web/lib/text_wiki/Text/Wiki/Render/Xhtml/Code.php
```diff
@@ -62,6 +62,13 @@ function token($options)
...
 		$text = trim($text);
 
+		/**
+         *  Here lied the text_highlighter factory and invocation.
+         *  We need to replace it with something as the library was broken from the beginning.
+         *  Whatever we use needs to take a `type` param and Do Highlighting To It.
+         *  I'm leaving an example and the else statement in as a fallback.
+         */
+
 //        if ($type == 'php') {
```
emmiegit, line 72: Could we not delete this other commented code too?

emmiegit (approved):
:wastebasket: :heavy_check_mark:

web/lib/date/Date.php
danieltharp: I had done some of that previously by breaking them out into managed Composer packages, but that requires an independent git repo for each one. When we were working on Text_Wiki that ended up being a bother and would generally slow down things since we moved to a monorepo.

We will also need to namespace every call to one of these libraries when we do that, which won't be bad for most of them but for Text_Wiki will be a large-ish task.

web/lib/text_wiki/Text/Wiki/Render/Xhtml/Code.php
danieltharp: I wanted to leave it in for now as a guide on what one implementation looked like, when it's replaced and working I'll remove this code as well.

PR #173, opened 2021-04-12
Add method for converting SyntaxTree into an owned form

emmiegit:
This adds a method `.to_owned()` (different from the trait [`ToOwned`](https://doc.rust-lang.org/std/borrow/trait.ToOwned.html)) which clones all the `Cow`s and owned fields within the structure to ensure the lifetime is `'static` (because it's not borrowing anything, so the object can live indefinitely).

This is necessary for ftml-wasm, to clone the whole structure to make it suitable for passing into JS. Presently it uses a hack: it serializes and deserializes the structure, which is extremely inefficient, even if it uses less lines of code.

### DIFF
--- a/ftml/src/tree/attribute/mod.rs
+++ b/ftml/src/tree/attribute/mod.rs
@@ -20,6 +20,7 @@
 
 mod safe;
 
+use super::clone::string_to_owned;
 use std::borrow::Cow;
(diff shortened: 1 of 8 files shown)


rossjrw (approved):

emmiegit:
thanks @rossjrw

PR #176, opened 2021-04-13
WJ-176: Replace gender with pronouns

danieltharp:
This PR is primarily around replacing the `gender` field in the database (`char` type) with a `pronouns` field (`varchar(30)` type) and updating references to it in the templates and DB seed. It also includes a regression and feature test.

It also fixes some configuration around phpunit and updates a gitignore file to make it...work.

### DIFF
--- a/install/local/dev/postgres/seed.sql
+++ b/install/local/dev/postgres/seed.sql
@@ -2850,20 +2850,15 @@ ALTER SEQUENCE public.private_user_block_block_id_seq OWNED BY public.private_us
 CREATE TABLE public.profile (
     user_id integer NOT NULL,
     real_name character varying(70),
-    gender character(1),
+    pronouns character varying(30),
     birthday_day integer,
     birthday_month integer,
     birthday_year integer,
     about text,
     location character varying(70),
     website character varying(100),
-    im_aim character varying(100),
-    im_gadu_gadu character varying(100),
-    im_google_talk character varying(100),
     im_icq character varying(100),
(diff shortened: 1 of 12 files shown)


install/local/dev/postgres/seed.sql
```diff
@@ -2850,20 +2850,15 @@ ALTER SEQUENCE public.private_user_block_block_id_seq OWNED BY public.private_us
...
     birthday_month integer,
     birthday_year integer,
     about text,
     location character varying(70),
     website character varying(100),
-    im_aim character varying(100),
-    im_gadu_gadu character varying(100),
-    im_google_talk character varying(100),
     im_icq character varying(100),
     im_jabber character varying(100),
```
emmiegit, line 2861: Honestly, I would cut all of them. I doubt the number of people active on Wikidot who use either of these services is more than like, 3 people.

danieltharp: All I wanted to do with this one was prove out that stuff can be removed in the seed and Ozone ORM and have it work as expected, given known useless fields. We had a larger conversation about replacing this with a morphable `profile_fields` table that I would roll further changes into.

emmiegit: Ah interesting, okay. And yeah I agree long term that's what we'll be doing for profiles.

danieltharp:
Marking this as ready as my actual browser testing works as expected with getting and setting these values.

emmiegit (approved):

PR #179, opened 2021-04-13
Fix profile website replacer.

emmiegit:
The `{$HTTP_SCHEMA}` inside the `replace` field wasn't valid, so I added a regex which captures both HTTP and HTTPS links.

This syntax issue was likely originally introduced in the big HTTP search and replace job.

Before:

image

After:

image


danieltharp (approved):

emmiegit:
thanks @pxdnbluesoul

danieltharp:
Just noting for posterity that I looked through the other usages of `{$HTTP_SCHEMA}` in the templates folder and didn't find anything else in a `replace` directive as it was here, so I think we're good.

PR #180, opened 2021-04-13
[FTML-87] Add support for HTML entities

emmiegit:
This PR adds the `[[char]]` block, which allows use of HTML entities. See [FTML-87](https://scuttle.atlassian.net/browse/FTML-87) for more information.

It supports the form both with and without `&..;` (so for instance, `[[char amp]]` and `[[char &amp;]]` are both valid), and the hexadecimal and decimal entities (e.g. `&#100;`, `&#xff;`)

The outputted element is still `Element::Text`. Actual escaping (for instance, forming an HTML entity in the final HTML output) is up to the escaping mechanism.

### DIFF
--- a/ftml/src/parsing/rule/impls/block/blocks/char.rs
+++ b/ftml/src/parsing/rule/impls/block/blocks/char.rs
@@ -105,6 +105,45 @@ fn find_entity(entity: &str) -> Option<Cow<str>> {
     None
 }
 
+/// Gets the appropriate character from the number specified in the string.
+///
+/// Using the passed radix, it gets the integer value, then finds the appropriate
+/// character, if one exists.
+///
+/// Then converts the character into a string with only that value.
+fn get_char(value: &str, radix: u32) -> Option<Cow<str>> {
+    let codepoint = match u32::from_str_radix(value, radix) {
+        Ok(codepoint) => codepoint,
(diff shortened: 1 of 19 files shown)


danieltharp (approved):

ftml/src/parsing/rule/impls/block/blocks/char.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    Some(Cow::Owned(ch.to_string()))
+}
+
+/// If a string starts with `&` and ends with `;`, those are removed.
+/// First trims the string.
+fn strip_entity(s: &str) -> &str {
+    let s = s.trim();
+
+    if s.starts_with('&') && s.ends_with(';') {
```
rossjrw: Is there any situation where the user would expect `&copy` to do something different to `&copy;`? If not, could at least the former be permitted?

rossjrw (approved):

ftml/src/parsing/rule/impls/block/blocks/char.rs
emmiegit: This is true, I think we can strip these separately. I'll adjust.

emmiegit:
thanks @pxdnbluesoul @rossjrw

PR #181, opened 2021-04-13
[WJ-440] Client/frontend monorepo design

Monkatraz:
Ready for review! This adds a Lerna, Estrella (esbuild), and Snowpack powered monorepo for handling frontend/client packages. It is intended to be very easy to use all-around and tries to avoid the pitfalls of large monorepos, namely performance.

- [x] Use Prettier directly instead of an eslint plugin (will have to make it ignore a bunch of files)
- [x] Smooth out Snowpack dev mode process
- [x] Document exactly how the build script works
- [x] Consider merging the test builder with the module builder
- [x] Create a web package template
- [x] Create some more detailed guides on how to make a package
- [x] Scream violently as you discover that https://github.com/folke/ultra-runner exists and can massively reduce the repo's complexity


danieltharp:
I'm busy until late this afternoon but I'd like to figure out the github actions workflow for building changes to this folder into the nginx container and having the dockerfile updated accordingly.

client/.npmrc
```diff
@@ -0,0 +1,2 @@
+# TODO: turn this off at some point
+legacy-peer-deps = true
```
emmiegit, line 2: What does this do?

client/build.js
```diff
@@ -0,0 +1,73 @@
+const { build, cliopts, basename, stdoutStyle: styl, fmtDuration } = require('estrella')
+const { nodeExternalsPlugin } = require('esbuild-node-externals')
+const { readdirSync, realpathSync } = require('fs')
+const { performance } = require('perf_hooks')
+
+const dev = !!cliopts.watch
```
emmiegit: I prefer `Boolean(..)` over `!!..` as more readable.

client/docs/package-template/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "package-name",
+  "license": "agpl-3.0-or-later",
+  "description": "Short package description.",
+  "version": "1.0.0",
+  "keywords": [
+    "wikijump"
+  ],
+  "scripts": {
+    "clean": "trash dist .tsbuildinfo",
```
emmiegit: `trash` is not a universal binary. This should be `rm -rf dist .tsbuildinfo` here.

**Edit:** to be clear, `/usr/bin/trash` is present if you have `trash-cli` installed (at least that's what the package is on my system), but it's not a UNIX standard like `/usr/bin/rm` is, and it's definitely going to be absent on a minimal or non-desktop Linux environment.

client/docs/package-template/package.json
```diff
@@ -0,0 +1,26 @@
...
+  },
+  "type": "module",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/scpwiki/wikijump.git"
+  },
+  "bugs": {
+    "url": "https://github.com/scpwiki/wikijump/issues"
```
emmiegit: Issues are disabled, the link for requests/bugs from the public is https://scuttle.atlassian.net/servicedesk/customer/portal/2

client/docs/vscode-template/settings.json
```diff
@@ -0,0 +1,10 @@
+{
+  "files.exclude": {
+    "**/.npm": true,
+    "**/node_modules": true,
+    // "**/dist/": true,
+    // "**/.tsbuildinfo": true,
+    // "**/package-lock.json": true
+  },
```
emmiegit: Are you using a JSON extension here? Comments aren't accepted as standard JSON.

client/packages/cm-tarnation/package.json
```diff
@@ -0,0 +1,38 @@
...
+  },
+  "type": "module",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/scpwiki/wikijump.git"
+  },
+  "bugs": {
+    "url": "https://github.com/scpwiki/wikijump/issues"
```
emmiegit: Ditto

client/packages/cm-tarnation/src/buffer.ts
```diff
@@ -0,0 +1,333 @@
+import { search, SearchOpts } from 'wj-util'
```
emmiegit: Shouldn't eslint be adding terminating semicolons to these files? Like we discussed in chat, developers use whatever local style, then eslint normalizes it.

client/packages/wj-util/package.json
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "type": "module",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/scpwiki/wikijump.git"
+  },
+  "bugs": {
+    "url": "https://github.com/scpwiki/wikijump/issues"
```
emmiegit: Ditto

client/packages/wj-util/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "wj-util",
+  "license": "agpl-3.0-or-later",
+  "description": "Collection of common utility methods.",
+  "version": "1.0.0",
+  "keywords": [
+    "wikijump"
+  ],
+  "scripts": {
+    "clean": "trash dist .tsbuildinfo",
```
emmiegit: Ditto

client/tsconfig.json
```diff
@@ -0,0 +1,53 @@
...
+    "declarationMap": true,
+    "emitDeclarationOnly": true,
+    "preserveWatchOutput": true
+  },
+
+  // Packages must be added here in order for their types to be compiled.
+  // Unfortunately, there isn't a decent way to do this automatically.
+  // The performance benefits are worth it, regardless.
+  "references": [
+    { "path": "./packages/wj-util" },
```
emmiegit: I'm of the opinion that wikijump helper packages should be accessible through a vendor prefix `@wikijump`, so here `@wikijump/util`.

emmiegit (commented):
Looking over the build structure and mostly ignoring the code, I have some early comments, but overall this seems fine.

I'm going to try and see how this behaves on my local machine.

emmiegit:
This seems to install and build fine on my machine.

client/build.js
Monkatraz: This is governed by an ESLint rule, actually. We can make a change for that.

client/.npmrc
Monkatraz: NPM has "peer dependencies", or dependencies that are... for lack of a better word, tangential? The main purpose of them is for packages that are modules or extensions to existing ones, like a plugin for Webpack. Webpack would be the peer dependency in that instance.

Recent NPM versions began to automatically resolve and install peer dependencies, which causes issues when the dependencies have incompatible versions (not necessarily actually incompatible). I think the whole ecosystem is still adjusting to it, and I was getting some install errors and I didn't want CI or something to fail.

client/docs/package-template/package.json
Monkatraz: The magic of Lerna, baby! This is actually the `trash-cli` dependency from the root monorepo `package.json`. I made sure specifically _not_ to use native utilities so there wouldn't be Windows <-> Linux issues. Lerna when it bootstraps makes sure to symlink `bin` utilities so that all packages can call them. For example, all packages could call `tsc`.

client/docs/vscode-template/settings.json
Monkatraz: VSCode accepts comments in its configuration files. Not sure exactly what the rules are for that, but this isn't a syntax error.

client/packages/cm-tarnation/src/buffer.ts
Monkatraz: I'm leaving ESLint rules alone until we all get together to discuss it.

client/docs/package-template/package.json
emmiegit: Interesting, I was not aware of this. This seems fine then, since it's designed expressly to be multiplatform.

client/docs/vscode-template/settings.json
emmiegit: Oh I see this is VSCode specifically. This is fine then.

client/tsconfig.json
Monkatraz: If you mean aliases, I don't think in this instance we should do that as it adds some complications to module resolution. If you just mean scoping an NPM package, I think that would complicate the monorepo design a little bit? I can look into it.

emmiegit: Ultimately as long as we have reasonable and consistent naming for wikijump-native modules/packages this should be fine.

client/docs/package-template/package.json
Monkatraz: Resolved via [this commit.](https://github.com/scpwiki/wikijump/pull/181/commits/3d9e77348d620dfc75f3bc09bc840a80eeb47f81)

Monkatraz:
While this isn't done, a lot of the core functionality is. I'd be fine with reviews being left on it, although I should note the folllowing:
* The `cm-tarnation` package is just a "hard-hitting" test package, although it probably will officially be merged in at some point. It's used by the editor for syntax highlighting several languages. The `wj-util` package will stay though, because there should be at least _something_ in the repo.
* A "bundled packages" system still needs to be added, but it's the last on the list of things to be added.

Monkatraz:
Alright, initial bundling of a website using Snowpack is in. Both build and dev modes work, and Snowpack responds to modules being rebuilt.

However, Snowpack currently has a few bugs related to module resolution (go figure with Node) and spams the console a little bit in dev mode. However, Snowpack does fully recover and mount the missing modules.

Additionally, I haven't figured out a good way to handle the build and dev modes of a website package when you're in the root. Currently, the best way to run a web dev mode is to use two terminals/processes. There is a command to run both at the same time, but closing the process doesn't close the Snowpack server. I suspect this is a Lerna bug. For right now, it can be managed.

Monkatraz:
Can't request reviews (don't have access), so consider this my request for reviews from @rossjrw , @pxdnbluesoul , and @ammongit

client/.eslintrc.js
```diff
@@ -0,0 +1,211 @@
+function useDefault(type, rules) {
+  return Object.assign({}, ...rules.map(rule => ({ [rule]: type })))
```
emmiegit, line 2: n/b: Isn't the preferred way nowadays to use an object comprehension instead of `Object.assign()`? A linter at work complains at me for that.

client/.gitignore
```diff
@@ -0,0 +1,14 @@
+# NPM
+node_modules
+package-lock.json
+lerna-debug.log
+npm-debug.log
+
+# Editors
+.vscode
+.idea
```
emmiegit: We should port these to the root level `.gitignore` instead.

client/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+```
+$ npm run dev
+```
+
+Development mode will begin with a full build in order to ensure that dependency resolution succeeds. After that, a development process will be started and packages will be rebuilt every time a file of theirs changes.
+
+### Commands
+
+See [`useful-commands.md`.](docs/useful-commands.md)
```
emmiegit: nitpick: periods should be outside of the link

client/docs/modules.md
```diff
@@ -0,0 +1,39 @@
...
+- src/
+  - index.ts
+  - **/*.ts
+- tests/
+  - *.ts
+- package.json
+- tsconfig.json
+```
+
+The `src/` folder is compiled into a `dist/` folder when the module is built. Similarly, files in the `tests/` folder are compiled into the `tests/dist/` folder when the tests are built.
```
emmiegit, line 16: n/b: Is it `tests/dist/`, or does it go in `dist/tests/`?

client/docs/modules.md
```diff
@@ -0,0 +1,39 @@
...
+
+Other than the inheritance of development dependencies and tooling configurations, these modules act like fairly normal NPM packages. If you need to add a dependency to one, you can either navigate to the module and do a normal `npm i <dep>` command, or manually add it to the `package.json`. After you do that, you need to make sure to run the `npm run bootstrap` command at the root. This ensures that all the fragile symlinks get restored by Lerna.
```
emmiegit, line 18: Do you mean `npm i --save <dep>`?

client/docs/modules.md
```diff
@@ -0,0 +1,39 @@
...
+
+To improve performance of the monorepo, building and watching of modules isn't done per module, but instead globally at root. It is recommended that you just build the entire monorepo rather than trying to build individual packages. This is not an insane recommendation because module building is 100% incremental. If a module has not changed, it won't get rebuilt.
```
emmiegit: "insane recommendation" is a weird way of saying it, say something like "This is not a performance issue because ..."

client/docs/modules.md
```diff
@@ -0,0 +1,39 @@
...
+
+There is a template for making modules, and it can be found in the `templates/module-template` folder.
+
+Here is how to use that template:
+
+1. Create a folder in `modules/` that is the name of your package.
+  It isn't strictly required that it is the name of your package, but it gets confusing otherwise.
```
emmiegit, line 27: This seems more like a footnote. I don't think we should ever have a case where the name and directory mismatch, the documentation should just assert that they're the same.

client/docs/modules.md
```diff
@@ -0,0 +1,39 @@
...
+  If you want to build the module as a Node package instead of for the browser, prefix the folder name with `node-`.
+
+2. Copy the contents of `templates/module-template` into the new folder.
+
+3. Edit `package.json`.
+  There is a few properties you should edit. They are: `name`, `description`, and `version`. Everything else is fine being left alone.
```
emmiegit:
```suggestion
  There are a few properties you should edit. They are: `name`, `description`, and `version`. Everything else is fine being left alone.
```

emmiegit: Any recommendations on sub-package versioning? I imagine it would probably be easiest to use [calver](https://calver.org/) for all of them, since "1.0" isn't really meaningful and they're entirely dependent on being in the Wikijump ecosystem.

client/docs/useful-commands.md
```diff
@@ -0,0 +1,19 @@
+```
+$ npm run update         Install dependencies, clean packages, boostrap, then build.
```
emmiegit: :ghost: 	
```suggestion
$ npm run update         Install dependencies, clean packages, bootstrap, then build.
```

client/docs/useful-commands.md
```diff
@@ -0,0 +1,19 @@
...
+$ npm i                  Install the root's packages. Be sure to run 'npm run bootstrap'.
```
emmiegit:
```suggestion
$ npm i                  Install the root's packages. Be sure to run 'npm run bootstrap' first.
```
?

client/docs/useful-commands.md
```diff
@@ -0,0 +1,19 @@
...
+$ npm run boostrap       Link packages together, hoist shared dependencies.
```
emmiegit:
```suggestion
$ npm run bootstrap      Link packages together, hoist shared dependencies.
```

client/docs/useful-commands.md
```diff
@@ -0,0 +1,19 @@
...
+$ npm run clean          Cleanup packages node_modules and tsc build data.
```
emmiegit:
```suggestion
$ npm run clean          Clean up packages node_modules and tsc build data.
```

client/docs/useful-commands.md
```diff
@@ -0,0 +1,19 @@
...
+$ npm run build          Build all packages.
+$ npm run dev            Start a dev. server for rebuilding packages when they change.
```
emmiegit:
```suggestion
$ npm run dev            Start a dev server for rebuilding packages when they change.
```

client/docs/websites.md
```diff
@@ -0,0 +1,34 @@
+# Monorepo: Websites
+
+This monorepo has special support for website packages, in the `web/` folder. These are quite a bit more complex than module packages. While technically there is no specific restriction on what can be used to build or develop a website package, it is recommended that you use [Snowpack](https://www.snowpack.dev/). This doc. will assume that you're using Snowpack and that you started with the `templates/website-template` configuration.
```
emmiegit: What possible reasons are there for using something other than Snowpack? Would it make sense to explain why a divergence would occur? In general it seems any standard new package should use Snowpack, so we should recommend it in stronger terms here.

client/docs/websites.md
```diff
@@ -0,0 +1,34 @@
...
+  - index.html
+  - **/*
+- package.json
+- snowpack.config.cjs
+- tsconfig.json
+```
+
+The `public` folder is where source files that do not need compiling or transformations by Snowpack are kept. The `src` folder is where the rest of the source files go, such as TypeScript, CSS, etc. The file structure of these two folders are compiled/built into a `dist/` folder by Snowpack.
+
+Other than the inheritance of development dependencies and tooling configurations, websites act like fairly normal NPM packages. If you need to add a dependency to one, you can either navigate to the website package and do a normal `npm i <dep>` command, or manually add it to the `package.json`. After you do that, you need to make sure to run the `npm run bootstrap` command at the root. This ensures that all the fragile symlinks get restored by Lerna.
```
emmiegit, line 20: Ditto re `npm i --save`

client/docs/websites.md
```diff
@@ -0,0 +1,34 @@
...
+
+If you are wondering how you actually make a website using Snowpack, I would look into the Snowpack documentation. It's a fairly straightforward process.
```
emmiegit: One thing I'm a little confused about with this is what "website" means here. This is for building the Wikijump frontend, why would there be websites within our website?

client/modules/wj-util/src/index.ts
```diff
@@ -0,0 +1,159 @@
...
+  // this wacky approach is faster as it avoids an iterator
+  const keys = Object.keys(obj) as (keyof T)[]
+  for (let i = 0; i < keys.length; i++) {
+    if (obj[keys[i]] === undefined) delete obj[keys[i]]
+  }
+  return obj as { [K in keyof T]: Exclude<T[K], undefined> }
+}
+
+/** Takes a string and escapes any `RegExp` sensitive characters. */
+export function escapeRegExp(str: string) {
```
emmiegit, line 86:
```suggestion
export function escapeRegExp(str: string): string {
```

client/modules/wj-util/src/index.ts
```diff
@@ -0,0 +1,159 @@
...
+      if (str.startsWith(sigil)) {
+        return unSigil(str.slice(sigil.length), sigils) as T
+      }
+    }
+  }
+  return str as T
+}
+
+/** Creates a simple pseudo-random ID, with an optional prefix attached. */
+export function createID(prefix = "") {
```
emmiegit, line 117:
```suggestion
export function createID(prefix = ""): string {
```

client/templates/website-template/src/index.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
```
emmiegit: Is this hardcoded English?

emmiegit (approved):
It builds fine on my machine, the structure, documentation, and code seem fine. I've left some comments on things, but overall I am fine with this being merged after my concerns are addressed.

client/.eslintrc.js
Monkatraz: Ah yeah, in this case, I can replace that with `{ ...rules.map(rule => ({ [rule]: type })) }`. Other uses of `Object.assign()` have to do with merging objects, rather than creating new ones. Sometimes linters are too hasty and assume that `Object.assign()` is now pointless, which totally gives the wrong idea.

Monkatraz: Wait, nevermind, I can't read my own code. No, the way this works is that is a giant argument list. The first `{}` object is the object that receives a giant list of arguments to be merged into it. In this case, I can't think of another way of doing this that isn't super imperative.

client/docs/modules.md
Monkatraz: It goes into `tests/dist` because `dist/` is effectively the emitted content of the NPM package, which shouldn't have built test files in it.

client/docs/modules.md
Monkatraz: `npm i <dep>` automatically adds the package as a dependency. If it didn't, that would be kinda weird.

client/docs/websites.md
Monkatraz: `web/` holds any number of websites. Wikijump already has at least two - the `.org` blog, and the actual wiki. One of these website packages would be the frontend proper, maybe just named `frontend`.

client/templates/website-template/src/index.html
Monkatraz: oh god wikidot oh no I didn't mean it I just copied this from somewhere and forgot to change it

client/modules/wj-util/src/index.ts
Monkatraz: If you want explicit function returns, that can be an eslint rule. Personally, I don't like that rule, because some return types can be really complex and it forces you to get super verbose or split everything into types.

Monkatraz: tbf it sounds like I'm just complaining about typed languages there, but it stems from how loose JS is. I don't actually know how many people set up their linter to enforce return types - I certainly don't see it that often. Usually I see it recommended because it speeds up the typechecker a little bit.

client/docs/modules.md
Monkatraz: Well, we have that case already when you prefix with `node-`. I feel like that information is important because I don't want to hide how the repo works fundamentally.

client/docs/modules.md
Monkatraz: Lerna is in independent mode, which means packages get versioned arbitrarily depending on whether or not they have changed since they've last been published. I left versioning entirely alone in this because it's a bit of a can of worms. If you don't mind, I'd like to handle that in a separate PR when we start actually uploading packages.

client/docs/useful-commands.md
Monkatraz: gosh darn TYPOS

client/docs/useful-commands.md
Monkatraz: No, you run it afterwards. Bootstrap requires that Lerna itself be installed already, and Lerna itself needs access to the installed modules.

client/docs/websites.md
Monkatraz: This goes back to me "not wanting to hide how the repo works", but I can think of a few reasons. If you were using a static site generator, say for a development blog, you wouldn't want to demand Snowpack because it's not needed. All that is technically needed is a `dev` and `build` command. The simplicity of that requirement makes me feel uneasy about demanding the usage of Snowpack.

client/.prettierrc.toml
```diff
@@ -0,0 +1,10 @@
+printWidth = 90
+semi = false
+singleQuote = false
+quoteProps = "preserve"
+trailingComma = "none"
+bracketSpacing = true
+arrowParens = "avoid"
+# let git handle normalizing
+endOfLine = "auto"
+embeddedLanguageFormatting = "off"
```
rossjrw, line 10: Is there a dealbreaker that means turning this on would be a terrible idea?

client/docs/modules.md
```diff
@@ -0,0 +1,39 @@
+# Monorepo: Modules
+
+Modules are the packages that can be found in the `modules/` folder. They're intended for pure ESM output, and are compiled rapidly using Estrella, an esbuild-based utility. They have an important file structure (don't change it), which looks like the following:
```
rossjrw:
```suggestion
Modules are the packages that can be found in the `modules/` folder. They're intended for pure ESM output, and are compiled rapidly using [Estrella](https://github.com/rsms/estrella), an [esbuild](https://github.com/evanw/esbuild)-based utility. They have an important file structure (don't change it), which looks like the following:
```

rossjrw:
```suggestion
Modules are the packages that can be found in the `modules/` folder. They're intended to produce self-contained ES modules for consumption by other utilities, and are compiled rapidly using Estrella, an esbuild-based utility. They have an important file structure (don't change it), which looks like the following:
```

client/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+To actually setup the repo, run this command:
+
+```
+$ npm run update
+```
+
+This will install dependencies, clean package modules, bootstrap, and then build all packages.
+
+Run `npm run bootstrap` any time you need to ensure that packages have their dependencies sorted out.
```
rossjrw:
```suggestion
'Bootstrap' refers to the process where Lerna symlinks dependencies within this monorepo (as opposed to e.g. uploading each package within this monorepo to NPM and installing with `npm i`). The symlinks are fragile and will break any time a package is reinstalled. Run `npm run bootstrap` in the client monorepo root any time you need to ensure that packages have their dependencies sorted out.
```

client/docs/modules.md
rossjrw: I think I agree with ais here - excise this line, but keep the one below it specifically about 'node-'

client/docs/useful-commands.md
```diff
@@ -0,0 +1,19 @@
+```
+$ npm run update         Install dependencies, clean packages, bootstrap, then build.
+$ npm i                  Install the root's packages. Be sure to run 'npm run bootstrap'.
```
rossjrw:
```suggestion
$ npm i                  Install the root's packages. Be sure to run 'npm run bootstrap' afterwards.
```

client/docs/useful-commands.md
```diff
@@ -0,0 +1,19 @@
...
+$ npm run bootstrap       Link packages together, hoist shared dependencies.
```
rossjrw:
```suggestion
$ npm run bootstrap      Link packages together, hoist shared dependencies.
```

client/docs/websites.md
rossjrw:
```suggestion
This monorepo has special support for website packages, in the `web/` folder. These are quite a bit more complex than module packages. While technically there is no specific restriction on what can be used to build or develop a website package, it is recommended that you use [Snowpack](https://www.snowpack.dev/). This documentation will assume that you're using Snowpack and that you started with the `templates/website-template` configuration.
```

client/docs/websites.md
rossjrw: I don't think this line really adds anything - we would have to hope that contributors aren't making websites willy-nilly
```suggestion

```

client/docs/websites.md
```diff
@@ -0,0 +1,34 @@
...
+
+There is a template for making websites, and it can be found in the `templates/website-template` folder.
+
+1. Create a folder in `web/` that is the name of your website.
```
rossjrw:
```suggestion
1. Create a folder in `web/` named the same as the website's package name.
```
'the name of the website' is a little ambiguous - for example this would imply `wikijump.org`, but `blog` would be just as valid (in fact maybe moreso if dots aren't allowed in package names? unsure)

client/docs/websites.md
```diff
@@ -0,0 +1,34 @@
...
+
+2. Copy the contents of `templates/website-template` into the new folder.
+
+3. Edit `package.json`, and change the `name` field. It should probably match the name of the folder.
```
rossjrw:
```suggestion
3. Edit `package.json`, and change the `name` field to the name of this package.
```

client/docs/websites.md
rossjrw: The documentation here generally seems to assume that the reader is working as an individual and can do whatever they want, but that's not really the case here, as this is a team project with checks and measures and shit. A little authoritativeness with the implied "and discuss with the rest of the team" wouldn't hurt, though this is fine for now. Generally though I think it's a not-so-great idea to have documentation that ambiguously either describes the repository as it currently is, or speculatively describes what it could be

client/modules/node-snowpack-workers/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "type": "commonjs",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/scpwiki/wikijump.git"
+  },
+  "bugs": {
+    "url": "https://scuttle.atlassian.net/servicedesk/customer/portal/2"
+  },
```
rossjrw, line 22: I'm personally of the opinion that stuff that's relevant only to published npm packages might as well be omitted from local package.jsons - in this case `license`, `keywords`, `repository` and `bugs` (`description` is pretty useful and also npm will drop a warning if it's omitted). This presumably applies to the other package.jsons too but this is the first one I encountered in the review. If you do choose to keep this info I'd at least update the repository link to point to each package's sub-sub-repository (maybe on the develop branch)

client/docs/modules.md
emmiegit: It does? That is not my understanding of how `npm install` works.

Okay then.

client/docs/modules.md
emmiegit: If they're bumped automatically, what does that mean exactly? Does it just change the minor version every time or something?

This doesn't seem like an argument against calver per se, but I also don't really get what lerna does with versions so.

client/docs/useful-commands.md
emmiegit: Okay, in that case:
```suggestion
$ npm i                  Install the root's packages. Be sure to run 'npm run bootstrap' afterwards.
```

client/docs/websites.md
emmiegit: Oh I see, okay.

client/docs/modules.md
Monkatraz: To be honest, I haven't played with Lerna much. iirc, you give it `major | minor | patch` and it bumps? I have to look into it, and calver seems appealing to me.

client/modules/wj-util/src/index.ts
```diff
@@ -0,0 +1,159 @@
...
+    else if (cmp > 0) max = index - 1
+  }
+
+  if (!precise) return { element: null, index }
+
+  return null
+}
+
+/** Checks if an array or object is empty. Will return true for non-objects. */
+export function isEmpty(obj: any) {
```
rossjrw, line 53: We don't have `noExplicitAny`?

client/docs/modules.md
Monkatraz: Just to explain why it would be weird - NPM tries to keep your `node_modules` folder clean and deduped, and minimizes anything that _isn't_ your dependencies. So, if you install a package, didn't save it, then installed another one, the first one would get deleted. So, it's assumed installing a package means adding it as a dependency, which it is 99% of the time.

client/modules/wj-util/src/index.ts
```diff
@@ -0,0 +1,159 @@
...
+
+/** Takes a string and escapes any `RegExp` sensitive characters. */
+export function escapeRegExp(str: string) {
+  return str.replace(/[.*+?^${}()|\[\]\\]/g, "\\$&")
+}
+
+/** Checks if a string has any of the provided sigils.
+ *  (e.g. `hasSigil('!string', '!') -> true`) */
+export function hasSigil<T extends string = string>(
+  str: unknown,
```
rossjrw, line 93:
```suggestion
  str: string,
```

client/modules/wj-util/src/index.ts
```diff
@@ -0,0 +1,159 @@
...
+  sigils: string | string[]
+): str is T {
+  if (typeof str !== "string") return false
+  if (typeof sigils === "string") return str.startsWith(sigils)
+  for (const sigil of sigils) if (str.startsWith(sigil)) return true
```
rossjrw, line 98: TIL you can have a single line for..of

I'm still gonna use `return sigils.some(sigil => str.startsWith(sigil))` tho

client/modules/wj-util/src/index.ts
```diff
@@ -0,0 +1,159 @@
...
+  return false
+}
+
+/** Removes sigils from a string recursively. */
+export function unSigil<T extends string>(str: T, sigils: string | string[]): T {
+  if (typeof str !== "string") throw new TypeError("str must be a string")
```
rossjrw:
```suggestion

```

Seems redundant?

client/modules/wj-util/src/index.ts
```diff
@@ -0,0 +1,159 @@
+import { isAnyObject } from "is-what"
```
rossjrw, line 1: There are quite a few utilities in this file - it's fine for now but if any more are needed I think it will be a good idea to break it out into a few

Also, I assume all of these are definitely going to be used by stuff you plan to add? I'd rather not have any unused stuff floating around in there

client/package.json
```diff
@@ -0,0 +1,76 @@
...
+    "build:web": "ultra -r --filter \"web/*\" build",
+    "build:cmd": "npm run build:modules && npm run build:web",
+    "build": "ultra --build build:cmd",
+    "dev:modules": "tsc -b -w && node scripts/build-modules.js -w -g",
+    "dev": "ultra dev:modules",
+    "test": "node scripts/build-modules.js -tests && uvu modules tests[/\\\\\\\\]dist",
+    "eslint": "eslint \"./**/*.js\" \"./**/*.ts\"",
+    "eslint-fix": "eslint \"./**/*.js\" \"./**/*.ts\" --fix",
+    "prettier": "prettier --check \"modules/**\" \"web/**\"",
+    "prettier-fix": "prettier --write \"modules/**\" \"web/**\"",
```
rossjrw, line 21: Would adding a pre-commit hook for `prettier-fix` be too much? (out of scope for this PR)

client/package.json
```diff
@@ -0,0 +1,76 @@
...
+    "filler-comment": "The scripts below are to help with the individual web/ packages.",
```
rossjrw, line 22: Lmao

It's a shame you can't define scripts with a description field

Is there a way to generate these commands for each web package? Maybe something like `dev:web:[package name]` or something? If not, could adding something here be added to the websites documentation?

client/scripts/build-modules.js
```diff
@@ -0,0 +1,186 @@
+const { build, cliopts, basename, stdoutStyle: styl, fmtDuration } = require("estrella")
```
rossjrw, line 1: I noticed the snowpack config was `.cjs` - it's not something I normally do, but how would you feel about making all scripts that use `require` be `.cjs` and having all `.js` and `.ts` implicitly use `import`?

rossjrw, line 1: I'd like to see a wad of documentation here explaining what this script does, even if it's redundant

client/scripts/build-modules.js
```diff
@@ -0,0 +1,186 @@
...
+  sourcesContent: true,
+
+  // estrella settings
+  tslint: false, // disables estrella's built-in tsc typechecking
+  quiet: true // silences estrella's logging, as we use our own logging
+}
+
+// prettier-ignore
+const DEMANGLE_TARGET_REGEX =
+  /^(chrome|and_(?:ff|chr)|edge|firefox|ios_saf|safari)\s+([\d.]+)/
```
rossjrw, line 42: It's a bit annoying that esbuild's creator doesn't seem interested in supporting browserslist - there's definitely no package that does this? (if not, might be an idea to make one lmao)

Are you sure the code below does everything right, doesn't miss any browsers, etc? Might be worth making a unit test for it

client/scripts/build-modules.js
```diff
@@ -0,0 +1,186 @@
...
+    index = `${dir}/src/index.ts`
+    package = `${dir}/package.json`
+  }
+
+  if (TESTS) {
+    try {
+      tests = filesOf(`${dir}/tests`).map(name => `./tests/${name}`)
+      if (tests.length === 0) {
+        console.log(fmt(`[${name}]`), "No tests, skipping.")
+        return
```
rossjrw, line 96: You're gonna hate me for this
```suggestion
        throw new Error()
```

client/modules/wj-util/tests/wj-util.ts
```diff
@@ -0,0 +1,75 @@
+import * as uvu from "uvu"
+import * as assert from "uvu/assert"
+
+import * as lib from "../src/index"
+
+const Utils = uvu.suite("wj-utils")
```
rossjrw, line 6:
```suggestion
import { suite } from "uvu"
import assert from "uvu/assert"

import * as lib from "../src/index"

const Utils = suite("wj-utils")
```

client/templates/module-template/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/scpwiki/wikijump.git"
+  },
+  "bugs": {
+    "url": "https://scuttle.atlassian.net/servicedesk/customer/portal/2"
+  },
+  "homepage": "https://github.com/scpwiki/wikijump#readme",
```
rossjrw, line 23: Barring the caveat from way up above, in the template this should be some sort of generic URL that would point to the package subsubrepo, right? That is what monorepos on NPM typically do anyway

client/templates/module-template/tests/test-template.ts
```diff
@@ -0,0 +1,10 @@
+import * as uvu from "uvu"
+import * as assert from "uvu/assert"
+
+const SuiteName = uvu.suite("SuiteName")
```
rossjrw, line 4:
```suggestion
import { suite } from "uvu"
import assert from "uvu/assert"

const SuiteName = uvu.suite("SuiteName")
```
Same suggestion here as above

client/templates/website-template/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "website-template",
+  "license": "agpl-3.0-or-later",
+  "private": true,
+  "version": "1.0.0",
```
rossjrw:
```suggestion
  "version": "0.0.0",
```

At least for the template

client/templates/website-template/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "url": "https://scuttle.atlassian.net/servicedesk/customer/portal/2"
+  },
+  "homepage": "https://github.com/scpwiki/wikijump#readme",
+  "eslintConfig": {
+    "extends": "../../.eslintrc.js"
+  },
+  "dependencies": {
+    "cm-tarnation": "^0.1.0",
+    "wj-util": "^1.0.0"
+  }
```
rossjrw:
```suggestion

```

client/templates/website-template/snowpack.config.cjs
```diff
@@ -0,0 +1,37 @@
...
+    src: "/",
+    public: {
+      url: "/",
+      static: true,
+      resolve: false
+    }
+  },
+  packageOptions: {
+    polyfillNode: true,
+    packageLookupFields: ["svelte"]
```
rossjrw, line 13:
```suggestion

```
Seems odd to assume that all future website packages will be using Svelte

client/templates/website-template/snowpack.config.cjs
```diff
@@ -0,0 +1,37 @@
...
+  optimize: {
+    // preload: true,
+    minify: true,
+    treeshake: true,
+    splitting: true,
+    sourcemap: true,
+    manifest: true,
+    target: "es2020"
+  },
+  plugins: ["@snowpack/plugin-svelte", "snowpack-workers"]
```
rossjrw, line 36:
```suggestion

```

client/templates/website-template/src/static/lib/index.ts
```diff
@@ -0,0 +1 @@
+export {}
```
rossjrw, line 1: I don't suppose you have a citation for the `static/lib` directory structure? Seems inconsistent to use `lib` sometimes, `src` other times, and `static` for precompiled files when semantically that seems to line up with what `public` is?

client/templates/website-template/src/static/css/index.css
```diff
@@ -0,0 +1,11 @@
+/* Normalize html, body elements */
```
rossjrw, line 1: I would replace this with a stub file - there's no reason to assume anything; we'll probably have a decent CSS backing at some point

client/tsconfig.json
```diff
@@ -0,0 +1,53 @@
...
+    "composite": true,
+    "declaration": true,
+    "declarationMap": true,
+    "emitDeclarationOnly": true,
+    "preserveWatchOutput": true
+  },
+
+  // Modules must be added here in order for their types to be compiled.
+  // Unfortunately, there isn't a decent way to do this automatically.
+  // The performance benefits are worth it, regardless.
```
rossjrw, line 34: This is **peak** documentation. 10/10 snippet

client/scripts/build-modules.js
```diff
@@ -0,0 +1,186 @@
...
+  return readdirSync(source, { withFileTypes: true })
+    .filter(dirent => dirent.isDirectory())
+    .map(dirent => dirent.name)
+}
+
+/** Gets the files (shallow) of a folder. */
+function filesOf(source) {
+  return readdirSync(source, { withFileTypes: true })
+    .filter(dirent => dirent.isFile())
+    .map(dirent => dirent.name)
```
rossjrw, line 185: What does 'dirent' mean?

client/docs/websites.md
```diff
@@ -0,0 +1,34 @@
...
+
+4. You're done!
```
rossjrw:
```suggestion
1. Create a folder in `web/` that is the name of your website.
1. Copy the contents of `templates/website-template` into the new folder.
1. Edit `package.json`, and change the `name` field. It should probably match the name of the folder.
1. You're done!
```
We can let Markdown generate the numbers

rossjrw (changes requested):
I was wrong and you were right - this is nowhere near as overcomplicated as I thought it was. There are a few irritating workarounds, sure, but yeah I do agree with you now that it's worth it. Kudos to the simplifications brought by estrella and ultra runner.

client/.prettierrc.toml
Monkatraz: No, it's just annoying to handle. Git automatically converts LF into CRLF for Windows, and CRLF to LF when committing. I don't want to do the tricks required to override this behavior, and it's a very normal feature in Git, so it's not brittle.

rossjrw: Sorry, I was referring specifically to `embeddedLanguageFormatting `

client/modules/node-snowpack-workers/package.json
Monkatraz: `license` is required if the package is to be published. `keywords` and `repository` should probably be kept too, and for the same reason. `bugs` is more debatable, but if a user goes to the repository to report a bug, and finds issues disabled, that's kind of annoying.

client/modules/wj-util/src/index.ts
Monkatraz: No, I don't tend to use that feature. My justification is simple: `any` can be found in the native TypeScript type definitions, as sometimes `any` is required to describe certain objects. Matching against _any_ generic JS object is really hard, and it's better to make the typechecker ignore it and just ensure you're doing it right. It's kinda like `unsafe` in Rust.

client/modules/wj-util/src/index.ts
Monkatraz: That code checks if `str` is a string or not. It's so that you can pass in objects willy-nilly and not worry about their type. Tarnation uses this a lot when demangling the grammar definition, it's a really useful feature.

client/modules/wj-util/src/index.ts
Monkatraz: Technically, both solutions here use an iterator, but one uses more function calls than the other. I think. I think breaking out of a `for of` is faster.

client/modules/wj-util/src/index.ts
Monkatraz: Yeah, it is. Artifact from when this cared less about the type.

client/modules/wj-util/src/index.ts
Monkatraz: Oh yeah, Tarnation uses `is-what` a lot when demangling the grammar definition. A bit of an artifact, we could get rid of the dependency. (Let's do that later, please)

client/package.json
Monkatraz: No, not at all. Prettier is explicitly supposed to not break code and I added these commands more-so for hooks and actions.

client/package.json
Monkatraz: I thought about it, it was hard. There are some features in run-parallel that I could maybe use, or I can make a Node script for it. I said screw it and decided not to for now.

client/scripts/build-modules.js
Monkatraz: That would be fine, I suppose. It makes sense. However, in the `root` package, I don't want the `package.json` to use `type: "module"` because that sometimes does some _Really Wacky Behavior_ with build tools.

client/scripts/build-modules.js
Monkatraz: Unit testing the root package isn't something I've thought about lmao

Yeah, I'm fairly sure this doesn't miss any browsers. I've worked with way, WAY too much regex. This is a very narrow matcher, and the code below seems fairly simple to me.

client/scripts/build-modules.js
Monkatraz: How would I possibly test `node-snowpack-workers`? That's something that requires something more akin to a browser test, which we have not figured out how to do at all.

client/modules/wj-util/tests/wj-util.ts
Monkatraz: Normally default exports are discouraged, and I usually code like they don't exist, except in certain circumstances. Also, `uvu` has more than just the `suite` function, it also has the `test` function.

client/templates/module-template/package.json
Monkatraz: Probably there should be? Personally, I'm find with letting people figure out where the package is - it's not too hard.

client/templates/website-template/package.json
Monkatraz: I'm not sure what a `0.0.0` dependency does with NPM. I don't like playing with fire.

client/templates/website-template/package.json
Monkatraz: oops nice catch

client/templates/website-template/snowpack.config.cjs
Monkatraz: It's not - it's a generic configuration option. All this does is add another field for Rollup to check for package dependencies. The editor already uses Svelte and I already had eslint configured for it.

client/templates/website-template/snowpack.config.cjs
Monkatraz: I'm not sure why this helps - why by default reduce Snowpack's capability? All these do is add file resolvers

client/templates/website-template/src/static/lib/index.ts
Monkatraz: `static` is a common convention for website assets, not compiled vs noncompiled. In a web-server, the root folder may contain a lot of other dynamic _stuff_ that changes, while `static` is just pure assets from the build process.

client/templates/website-template/src/static/css/index.css
Monkatraz: It's a filler file, with some example CSS. It's just so you know where it goes, and so I used a common reset to demonstrate. I suppose I could get rid of it?

client/scripts/build-modules.js
Monkatraz: Directory entity, which might be various things, such as a file or another directory.

client/docs/websites.md
Monkatraz: I uh... don't like this. I'm all for being terse but, this just reads really weird.

client/.prettierrc.toml
Monkatraz: Oh, well, it would be inconsistent. If you had some embedded `HTML` in some JS, and had this on, it would format the `HTML`, but we don't format `HTML`.

client/modules/node-snowpack-workers/package.json
rossjrw: Are any of these going to be published?

client/modules/wj-util/src/index.ts
rossjrw: Ah, sorry, I just left this comment at the top of the file but didn't mean to refer to the import specifically (good point, though :P)

client/modules/node-snowpack-workers/package.json
Monkatraz: I thought so? Some of the stuff we make might be pretty neat and worth publishing. Also, if there is any sister projects, having available our packages might be nice.

client/modules/wj-util/src/index.ts
Monkatraz: Oh yeah, all of these utilities are used. It's kind of a dumping ground for one-off functions.

client/package.json
rossjrw: Created [WJ-454](https://scuttle.atlassian.net/browse/WJ-454)

client/package.json
rossjrw: That's fair, in that case, officially requesting that this process be documented

client/scripts/build-modules.js
rossjrw: True. I take this back, no reason to play with fire unless there's a great need to. I might consider swapping the .cjs for .js at some point, lmao

client/scripts/build-modules.js
rossjrw: The line below this suggestion is a `catch` that emits the exact same console.log

client/package.json
Monkatraz: It is partially documented (in a sort of cop-out way) in the `websites.md` doc, but I'd rather make a solution for it than agonizingly describe how you do this correctly. If you don't mind, I'd rather fix it in a separate PR (I'm thinking of making a Node script for it).

client/scripts/build-modules.js
Monkatraz: oh right. it's 4am. I don't see the need to throw an error, and then have to catch it, when you can just return out of the function. plus, the two code paths mean two separate things even if they emit the same log

client/templates/website-template/package.json
rossjrw: It just means unversioned - the code is always considered to have 'changed' e.g. if you push it to npm it will override the existing code and won't complain, even if you didn't increment it in the meantime. I'd rather use a convention like that than have someone accidentally commit to a 1.0.0 release when making a new package.

client/templates/website-template/snowpack.config.cjs
rossjrw: Right, but this is the template package, it should be as minimal as possible

client/templates/website-template/package.json
Monkatraz: TIL

client/templates/website-template/snowpack.config.cjs
rossjrw: In that case, please compile a list of all other possible file resolvers and add them to the template 😜

client/templates/website-template/snowpack.config.cjs
Monkatraz: I uh... I just don't see the need, lol. I was even willing to expand this template if we added other file types, so that there was a common source of configuration "knowledge"

client/templates/website-template/src/static/lib/index.ts
rossjrw: Do you have a citation? I'd like to know what dirs would be expected in `static` in this structure, what dirs are siblings to `static` e.g. `electronic`, `high-voltage` etc

Monkatraz: uh I don't have a citation but I remember reading about it from website boilerplate templates. I made up the `lib` thing myself when trying to figure out how to sort mainframe, and I found it to be a nice place for components and code, but nothing else.

client/docs/websites.md
rossjrw: Would you at least consider squishing the lines together so it gets `ol` + auto-numbering even if the actual numbers in the source are wrong? (I think that's how it works anyway)

```suggestion
1. Create a folder in `web/` that is the name of your website.
2. Copy the contents of `templates/website-template` into the new folder.
3. Edit `package.json`, and change the `name` field. It should probably match the name of the folder.
4. You're done!
```

client/.prettierrc.toml
rossjrw: Good point

client/docs/websites.md
Monkatraz: does GFM not handle line breaks like that?
oops
god darn it I can't accept the commit suggESTSTION

client/templates/website-template/src/static/lib/index.ts
rossjrw: image

is that not... what goes here? what is there other than components and code?

Monkatraz: Styling, yaml/json data, etc. went outside the `lib` folder

client/docs/websites.md
```diff
@@ -0,0 +1,33 @@
...
+
+There is a template for making websites, and it can be found in the `templates/website-template` folder.
+
+1. Create a folder in `web/` named the same as the website's package name.
+
+2. Copy the contents of `templates/website-template` into the new folder.
+
+3. Edit `package.json`, and change the `name` field to the name of this package.
+
+4. You're done!
```
rossjrw:
```suggestion
1. Create a folder in `web/` named the same as the website's package name.
2. Copy the contents of `templates/website-template` into the new folder.
3. Edit `package.json`, and change the `name` field to the name of this package.
4. You're done!
```

rossjrw (approved):

client/docs/useful-commands.md
Monkatraz: this is already in the repo, I think ais suggested it, what is happening

client/scripts/build-modules.js
Monkatraz: there is actually a jsdoc standard for this, which we should just probably agree on using at some point

Monkatraz:
image

if another pr beats this in comments, I'll be impressed

also, 90% of the review comments were my documentation. how about next time I give you no documentation, huh? how would you like that? fools

Monkatraz:
Okay, I'm going to merge this. I know @pxdnbluesoul hasn't done a review, and may want to do so - but a few things are blocked by this PR getting merged. I am 110% willing to make any changes that may be requested, and will try to do so ASAP in another PR.

PR #184, opened 2021-04-13
Change page length checks to characters, not bytes

emmiegit:
This makes it consistent with both forum posts and the error message describing them as "characters".

See also this [@wikidotbugs tweet](https://twitter.com/wikidotbugs/status/1382079705231331333).


rossjrw (approved):

emmiegit:
thanks @rossjrw

PR #188, opened 2021-04-14
[FTML-74] Implement HTML renderer

emmiegit:
The HTML renderer was largely implemented in `ftml-legacy` but has not yet been carried over. This PR adds HTML rendering for all currently-extant `Element`s, as well as other changes needed for the render process to work. For instance, `PageInfo` describes the page being rendered, which needs to be retrieved and passed to the renderer.

A needed concept is the `Handle`, which contacts the backend for various information related to the rendering context. This is not yet implemented, and still needs to evolve as we add more elements and integrate ftml with the PHP.

It also adds HTML tests to match the existing JSON tests -- if the generated HTML doesn't match what was expected, it'll error. This has caught some bugs, but I will address them in a future PR.


danieltharp (approved):

ftml/src/render/html/element/iframe.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+pub fn render_html(log: &slog::Logger, ctx: &mut HtmlContext, contents: &str) {
+    debug!(
+        log,
+        "Rendering html block (submitting to remote for iframe)";
+        "contents" => contents,
+    );
+
+    let iframe_url = ctx.handle().post_html(log, ctx.info(), contents);
```
Monkatraz, line 47: What does this do? Is there some sanitation done on iframe urls?

ftml/src/render/html/element/text.rs
```diff
@@ -0,0 +1,73 @@
...
+    contents: &str,
+) {
+    debug!(
+        log,
+        "Rendering code block";
+        "language" => language.unwrap_or("<none>"),
+        "contents" => contents,
+    );
+
+    // TODO: syntax highlighting based on 'language'
```
Monkatraz, line 70: Also, isn't there some madness to handle where WD will provide urls to the raw content of each code block? Not sure if that's in FTMLs territory, but it'll probably require parsing the AST for it.

ftml/README.md
```diff
@@ -148,6 +159,14 @@ let result = ftml::parse(&log, &tokens);
...
 // occurred during the parsing process.
 let (tree, warnings) = result.into();
+
+// Finally, we render with our renderer. Generally this is `HtmlRender`,
+// but you could have a custom implementation here too.
+//
+// You must provide a `PageInfo` struct, which describes the page being rendered.
+// You must also provide a handle to provide various remote sources, such as
+// module content, but this is not stabilized yet.
+let html_output = HtmlRender.render(&log, &page_info, &tree);
```
Monkatraz, line 169: It may be helpful to provide a way of rendering without a `page_info` object. Rendering small chunks of inline text or an editor previewing an otherwise empty draft may require making up the `page_info`, and that feels like something FTML should be able to handle.

ftml/src/render/html/builder.rs
```diff
@@ -30,6 +31,9 @@ macro_rules! tag_method {
     };
 }
 
+/// These are HTML tags which do not need a closing pair.
+const SOLO_HTML_TAGS: [&str; 3] = ["br", "br", "input"];
```
Monkatraz: Should the full list of self-closing tags be put in here? As these are only the ones that FTML is currently using - but common tags like the `` tag self-close, too.

Monkatraz (approved):
I'm not good enough at Rust to properly review the entirety of this, but this looks fine to me. Most of my comments are about hypotheticals or "what's this?".

ftml/src/render/html/builder.rs
Monkatraz: Oh, also, `br` is in there twice. I assume the other one is supposed to be `hr`?

ftml/README.md
Monkatraz: Actually, an FTML rendering context for inline text that avoids parsing and processing things like modules would probably help too. This pops up sometimes when writing Markdown renderers, because it's helpful when you have small chunks of formatted text to render. The easiest example I can come up with is user profiles - imagine having a `tagline` property in your profile, which is a chunk of text displayed with your user avatar. If it had formatting, it seems overkill to render it with the full FTML renderer.

ftml/src/render/html/element/iframe.rs
emmiegit: All `[[html]]` blocks aren't injected directly inside the document, but rather are hosted in an iframe on `wdfiles.com`. This (temporary) method tells the backend to create a HTML document with the given contents and return its URL for the iframe.

I'll add a brief comment to explain this.

ftml/src/render/html/element/text.rs
emmiegit: Oh right I forgot about that. Lemme add a stub for that.

ftml/README.md
emmiegit: I have two points:

First, it's not that much data, just some strings that you can pretty easily mock if it's an unattached draft or something.

Second, the data is needed to properly render the page, even if it's only a snippet. Consider `[[iftags]]`, what is that supposed to do if the relevant tags aren't known?

ftml/src/render/html/builder.rs
emmiegit: Yeah I can do that.

ftml/README.md
Monkatraz: Good points, although on the first point I will say that every JS<->Rust interaction has a performance and latency hit, even if seemingly insignificant. The editor may be firing a preview on nearly every keystroke, keep in mind.

On the second point, yeah that's tough. In an inline context, I would just assume that `[[iftags]]` isn't valid, but in a preview context _shrug_.

emmiegit: You only need to build the Rust/WebASM version of the `PageInfo` struct once, right? I guess the title could change but that's not exactly on the hot path, and everything else shouldn't change at all during an editor session.

ftml/src/render/html/builder.rs
emmiegit: Also yes the second one was meant to be `hr`. I had replaced the `HashSet` with an array since it was faster and type-copied the name wrong.

ftml/README.md
Monkatraz: I don't think that's the case - when I build the `PageInfo` struct, I'm fairly sure once I feed it into the `html_render` function it'll get, uh, `free`ed? I'm not sure what happens to it - it probably has something to do with Rust's borrowing. I can't reuse syntax trees, as an example.

emmiegit: Oh, hmm. I thought I set everything up so they can be reused, but JS must be consuming the ownership of something. I will take a look.

emmiegit: Okay, I've added the method `copy()` to all the structures. This produces a second instance of the same object, which is a very cheap operation. This way you can reuse the objects you receive / create.

emmiegit:
thanks @pxdnbluesoul @Monkatraz

PR #190, opened 2021-04-15
Add Nginx and Datadog containers to dev

danieltharp:
This provides dev with the same enhancements we've been testing locally around moving nginx to its own container and reducing the php-fpm container workload to just PHP stuff. It adds a GitHub Action to build and push the nginx container to ECR. It also adds the Datadog agent to the ECS task definition so it should be able to collect info from the containers. The Datadog stuff is in early days but it's an important first step.

The pull will be needed to fix dev as it's currently looking for the latest builds of each and can't find them. I think. Kinda learning but I think in the future we'll need to trigger new builds of all the containers when we deploy new terraform changes around containers. This will help prove the likelihood of that.

### DIFF
--- a/install/aws/dev/terraform/deploy/task-definitions.tf
+++ b/install/aws/dev/terraform/deploy/task-definitions.tf
@@ -194,24 +194,29 @@ module "reverse-proxy" {
 output "cache_json" {
   description = "Container definition in JSON format"
   value       = module.cache.json_map_encoded_list
+  sensitive   = true
 }
 
 output "database_json" {
   description = "Container definition in JSON format"
   value       = module.database.json_map_encoded_list
+  sensitive   = true
 }
 
 output "php-fpm_json" {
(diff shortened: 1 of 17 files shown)


install/aws/dev/terraform/deploy/task-definitions.tf
```diff
@@ -155,16 +224,19 @@ module "reverse-proxy" {
     {
       sourceVolume  = "docker-socket"
       containerPath = "/var/run/docker.sock"
+      readOnly      = true
+
```
emmiegit: Extra line?

danieltharp:
Noting that after running both the php and nginx workflows again, the cluster was able to launch. Terraform assumes you want to pull an image tagged with `develop` and we age that one out pretty rapidly in favor of images tagged with the commit hash.

emmiegit (approved):

PR #191, opened 2021-04-15
Use new dockerfile for dev php-fpm

danieltharp:
I goofed. This should've been part of #190. Dev is currently on the old monolithic config and I think it's causing the current errors we're seeing.

### DIFF
--- a/install/aws/dev/docker/php-fpm/Dockerfile
+++ b/install/aws/dev/docker/php-fpm/Dockerfile
@@ -46,7 +46,7 @@ RUN apk add --update --no-cache \
 # Install the Wikijump repository
 WORKDIR /var/www
 
-    git clone \
+RUN git clone \
         --depth 10 \
         --branch "${WIKIJUMP_REPO_BRANCH}" \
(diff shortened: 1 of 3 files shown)


install/aws/dev/docker/php-fpm/Dockerfile
```diff
@@ -34,74 +32,65 @@ RUN apk add --update --no-cache \
...
+    /src/setup-memcached.sh
 
 # Install the Wikijump repository
 WORKDIR /var/www
-RUN git clone \
-    --depth 10 \
-    --branch "${WIKIJUMP_REPO_BRANCH}" \
-    "${WIKIJUMP_REPO}" "${WIKIJUMP_REPO_DIR}"
+
+    git clone \
```
emmiegit: No `RUN`? This isn't a continuation of a previous command, this seems like an error.

danieltharp: You're right, the `RUN` statement was on a local-only line, thanks.

emmiegit (approved):

PR #193, opened 2021-04-16
Remove wikijump_token7

danieltharp:
This PR removes the `wikijump_token7` cookie and POST field from the AJAX Module Flow Controller and cookie setters. As far as we can determine, it was sort of a primitive anti-CSRF token, but totally ineffective.

We will implement proper CSRF tokens where it's sensible to do so, down the road.


emmiegit (approved):
:wastebasket: :cookie:

PR #195, opened 2021-04-18
[WJ-456] Add Sheaf's (frontend editor) core

Monkatraz:
This PR adds Sheaf's core, which is a wrapper around CodeMirror 6. This is effectively a basic implementation of CM6 that is intended to be used with UI components.

This PR is also the first in quite a few, and is a little bit messy. My standards for code aren't too high here - mainly because a bunch of code needs to be moved over from Mainframe (or created from scratch) to get the editor in a "it's all there" state. Once that has been accomplished, I can begin refactors and clean ups of the code.

The general train of PRs is probably gonna look something like this:
* Add the editor core and test environment (this PR)
* Create an FTML WASM package (#203)
* Create an FTML language package (most of this work will be converted from Mainframe) (#207)
* Create a Sheaf UI component, likely in Svelte
* Refactor, clean up all of this


client/modules/sheaf-core/src/extensions/indent-hack.ts
```diff
@@ -0,0 +1,64 @@
...
+import { EditorState } from "@codemirror/state"
+import {
+  EditorView,
+  ViewPlugin,
+  ViewUpdate,
+  Decoration,
+  DecorationSet
+} from "@codemirror/view"
+import { RangeSetBuilder } from "@codemirror/rangeset"
+import type { Line } from "@codemirror/text"
```
emmiegit, line 15: I didn't know this was a thing.

client/modules/sheaf-core/src/index.ts
```diff
@@ -0,0 +1,204 @@
...
+    indentHack,
+    keymap.of([
+      ...closeBracketsKeymap,
+      ...defaultKeymap,
+      ...searchKeymap,
+      ...historyKeymap,
+      ...foldKeymap,
+      ...commentKeymap,
+      ...completionKeymap,
+      { key: "Mod-l", run: openLintPanel, preventDefault: true },
```
emmiegit, line 196: I thought we decided on something else? Given that Ctrl+L normally selects the address bar for easy URL changing.

client/modules/sheaf-core/src/print-tree.ts
```diff
@@ -0,0 +1,67 @@
...
+  tree: Tree,
+  input: Input | string,
+  options: { from?: number; to?: number; start?: number; includeParents?: boolean } = {}
+): string {
+  const cursor = tree.cursor()
+  if (typeof input === "string") input = stringInput(input)
+  const { from = 0, to = input.length, start = 0, includeParents = false } = options
+  let output = ""
+  const prefixes: string[] = []
+  for (;;) {
```
emmiegit: The convention in JS/TS isn't `while (true) {`?

client/modules/sheaf-core/src/theme.ts
```diff
@@ -0,0 +1,291 @@
...
+      borderRadius: "0.125rem"
+    },
+    ".cm-searchMatch.selected": {
+      backgroundColor: selection,
+      boxShadow: `0 0 0 0.075rem ${accent}`
+    },
+
+    ".cm-line": {
+      "& ::selection": { color: "inherit !important" },
+      "&::selection": { color: "inherit !important" }
```
emmiegit, line 98: These are different?

client/modules/wj-util/src/index.ts
```diff
@@ -156,3 +156,138 @@ export function perfy(meta?: string, threshold?: number): () => number {
...
+ *  The only difference is that this function uses `requestIdleCallback` instead.
+ *  If `requestIdleCallback` isn't available, it will use `createAnimQueued` instead.
+ * @see createAnimQueued */
+export function createIdleQueued<T extends AnyFunction>(fn: T, timeout = 100) {
+  if (!HAS_IDLE_CALLBACK) return createAnimQueued(fn)
+  let queued: boolean
+  return (...args: Parameters<T>): void => {
+    if (queued !== true) {
+      queued = true
+      ;(window as any).requestIdleCallback(
```
emmiegit, line 284: Is this something prettier did? It looks very weird.

emmiegit:
For `client/web/dev-sandbox/public/static/fonts/*` can you provide some document in that directory, even if it's short, that mentions where these font files come from and demonstrates that we have licensing permissions to use them?

client/web/dev-sandbox/public/static/misc/ftml-test.ftml
```diff
@@ -0,0 +1,258 @@
...
+,,quotation''
+<<quotation>>
+>>quotation<<
+dots...
+em -- dash
+
+This //text// gets **parsed**.
+@@This //text// does not get **parsed**.@@
+
+HTML entities: @<U umlaut: &#252;>@
```
emmiegit, line 61: fyi we are not supporting special HTML entity behavior in `@< .. >@` raws: them and `@@ .. @@` behave the same. Instead entities are done through `[[char ...]]`.

Monkatraz:
> For `client/web/dev-sandbox/public/static/fonts/*` can you provide some document in that directory, even if it's short, that mentions where these font files come from and demonstrates that we have licensing permissions to use them?

Oh yeah, totally can. FYI when I went looking for fonts I made sure to find MIT compatible fonts, and afaik they were basically all open font license. Let me get the sources.

client/web/dev-sandbox/public/static/misc/ftml-test.ftml
```diff
@@ -0,0 +1,258 @@
...
++ Level 1 Heading
+++ Level 2 Heading
++++ Level 3 Heading
+++++ Level 4 Heading
++++++ Level 5 Heading
+++++++ Level 6 Heading
+
++ This section appears in the TOC
++* And this one does not
+++* Neither does this one
```
emmiegit, line 84: wtf this is a wikidot feature?

I don't see anything about it in https://www.wikidot.com/doc-wiki-syntax:headings

emmiegit (approved):
I can't really comment on most of the Typescript code, but what I do understand in this PR makes sense.

client/modules/sheaf-core/src/extensions/indent-hack.ts
Monkatraz: Importing types? Yeah, when you have non `tsc` build tools, it's important for them to distinguish between a type import and a, uh, import-import, because it's obviously significantly more complex if you have to resolve imports for their AST.

client/modules/sheaf-core/src/index.ts
Monkatraz: There really isn't many other safe bindings, and we didn't decide on anything else. IIRC Ctrl + L is a duplicate of another binding, I think Ctrl + K. This is something we'll have to just test and hope is fine, I think.

client/modules/sheaf-core/src/print-tree.ts
Monkatraz: It's a bit of a conflict - and this isn't directly my code. CodeMirror internals use `for(;;)`, while yeah, I would personally use `while (true)`. I think there might be a lint rule for that?

client/modules/sheaf-core/src/theme.ts
Monkatraz: One is "if parent is selected" and another is "if something in the parent is selected"

client/modules/wj-util/src/index.ts
Monkatraz: Yeah, sometimes code gets mangled, especially when you're doing a weird type cast like this because there isn't a decent way in TS to assert the existence of something. There are proposals for it.

client/web/dev-sandbox/public/static/misc/ftml-test.ftml
Monkatraz: I wouldn't get too hung up about this test file (feel free to update it whenever, though). But that is good to know, thank you.

client/web/dev-sandbox/public/static/misc/ftml-test.ftml
Monkatraz: Well, this absolutely is copied from Wikidot. Let me see if I can find it...

https://www.wikidot.com/doc-wiki-syntax:table-of-contents

client/modules/wj-util/src/index.ts
emmiegit: I was referring to the `;`, I understand why `(x as any)` has to happen on occasion.

Monkatraz: Oh, sorry, yeah - that's Prettier. The idea is, when you're running without semicolons, is that you put the semicolon at the _start_ of the line when you need them, because it's super obvious.

A way of avoiding that would be TypeScript's original type cast assertion - `<any>window`. However, it's somewhat discouraged because it's incompatible with JSX/TSX.

client/modules/sheaf-core/src/print-tree.ts
Monkatraz: Okay, I've looked into this, and the result is confusing to me. Apparently, eslint's [no-constant-expression](https://eslint.org/docs/rules/no-constant-condition) rule, enabled by default, actually _disallows_ `while (true)`, while accepting `for (;;)`. If I had a guess why that is, it's because `for (;;)` is a bit more... in your face and obvious? There doesn't seem to be a way to configure it to work the other way around, which is strange.

client/modules/sheaf-core/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "sheaf-core",
```
rossjrw, line 2: triple-nested monorepo when :eyes:

client/modules/sheaf-core/src/extensions/indent-hack.ts
```diff
@@ -0,0 +1,64 @@
...
+
+const WHITESPACE_REGEX = /^\s+/
+
+function indentDeco(view: EditorView) {
+  // get every line of the visible ranges
+  const lines = new Set<Line>()
```
rossjrw, line 21: Why a set here and not an array? We don't want to fail to wrap lines if they're identical, right?

client/modules/sheaf-core/src/index.ts
```diff
@@ -0,0 +1,204 @@
...
+import { EditorState, Extension, Compartment } from "@codemirror/state"
+import {
+  EditorView,
+  ViewPlugin,
+  ViewUpdate,
+  drawSelection,
+  keymap
+} from "@codemirror/view"
+
+import { highlightActiveLine, highlightSpecialChars } from "@codemirror/view"
```
rossjrw:
```suggestion
  keymap,
  highlightActiveLine,
  highlightSpecialChars
} from "@codemirror/view"
```

client/modules/sheaf-core/src/index.ts
```diff
@@ -0,0 +1,204 @@
...
+    const updateHandler = ViewPlugin.define(() => ({
+      update: (update: ViewUpdate) => {
+        // update store on change
+        if (update.docChanged) updateState()
+        // get active lines
+        if (update.selectionSet || update.docChanged) {
+          const activeLines: Set<number> = new Set()
+          for (const r of update.state.selection.ranges) {
+            const lnStart = update.state.doc.lineAt(r.from).number
+            const lnEnd = update.state.doc.lineAt(r.to).number
```
rossjrw: Just as a general comment, could we try to avoid unnecessarily-shortened variable names? It can be difficult to work out what they mean
```suggestion
          for (const range of update.state.selection.ranges) {
            const lnStart = update.state.doc.lineAt(range.from).number
            const lnEnd = update.state.doc.lineAt(range.to).number
```

client/modules/sheaf-core/src/index.ts
```diff
@@ -0,0 +1,204 @@
...
+            if (lnStart === lnEnd) activeLines.add(lnStart - 1)
+            else {
+              const diff = lnEnd - lnStart
+              for (let i = 0; i <= diff; i++) {
+                activeLines.add(lnStart + i - 1)
```
rossjrw:
```suggestion
              for (let lineNo = 0; lineNo <= diff; lineNo++) {
                activeLines.add(lnStart + lineNo - 1)
```
Ranges in JS has always bothered me. There should be a succinct way of doing this that doesn't involve 3rd-party libs - `activeLines.extend([lnStart..lnEnd].map(n => n -1))` or something. But alas

client/modules/sheaf-core/src/index.ts
```diff
@@ -0,0 +1,204 @@
...
+  /** The `Text` object of the editor's current state. */
+  get doc() {
+    return this.view.state.doc
+  }
+
+  get scrollTop() {
+    return this.view.scrollDOM.scrollTop
+  }
+
+  set scrollTop(val: number) {
```
rossjrw, line 124: What does `val` represent?

client/modules/sheaf-core/src/index.ts
```diff
@@ -0,0 +1,204 @@
...
+    indentOnInput(),
+    bracketMatching(),
+    closeBrackets(),
+    highlightSelectionMatches(),
+    autocompletion(),
+    rectangularSelection(),
+    highlightActiveLine(),
+    EditorView.lineWrapping,
+    indentHack,
+    keymap.of([
```
rossjrw, line 188: Is there a way to generate documentation from these keymap settings? It would be nice to have a list of all configured combinations

client/modules/sheaf-core/src/print-tree.ts
```diff
@@ -0,0 +1,67 @@
+/**
+ * @file Exports the `printTree` (AST stringify) function for Sheaf.
+ */
```
rossjrw, line 3: There's a printTree in `index.ts` - is that one still in use?

client/modules/sheaf-core/src/print-tree.ts
```diff
@@ -0,0 +1,67 @@
...
+
+import { Input, NodeType, stringInput, Tree, TreeCursor } from "lezer-tree"
+
+// credit to: https://discuss.codemirror.net/u/grayen/summary
```
rossjrw: Their TOS puts user contributions under CC BY-SA-NC, I'm not sure NC is compatible here? Is this taken from e.g. a specific post or did you get help from this person personally?

client/modules/sheaf-core/tests/test-template.ts
```diff
@@ -0,0 +1,10 @@
+import * as uvu from "uvu"
+import * as assert from "uvu/assert"
+
+const Editor = uvu.suite("EditorCore")
+
+Editor.skip("test name", () => {
```
rossjrw, line 6: What is `skip`?

client/modules/wj-util/src/index.ts
```diff
@@ -156,3 +156,138 @@ export function perfy(meta?: string, threshold?: number): () => number {
...
+  return new Promise(resolve => setTimeout(resolve, ms))
+}
+
+/** Creates and returns a promise that resolves when an invokation of `requestAnimationFrame()` fires its callback. */
+export function animationFrame(): Promise<number> {
+  return new Promise(resolve => requestAnimationFrame(resolve))
+}
+
+// Credit: https://gist.github.com/beaucharman/e46b8e4d03ef30480d7f4db5a78498ca
+// Personally, I think this is one of the more elegant JS throttle functions.
```
rossjrw, line 173: Noting that I couldn't find a licence for this snippet

client/types/util.d.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * @file Misc. type declarations utilized across the repo.
+ */
```
rossjrw, line 3: Could this be switched to a regular types file with the types imported as needed rather than a .d.ts?

client/types/util.d.ts
```diff
@@ -0,0 +1,41 @@
...
+
+// TODO: determine if this file & folder is a good idea
+
+export declare global {
+  /** Represents any function, without using the {@link Function} object. */
+  type AnyFunction<T = unknown> = (...args: any) => T
```
rossjrw: I think there's a convention to use `R` for return types?
```suggestion
  type AnyFunction<R = unknown> = (...args: any) => R
```

client/types/util.d.ts
```diff
@@ -0,0 +1,41 @@
...
+    ? { 0: PromiseValue<Value>; 1: Value }[PromiseType extends Promise<unknown> ? 0 : 1]
+    : Otherwise
+
+  /** All JS primitive values. */
+  type Primitive = string | number | bigint | boolean | symbol | null | undefined
+
+  /** _Strictly_ represents a `{ 'key': value }` object with only primitives. */
+  type PlainObject = Record<string, Primitive>
+
+  /** Represents either the value `T` or the value wrapped in `PromiseLike<T>`. */
```
rossjrw:
```suggestion
  /** Represents either a value of the given type or a promise that resolves to one. */
```

rossjrw (approved):
This looks good. I appreciate your layout of future PRs, there are some comments I would've made otherwise (wj-utils is looking a little long in the tooth), but they can wait until later.

Approving, barring licensing concerns detailed below

client/modules/sheaf-core/package.json
Monkatraz: pretty sure lerna can handle that, so don't joke

client/modules/sheaf-core/src/extensions/indent-hack.ts
Monkatraz: We don't want repeats. So, if we're checking line 15, and we somehow get line 15 again, we don't want to indent line 15 twice because it was in the list twice.

client/modules/sheaf-core/src/index.ts
Monkatraz: A bunch of this code is older, or comes from the internals of CodeMirror, but yes, I normally wouldn't do that. Also, I planned to clean this up a bunch anyways - I'd rather not be in the weeds renaming variables right now.

client/modules/sheaf-core/src/index.ts
Monkatraz: Pixels, I suppose. It's the mostly sorta-dimensioned unit that you interact with the DOM using.

client/modules/sheaf-core/src/index.ts
Monkatraz: Probably, but I suspect some deduplication and "namespacing" would be required, as some commands probably overlap.

client/modules/sheaf-core/src/print-tree.ts
Monkatraz: I'm confused - the one imported in `index.ts` _is_ this one.

client/modules/sheaf-core/src/print-tree.ts
Monkatraz: Good point - it was taken from a conversation in the forums, that I did participate in. We can rewrite this function, if need be.

client/modules/sheaf-core/tests/test-template.ts
Monkatraz: Skips the test, but notes that a test was skipped in the final test report.

client/modules/wj-util/src/index.ts
Monkatraz: It seems like Github Gists have no default license - and it's in the same gray territory of using Stack Overflow comments. I'm sure it's 100% fine for this to be used - but if you're concerned about it, I guess it can be rewritten.

client/types/util.d.ts
Monkatraz: They can be imported from a `.d.ts` too, it's just that this is doing `export declare global` because some types get used so often. I was wondering what you thought about that, personally I don't mind the globals.

client/modules/sheaf-core/src/print-tree.ts
rossjrw: I never said anything 😶

client/types/util.d.ts
Monkatraz: There is no specific convention, and usually if it's just one type variable you just name it `T`. I don't really care though?

client/modules/sheaf-core/src/print-tree.ts
rossjrw: Sadly I think it does need be

client/modules/wj-util/src/index.ts
rossjrw: At least SO is explicitly CC BY-SA https://stackoverflow.com/help/licensing

I'd really prefer not to take any risks here, but this can be cleaned up later if need be

client/modules/sheaf-core/src/print-tree.ts
Monkatraz: gah I'll just rip out the entire function

client/types/util.d.ts
rossjrw: For me there's two situations:

1. Types should be kept in the sub-repository that they pertain to, except where they really do need to be cross-repository in which case they should be kept a level up. This results in lots of type definition files across several subrepos, so it's hard to tell at a glance which global definitions file a given type is coming from. In that case it's easier to go for standard types files with explicit imports.
2. There is a single global types files for all the subrepos (this is the current situation). In this case it's hard to know from the types file which if any are being used, and if so, which subrepos are using them.

Either way, for me it boils down to explicit imports being better in the long run.

Right now it's another of those things that is basically fine for now up until it isn't. It's not a big deal to navigate or to fix while there are so few of them, but if this file becomes much bigger I'd like to see it become more explicit.

Monkatraz: An alias could work here. Aliasing this folder to `@types` would let you get easy imports to all the common types. Does that sound good? (if so let's do that in another commit/PR, this is already getting kinda messy for me on my machine)

Also, FYI: I didn't want to make this a package because it wouldn't actually like, compile to anything? It would just be filler in the module list.

emmiegit: I've said before that I think we should have our aliases live under `@wikijump` for standard naming, as there will definitely be further shared libraries in the future. So this would be `@wikijump/types`.

Monkatraz:
Okay, I'm going to merge this - but I'm also going to leave a list of small things to look into changing and/or adjusting. I know this PR isn't perfect so I don't want to make it seem like the "review" is done, I just need to move forward with this process.

* Look into updating eslint rules about loops (it's weird that I can't seem to find a way to enforce `while (true)`)
* Rewrite the `debounce` and `throttle` functions
* Replace the editor tree pretty print
* Investigate a nice solution for commonly used TS types (probably `@wj/types` or `@wikijump/types`)
* Investigate module scoping, e.g. `@wikijump/util` rather than `wj-util`
* Consider adopting `<T>x` syntax, if it seems preferred
* Find some keybinding for opening the lint panel that isn't `Ctrl + L`

PR #197, opened 2021-04-18
Bring parity to seed files

emmiegit:
Not all of the seed files had the same content, missing some more recent changes. This PR simply ensures that all the `seed.sql` files mirror that of the local version.


rossjrw (approved):

emmiegit:
thanks @rossjrw

PR #202, opened 2021-04-19
[FTML-93] Fix HTML rendering issues

emmiegit:
Follow-up to #188 	

In producing HTML tests based on AST test inputs, several issues were found. [FTML-93](https://scuttle.atlassian.net/browse/FTML-93) is a collective issue to resolve these. This does not include any changes to the collapsible, that will be a separate PR.

### DIFF
--- a/ftml/src/render/html/element/text.rs
+++ b/ftml/src/render/html/element/text.rs
@@ -70,7 +70,11 @@ pub fn render_code(
     // TODO: syntax highlighting based on 'language'
 
     ctx.html()
-        .div()
+        .pre()
         .attr("class", &["code"])
-        .inner(log, &contents);
+        .contents(|ctx| {
+            ctx.html()
+                .code()
+                .inner(log, &contents);
+        });
 }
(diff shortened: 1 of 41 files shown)


ftml/src/render/html/element/text.rs
```diff
@@ -69,8 +61,10 @@ pub fn render_code(
 
     // TODO: syntax highlighting based on 'language'
 
-    ctx.html()
-        .div()
+    ctx.html() //
+        .pre()
         .attr("class", &["code"])
```
Monkatraz: IIRC, the way it was "recommended" (HTML5 spec, I think?) to handle code blocks of a specific language type was to use a class, specifically `language-xxxx` or `lang-xxxx`. Should there be a default `language-none` (or something) class for unlabeled code blocks?

Monkatraz (approved):
Looks good to me.

ftml/src/render/html/element/text.rs
emmiegit: I considered that but didn't end up adding one. I'll add it now.

emmiegit:
thanks @Monkatraz

PR #204, opened 2021-04-19
[FTML-80] Add text renderer to ftml

emmiegit:
This adds `TextRender`, which outputs a plaintext, unformatted string as output. This allows searching the final full text of the article, or assists with "printer-friendly" outputs, or permits articles to be rendered to computer terminals.

The PR has some other changes added as part of this:
* Adds `site` to `PageInfo`
* Renames `slug` -> `page` in `PageInfo`
* Moves `Handle` to be a render-level concept
* Adds some more stubs to that object
* Fixes collapsible output (see [FTML-93](https://scuttle.atlassian.net/browse/FTML-93))
* Moves URL schemes to a top-level module
* Adds plaintext test cases


Monkatraz (approved):
I can't comment too much on the lower level stuff of this PR, although considering what it adds it seems elegant.

Not that it's immediately necessary, but I don't think there is an exported WASM `render_text` function for `TextRender`.

Oh and, this seems worthy of a version bump.

emmiegit:
thanks @Monkatraz

PR #205, opened 2021-04-19
[FTML-97] Place all slog logging behind default feature

emmiegit:
This creates a default feature `has-log`, which instantiates `slog::Logger` as expected. However, if you build without default features, then all logging are replaced with no-ops. This has considerable performance improvements for the WebAssembly target, so the documentation has been updated to recommend this.

Due to the complexity of the slog logging macros, this PR does not properly "consume" unused logging variables, so instead it just builds with `RUSTFLAGS=-Aunused_variables` to suppress these warnings. Obviously during normal development, use the default target and take heed to any compiler warnings.

**Edit:** This also updates GitHub actions to build the non-logged form, and while I was at it, I properly added `RUSTFLAGS = -D warnings` for all builds (except where we're suppressing warnings obviously).

### DIFF
--- a/.github/workflows/ftml.yaml
+++ b/.github/workflows/ftml.yaml
@@ -20,7 +20,9 @@ jobs:
         with:
           toolchain: stable
       - run: cd ftml && cargo build --all-features --release
+        env: RUSTFLAGS: -D warnings
       - run: cd ftml && cargo test --all-features -- --nocapture --test-threads 1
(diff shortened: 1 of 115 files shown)


Monkatraz (approved):
That is _a lot_ of renames. We'll be able to see soon the difference this makes! (in WASM, at least)

emmiegit:
thanks @Monkatraz

PR #207, opened 2021-04-19
[WJ-456] Add the Tarnation grammar generator/system

Monkatraz:
This is _easily_ the biggest module that will be added for the initial Sheaf implementation. It is the heart and soul of how Sheaf will be parsing FTML, as the other parsing solutions available for CodeMirror 6 aren't up to the task.

It's quite complex. I do think it's coded fairly cleanly, although it's still a WIP. It's very functional, though. It'll likely get many updates as Sheaf gets new features.

There is probably too much here to grasp in one code review, and I don't really feel like skyrocketing past 150 comments. I also can't just split this up into chunks either, lol. My hope is once I get an editor instance with HTML live preview, we can relax and go over some of these modules in detail. I'm very anxious to get the editor in a state where it can actually get worked on.

### DIFF
--- a/client/modules/cm-tarnation/src/buffer.ts
+++ b/client/modules/cm-tarnation/src/buffer.ts
@@ -1,0 +1,355 @@
+import { search, SearchOpts } from "wj-util"
+import { TokenizerStack, SerializedTokenizerStack } from "./tokenizer"
+import { ParserElementStack, ParserStack, SerializedEmbedded } from "./parser"
+import type { Tree } from "lezer-tree"
+
+// ---- CONTEXT
+
+/** State storage class for Tarnation.
+ *  The parser and tokenizer should be able to restart if given
+ *  the information available in this object. */
+export class Context {
+  constructor(
+    public pos: number,
+    public tokenizer: TokenizerStack = new TokenizerStack(),
+    public parser: ParserStack = new ParserStack(),
+    public embed: SerializedEmbedded = { pending: [], parsers: [] }
+  ) {}
+}
+
+// ---- BUFFER
+
+/** Represents a Lezer token.
+ *  The `tree` value is for storing a reusable form of this token and its children. */
+type TokenData = [id: number, from: number, to: number, children: number, tree?: Tree]
+
+/** @see Buffer */
+type BufferElement = BufferToken | Checkpoint
+
+/** Stores the token and checkpoint information about the previous parse.
+ *  New tokens, and context snapshots ({@link Checkpoint}), are stored in this buffer. */
+export class Buffer {
+  /** The latest (by position) checkpoint in the buffer. */
+  declare lastCheckpoint?: Checkpoint
+
+  constructor(
+    /** The raw array storing the buffer's data. */
+    public buffer: BufferElement[] = [],
+    /** The previous `Buffer` linked, if any. */
+    public prev: Buffer | null = null
+  ) {}
+
+  /** Number of elements in the buffer. */
+  get length() {
+    return this.buffer.length
+  }
+
+  /** Fully compiles the buffer's data
+   *  into a {@link Tree `Tree.build`} compatible format. */
+  compile() {
+    // verbose approach to maximize speed
+    const buffer: number[] = []
+    const reused: Tree[] = []
+    for (let i = this.buffer.length - 1; i >= 0; i--) {
+      const node = this.buffer[i]
+      if (node instanceof Checkpoint) continue
+      const token = node.compile()
+      if (!token[4]) {
+        buffer.push(token[3], token[2], token[1], token[0])
+      } else {
+        const [, start, end, size, tree] = token
+        reused.push(tree)
+        const idx = reused.length - 1
+        buffer.push(-1, start + tree.length, start, idx)
+        // skip past cached tree
+        if (size >= 8) {
+          let left = (size - 4) / 4
+          while (left !== 0) {
+            i--
+            if (this.buffer[i] instanceof BufferToken) {
+              left--
+              // add a filler/repeat node using context hash
+              buffer.push(-2, end, start, 0)
+            }
+          }
+        }
+      }
+    }
+    // our buffer is inverted (push+reverse is faster than unshift usually)
+    buffer.reverse()
+    return { buffer, reused }
+  }
+
+  // get tokens() {
+  //   // verbose approach to maximize speed
+  //   const tokens: BufferToken[] = []
+  //   for (let i = this.buffer.length - 1; i >= 0; i--) {
+  //     const node = this.buffer[i]
+  //     if (node instanceof Checkpoint) continue
+  //     tokens.unshift(node)
+  //   }
+  //   return tokens
+  // }
+
+  /** Add a new token or context snapshot to the buffer. */
+  add(node: TokenData | Context) {
+    if (node instanceof Context) {
+      const checkpoint = new Checkpoint(node, this.lastCheckpoint)
+      this.buffer.push(checkpoint)
+      this.lastCheckpoint = checkpoint
+      return checkpoint
+    } else {
+      const token = new BufferToken(node, this.lastCheckpoint)
+      this.buffer.push(token)
+      return token
+    }
+  }
+
+  /** Get an element from the buffer. */
+  get(index: number) {
+    return this.buffer[index]
+  }
+
+  /** {@link search} comparator function.  */
+  private static _searchComparator = ({ pos }: BufferElement, target: number) =>
+    pos === target ? true : pos - target
+
+  /** Searches for the closest element to the given position. */
+  search(pos: number, opts?: SearchOpts) {
+    return search(this.buffer, pos, Buffer._searchComparator, opts)
+  }
+
+  /** Finds the closest {@link Context} behind the given `before` value,
+   *  but after the given `start` value. */
+  findContext(start: number, before: number) {
+    // binary search for closest position to `before`
+    let idx = this.search(before, { precise: false })?.index ?? this.buffer.length
+    for (; idx >= start; idx--) {
+      const node = this.buffer[idx]
+      if (node instanceof Checkpoint && node.pos <= before) {
+        return { context: node.context, index: idx }
+      }
+    }
+    return null
+  }
+
+  /** Cuts the buffer at the given position (or index). All elements after the given position will be removed. */
+  cut(at: number, indexed = false) {
+    let idx: number | undefined
+    if (indexed) idx = at
+    else idx = this.search(at)?.index
+
+    if (idx) {
+      const node = this.buffer[idx]
+      if (node instanceof Checkpoint) this.lastCheckpoint = node
+      else this.lastCheckpoint = node.checkpoint
+      this.buffer = this.buffer.slice(0, idx + (node instanceof BufferToken ? 0 : 1))
+      return this
+    }
+
+    throw new Error(`Failed to cut precisely! (${at})`)
+  }
+
+  /** Counts the number of checkpoints (context snapshots) in the buffer. */
+  countCheckpoints() {
+    if (!this.lastCheckpoint) return 0
+    let count = 0
+    let checkpoint: Checkpoint | undefined = this.lastCheckpoint
+    while (checkpoint) {
+      count++
+      checkpoint = checkpoint.prev
+    }
+    return count
+  }
+
+  /** Returns a copy of this buffer. */
+  clone() {
+    return new Buffer([...this.buffer])
+  }
+
+  // cursor(index?: number) {
+  //   return new BufferCursor(this, index)
+  // }
+}
+
+// export class BufferCursor {
+
+//   declare private buffer: BufferToken[]
+//   declare private index: number
+//   declare private token: TokenData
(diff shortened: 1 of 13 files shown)


client/modules/cm-tarnation/src/buffer.ts
```diff
@@ -0,0 +1,355 @@
...
+          }
+        }
+      }
+    }
+    // our buffer is inverted (push+reverse is faster than unshift usually)
+    buffer.reverse()
+    return { buffer, reused }
+  }
+
+  // get tokens() {
```
emmiegit: No commented out code in develop.

Either leave it unused with a comment explaining why it's not used, or add a comment here explaining when and why this block will be uncommented/added.

client/modules/cm-tarnation/src/buffer.ts
```diff
@@ -0,0 +1,355 @@
...
+    }
+    return count
+  }
+
+  /** Returns a copy of this buffer. */
+  clone() {
+    return new Buffer([...this.buffer])
+  }
+
+  // cursor(index?: number) {
```
emmiegit, line 159: Ditto re commented out code

client/modules/cm-tarnation/src/buffer.ts
```diff
@@ -0,0 +1,355 @@
...
+    }
+  }
+
+  /** Ensures that all positions in the checkpoint chain are fresh
+   *  by recalculating them. */
+  update() {
+    if (this.prev) {
+      this.prev.update()
+      this.last = this.prev.pos
+    } else this.last = 0
```
emmiegit: Is this what prettier outputs? Mismatched brace usage looks really bizarre.

client/modules/cm-tarnation/src/index.ts
```diff
@@ -0,0 +1,223 @@
...
+  nestLanguages?: LanguageDescription[]
+  /** Configuration options for the parser, such as node props. */
+  configure?: ParserConfiguration
+  /** A list of aliases for the name of the language. (e.g. 'go' -> ['golang']) */
+  alias?: string[]
+  /** A list of file extensions. (e.g. ['.ts']) */
+  extensions?: string[]
+  /** The 'languageData' field for the language.
+   *  CodeMirror plugins use this data to interact with the language. */
+  languageData?: Record<string, any>
```
rossjrw: https://xkcd.com/703/

client/modules/cm-tarnation/src/index.ts
```diff
@@ -0,0 +1,223 @@
...
+}
+
+/** Global handler for a Tarnation language.
+ *  The language constructed will not be processed until the `load` function is called.
+ *  @see TarnationLanguageDefinition */
+export class TarnationLanguage {
+  declare description: LanguageDescription
+  declare grammar?: Grammar
+  declare nodes?: NodeMap
+  declare support?: LanguageSupport
```
rossjrw, line 73: What does `declare` do in this context? (docs if possible, I couldn't find anything)

client/modules/cm-tarnation/src/index.ts
```diff
@@ -0,0 +1,223 @@
...
+   *  just return the previously loaded language. */
+  load() {
+    if (this.description?.support) return this.description.support
+    const def = isFunction(this.grammarData) ? this.grammarData() : this.grammarData
+    this.grammar = new Grammar(def)
+
+    const facet = defineLanguageFacet(this.languageData)
+    const facetProp = languageDataProp.set(Object.create(null), facet)
+
+    const nodes = (this.nodes = new NodeMap())
```
rossjrw, line 121: prettier really went to town here huh

client/modules/cm-tarnation/src/index.ts
```diff
@@ -0,0 +1,223 @@
...
+
+    return this.support
+  }
+}
+
+export class State {
+  cache = new BufferCache()
+  tokenizer = new Tokenizer(this)
+
+  constructor(public language: TarnationLanguage) {}
```
rossjrw, line 159: Oh huh, this syntax is handy

client/modules/cm-tarnation/src/index.ts
```diff
@@ -0,0 +1,223 @@
...
+      map.set(name, spec.id)
+      types.push(spec)
+      return spec
+    }
+    if (!(spec instanceof NodeType)) {
+      const id = map.size
+      const props: (NodePropSource | [NodeProp<any>, any])[] = [...(spec.props ?? [])]
+      if (spec.name && spec.name[0].toUpperCase() !== spec.name[0]) {
+        props.push(styleTags({ [`${spec.name}/...`]: tags[spec.name] ?? NodeType.none }))
+      }
```
rossjrw, line 219: Why is capitalisation on the first letter so significant?

client/modules/cm-tarnation/src/parser.ts
```diff
@@ -0,0 +1,445 @@
...
+              this.pos = context.pos
+              // if (this.debug) console.log(`starting at: ${editorContext.state.doc.lineAt(this.pos).number}`)
+              break
+            }
+          }
+        }
+      }
+    }
+
+    if (!this!.context) {
```
rossjrw: Why does TS think `this` could be null?

client/modules/cm-tarnation/src/buffer.ts
```diff
@@ -0,0 +1,355 @@
...
+  }
+
+  /** Get an element from the buffer. */
+  get(index: number) {
+    return this.buffer[index]
+  }
+
+  /** {@link search} comparator function.  */
+  private static _searchComparator = ({ pos }: BufferElement, target: number) =>
+    pos === target ? true : pos - target
```
rossjrw, line 104: Why this an arrow function but not any of the other methods?

client/modules/cm-tarnation/src/buffer.ts
```diff
@@ -0,0 +1,355 @@
...
+  /** Finds the closest {@link Context} behind the given `before` value,
+   *  but after the given `start` value. */
+  findContext(start: number, before: number) {
+    // binary search for closest position to `before`
+    let idx = this.search(before, { precise: false })?.index ?? this.buffer.length
+    for (; idx >= start; idx--) {
+      const node = this.buffer[idx]
+      if (node instanceof Checkpoint && node.pos <= before) {
+        return { context: node.context, index: idx }
+      }
```
rossjrw, line 120: I only just had this thought... do you abbreviate `index` to `idx`?

rossjrw (approved):
There are a lot of files here that are really asking to be split into multiple - particularly the `index.ts` hotpot and files that are already split into sections with comments. Happy to wait until the cleanup PR to resolve these though.

I would like to see more code documentation across the board - there are vast swathes of uncommented code here with little more than a `// filthy` to break them up

client/modules/cm-tarnation/src/index.ts
Monkatraz: it's weird how this isn't documented

Using declare here means that were just describing the type of a property of the class, rather than actually emitting an initializer.

```ts
class {
  foo: string
}
// emit
class {
  foo
}
```

```ts
class {
  declare foo: string
}
// emit
class {
}
```

client/modules/cm-tarnation/src/index.ts
Monkatraz: It's convention in grammars to have scopes/nodes to have their name capitalized if they're meaningful. It comes from Lezer, in this case.

Tarnation uses this to differentiate the special node types from custom ones, as all custom names will be capitalized. This is typed and described in the definitions file.

client/modules/cm-tarnation/src/buffer.ts
Monkatraz: It was just a convention, I believe? With a binary search you usually provide a tiny comparator function so when I factored it out into a static it stayed as one.

client/modules/cm-tarnation/src/buffer.ts
Monkatraz: In loops I tend to abbreviate the loop counter, so yeah I like `idx`. It's a little weird with the return here, but whatevs.

client/modules/cm-tarnation/src/index.ts
Monkatraz: I have the best comments. Look, it does what it says on the tin.

client/modules/cm-tarnation/src/index.ts
Monkatraz: lol I never noticed this, it makes sense though

client/modules/cm-tarnation/src/index.ts
Monkatraz: si although sometimes it's a bit weird, obviously can't use declare with it, probably not override either

client/modules/cm-tarnation/src/parser.ts
Monkatraz: Honestly I'm not sure, I probably put that there at some point when Typescript was being a big baby and being dumb. It's probably not needed now.

Monkatraz:
> I would like to see more code documentation across the board - there are vast swathes of uncommented code here with little more than a // filthy to break them up

Yeah, I agree. It was difficult to keep writing comments because this code changed A LOT. I did comment what I felt was most important, which is the glorious definitions file. But yeah stuff like the demangler needs comments

rossjrw:
I'm pretty sure that's what the same thing without `declare` does. In this
case tsc would complain if the property wasn't set in the constructor.
Maybe `declare` is for when the value of that field is set outside of the
constructor? If that's the case, probably best to remove them where they're
not needed for the extra validation

Monkatraz:
> I'm pretty sure that's what the same thing without `declare` does. In this
> case tsc would complain if the property wasn't set in the constructor.
> Maybe `declare` is for when the value of that field is set outside of the
> constructor? If that's the case, probably best to remove them where they're
> not needed for the extra validation

I mean, I've checked the output. That's what it does. Using declare avoids the performance cost of initializers.

Monkatraz:
To be clear, it's not directly documented but you can find it in the TS repo. Declare prevents TS from emitting initializers, and that's it's specific purpose. Doing anything in or outside the constructor makes no difference.

client/modules/cm-tarnation/src/parser.ts
```diff
@@ -0,0 +1,445 @@
...
+
+  pos = 0
+
+  constructor(
+    private state: State,
+    private input: Input,
+    private start: number,
+    private editorContext?: EditorParseContext
+  ) {
+    this.caching = !!editorContext?.state
```
emmiegit: nitpick: Prefer `Boolean()` over `!!`

emmiegit (commented):
I honestly don't understand most of what this does.

I think my comprehension could be helped a bit by splitting things into submodules more aggressively (as I tend to do) but there's nothing about this that's horribly weird.

Monkatraz:
> I honestly don't understand most of what this does.
> 
> I think my comprehension could be helped a bit by splitting things into submodules more aggressively (as I tend to do) but there's nothing about this that's horribly weird.

Apologies for the mixed together files - it was easier to handle at the beginning. I really do intend to organize the code a bit more.

About the function of this code: It's complicated. The most alien parts are certainly how it interfaces with CM6, as it has to be entirely incremental and also capable of representing its entire syntax tree as a flat list of nodes.

The `grammar/` folder is probably the most important, and doesn't interact with CM6 at all. All the grammar does when it's eventually compiled is inform if a given position in a string has a matched rule, and what the details of that match are.

client/modules/cm-tarnation/src/buffer.ts
rossjrw: Is this codified somewhere? If not, can it be, just so that we're all aware of it?

client/modules/cm-tarnation/src/buffer.ts
Monkatraz: No, I just missed this when I was doing the styling pass.

client/modules/cm-tarnation/src/buffer.ts
Monkatraz: Fixed, also, agree with @rossjrw that we need a style guide of some sort if we have rules like this. Not that I disagree with it - it's just that this early in development I wasn't expecting the commented code to be an issue. I made a commit to fix it.

client/modules/cm-tarnation/src/index.ts
Monkatraz: I made the comment less stupid. It wasn't obvious from the comment that this field is actually inherited from CM6's base Language class.

Monkatraz:
thanks a ton to @ammongit , @rossjrw 

Going to merge - I've realized I'll need another PR for `ftml-wasm` (worker module form of it, requires an esbuild plugin.). I've added the remaining issues to [WJ-467](https://scuttle.atlassian.net/browse/WJ-467).

I also want to say that I've kinda been doing all of this myself for a little bit now - so I'm a little exhausted (not too bad) with code, and doing a lot of little changes (e.g. polishing the monorepo) is difficult to motivate myself for right now. If anyone wants to make PRs to help with some of the other misc. tasks, that would be really helpful.

PR #208, opened 2021-04-19
[FTML-108] Resolve no-log use warnings

emmiegit:
Address warnings when compiling without logs. Warning suppression is no longer necessary.

I've created [`slog-mock`](https://github.com/ammongit/slog-mock), which allows consuming logging calls and suppressing use warnings, while still compiling to nothing.


Monkatraz (approved):
This is definitely delving into the realm of Rust where a good grasp of the language is needed to precisely understand it. Regardless, you're the "Rust person" and I trust that you know what you're doing. Thanks for all the work on getting the logging to play nice with WASM.

emmiegit:
thanks @Monkatraz

PR #209, opened 2021-04-19
[FTML-99] Use camelCase for wasm exports

emmiegit:


Monkatraz:
I'm actually gonna advocate for _not_ changing the style here.

1. It's not that weird to use snake_case in JS, hell some people prefer it (I would if the JS standard library wasn't camelCase)

2. I think if you make this change for the function names, it would probably make sense for the serialized objects to use camelCase and not kebab-case

Also, with JS, it would probably make more sense to capitalize abbreviations in their entirety, e.g. `renderHTML`. Objects such as the native `DOMParser` demonstrate this convention.

emmiegit:
Alright, if this is preferable for you.

PR #210, opened 2021-04-19
[FTML-107] Limit number of lines per block

emmiegit:
To avoid OOM-based DOS attacks, this limits the number of newlines created per `[[newlines]]` block to 100, and adds a test for this.


Monkatraz (approved):
Looks good to me.

The warning emitted by this change does make me wonder if `make_warn()` should accept an optional message - I can imagine that in the editor, getting the warning `block-malformed-arguments` would be fairly ambiguous without additional work taken by the editor to "guess" what emitted the warning.

emmiegit:
I agree that warnings are very limited. I'm stayed away from string messages because it can lead to hardcoded English and difficulty for automated systems (like editors). I envision we'll have an epic later that fleshes them out more, to make the editing experience richer, but that's down the line.

emmiegit:
thanks @Monkatraz

Monkatraz:
> I agree that warnings are very limited. I'm stayed away from string messages because it can lead to hardcoded English and difficulty for automated systems (like editors). I envision we'll have an epic later that fleshes them out more, to make the editing experience richer, but that's down the line.

How do I keep forgetting about translations? I feel like I'm being affected by Wikidot.

Yeah, in that case, I'm not sure of the best solution. If I had a naive attempt at a solution, it would be something like `block-malformed-arguments` having "metadata"? e.g. `{ argument: "newlines", issue: "exceeded-limit" }`. It would still require each warning type to have a fair amount of processing, but there wouldn't be any parsing of the source by the editor. Far off in the future for this issue, probably.

emmiegit:
Yeah definitely. We can go through and audit all the "generic" warnings and provide more context. Produce a big fancy spec where warnings have a lot of nice useful structured data.

PR #211, opened 2021-04-20
[FTML-82] Change "modifier" to be a block property rather than ad-hoc derived.

emmiegit:
Previously we did `name.ends_with('_')` to determine if the modifier flag should be set. This PR changes the block rule so that it's derived during block parsing and handed as an argument.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #212, opened 2021-04-20
Update FTML compile settings (WASM-specific)

Monkatraz:
Removes the recommendation to compile with `opt-level = 's'`, as this caused a fairly significant performance hit if done. This also adds the `wasm-opt` `-c` flag, which causes the optimizer to be ran multiple times until the file size converges. This has some nice effects on binary size.


emmiegit (approved):

Monkatraz:
thanks @ammongit

PR #214, opened 2021-04-20
[WJ-456] Add a web-workerized version of ftml-wasm

Monkatraz:
This adds a web-workerized version of `ftml-wasm` as a package. Getting this to work I'm fairly sure hurt my soul.

The code is a bit messy, and there are some questions about how we want to handle "worker modules" in the future. I wanna work this out with @rossjrw later, but for right now it'll have to do.


client/modules/ftml-wasm-worker/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/scpwiki/wikijump.git"
+  },
+  "bugs": {
+    "url": "https://scuttle.atlassian.net/servicedesk/customer/portal/2"
+  },
+  "homepage": "https://github.com/scpwiki/wikijump#readme",
```
emmiegit, line 23: Not relevant to this PR, but this reminds me that at some point we're going to need to change all these to https://www.wikijump.org/

client/modules/ftml-wasm-worker/src/worker/ftml.worker.ts
```diff
@@ -0,0 +1,77 @@
...
+  },
+
+  async renderHTML(raw: ArrayBuffer) {
+    await ready
+    const str = decode(raw)
+    const { html } = FTML.render(str)
+    return transfer(html)
+  },
+
+  async renderStyle(raw: ArrayBuffer) {
```
emmiegit: What is this? Doing the rendering but only extracting the CSS?

I think I've seen this elsewhere, why can you run the render once and then use both the HTML and CSS?

emmiegit (approved):
Like other JS-intensive stuff, I don't really understand most of this, but I don't see anything wrong on the ftml part

client/modules/ftml-wasm-worker/package.json
Monkatraz: Yeah a lot of the "meta" client repo stuff will need to be looked at. The repo is in a very nebulous state right now

client/modules/ftml-wasm-worker/src/worker/ftml.worker.ts
Monkatraz: I'll have to see how it plays out - but the idea is that when you're just transferring strings back and forth, you can use `ArrayBuffer` encoded strings to transfer between the web-worker and the main thread with very low latency and performance cost.

To explain a bit further, when you "return" an object to the main thread, or the other way around, the browser has to do something called a "structured clone", which duplicates the entire object. When you "transfer" an `ArrayBuffer`, that doesn't happen.

Considering that for SCP-1730 the HTML is ~220KB, I don't want to be cloning objects constantly that are of that size, or larger.

emmiegit: Makes sense.

But would it be possible to have two `ArrayBuffer`s, one for HTML and one for CSS? Then you only do a single request and get both bits of data.

Monkatraz: Maybe? Honestly, I haven't thought about it. I'm not sure if you can do that in one "return", at least with `threads`.

(`threads` is the NPM package I'm using to abstract web-worker handling - it's very nice for handling a lot of the agonizing chores you have to do.)

Monkatraz: image

Oops. Yeah, you can. Let me add that.

Monkatraz: image

There we go.

client/.eslintrc.js
```diff
@@ -193,6 +193,12 @@ module.exports = {
       parserOptions: { createDefaultProgram: true },
       rules: { ...baseRules, ...typeRules }
     },
+    // TypeScript (Worker)
+    {
+      files: ["*.worker.ts"],
+      env: { worker: true, es2021: true },
+      rules: { ...baseRules, ...typeRules }
+    },
```
rossjrw, line 201: I can't believe we didn't think to do this before

client/modules/ftml-wasm-worker/src/index.ts
```diff
@@ -0,0 +1,168 @@
...
+}
+
+/** Tokenizes a string of wikitext. */
+export async function tokenize(str: string) {
+  type Return = Binding.IToken[]
+  return await invoke<Return>(() => module.worker.tokenize(transfer(str)))
+}
+
+/** Parses a string of wikitext. This returns an AST and warnings list, not HTML.
+ *  @see renderHTML */
```
rossjrw: Should this be `render`?

client/modules/ftml-wasm-worker/src/worker/ftml.worker.ts
```diff
@@ -0,0 +1,72 @@
+/**
+ * @file Web-workerized version of "ftml-wasm".
+ */
+
+import * as FTML from "ftml-wasm"
+import { expose, Transfer, encode, decode } from "./lib"
+
+const ready = FTML.loading
```
rossjrw, line 8: What is this?

client/types/files.d.ts
```diff
@@ -0,0 +1,59 @@
...
+
+declare module "*.svg" {
+  const ref: string
+  export default ref
+}
+
+declare module "*.bmp" {
+  const ref: string
+  export default ref
+}
```
rossjrw, line 29: 2005 called

rossjrw (changes requested):
Should `ftml_wasm/vendor` be committed? Seems to me that it should have been ignored

Monkatraz:
> 
> 
> Should `ftml_wasm/vendor` be committed? Seems to me that it should have been ignored

Did you miss the whole ordeal about this? There is a whole bunch of issues with attempting to build the FTML WASM binary automatically. Instead, it's vendored into the module and there is a test that automatically checks if it's out of date. This is so that:
* You don't need to install Rust in order to compile the JS
* Ultra-runner doesn't have to somehow track files outside of module
* Builds won't break due to FTML updating, instead they'll just be out of date
* Builds won't take over a minute due to FTML compiling
* The build process won't be complicated by somehow handling the Rust compile

client/types/files.d.ts
Monkatraz: lol yeah I copied this from Mainframe, which copied it from a Snowpack template. I'm not sure why `bmp` is in there but whatever lol

client/modules/ftml-wasm-worker/src/worker/ftml.worker.ts
Monkatraz: `FTML.loading` is simply a promise that resolves when the library has fully instantiated the WASM module.

client/modules/ftml-wasm-worker/src/index.ts
Monkatraz: Yes, good catch.

client/modules/ftml-wasm-worker/src/worker/ftml.worker.ts
rossjrw: Could you document that?

Monkatraz: image

It, uh, is documented.

rossjrw:
I did miss the whole ordeal about this, yeah. Given that many concerns that does seem fine to keep that for now - I just thought it was a gitignore oversight.

I do think that these complications are things that will need to be addressed in the future. The user *should* have to install Rust to compile it, and this should theoretically be happening in a Docker container anyway so it shouldn't be that big a deal. We can cache super-aggressively so that it only has to be built once.

But these are problems for not-now and this is a fine compromise in the meantime. I would like this temporary build process to be documented somewhere, though, because not all future contributors will have been present for this whole ordeal. (If it is already documented somewhere, I am a dum dum, please point me to it)

Monkatraz:
I do think requiring a contributor to install Rust so they can make some edits to a module is a tad annoying. It would also be required for development builds.

Also, yeah, sure, it should be documented. But I've been doing everything myself and writing up a bunch of documentation about the current architecture of the repo is rather hard to motivate myself for right now. I'm really trying to get a version of Sheaf working so that FTML, this repo, and ofc Sheaf itself can all be developed on and tested. Right now, it will be the biggest implementation of all the "new technology".

If you want to make a PR to do _any_ of these small tweaks, documentation files, etc. that would be helpful.

rossjrw (approved):

Monkatraz:
thanks @ammongit , @rossjrw

PR #215, opened 2021-04-20
Ozone/Laravel Integration

danieltharp:
This PR is a pretty significant one. Instead of running all requests through Ozone unless an nginx route told it to go to Laravel, now Ozone is nothing more than a controller within Laravel, to be called when a request is made without a matching route. The benefit of this is that we can now use the full Laravel framework _within_ the Ozone framework (and vice versa). The first piece of that integration has been published here, any web request will check if the user has a valid Ozone session and log them in as a Laravel user on the fly.

The nginx config has been updated, and the files renamed to indicate their respective usage now. There is some unrelated cleanup to the ParameterList.php file which happened while troubleshooting the integration, but they're all stuff suggested by automated code inspection.


web/lib/ozoneframework/php/core/ParameterList.php
```diff
@@ -36,46 +36,47 @@ public function initParameterList($runData) {
 			$qs =  $_SERVER['QUERY_STRING'];
 			/* Check if there is a "?" char - if so, remove it. */
 			$qs = preg_replace('/\?.*$/', '', $qs);
-			$splited = explode("/",$qs);
+			$splited = explode('/',$qs);
```
emmiegit: nitpick:
```suggestion
			$splited = explode('/', $qs);
```

web/lib/ozoneframework/php/core/ParameterList.php
```diff
@@ -36,46 +36,47 @@ public function initParameterList($runData) {
...
+                $uriPairs = explode('&', $uri);
+                foreach ($uriPairs as $uriPair) {
+                    $u = explode('=', $uriPair);
+                    $key = $u[0];
+                    $value = $u[1];
+                    $this->parameterArray[$key] = urldecode($value);
+                    $this->parameterTypes[$key] = 'GET';
+                    $this->parameterFrom[$key] = 0;
+                    $this->allParameters['GET'][$key] = urldecode($value);
+                }
```
emmiegit, line 59: This section appears extra indented for some reason

web/templates/layouts/WikiLayout.tpl
```diff
@@ -48,7 +48,7 @@
 
  	<meta http-equiv="content-type" content="text/html;charset=UTF-8"/>
     <meta http-equiv="content-language" content="{$site->getLanguage()}"/>
-
+    <meta name="csrf-token" content="{php}echo(csrf_token());{/php}"/>
```
emmiegit, line 51: Is `{php}...{/php}` different from `{...}`?

emmiegit (approved):
:tada:

web/templates/layouts/WikiLayout.tpl
danieltharp: It seems to be, the same command that way threw an error, I believe that syntax is just for variables that were passed to the template.

web/lib/ozoneframework/php/core/ParameterList.php
danieltharp: It actually is as intended, the diff view doesn't show quite how deep into nested statements that bit is.

danieltharp:
I'm gonna be honest. For me, this was the single most important part of the PHP side, making these two codebases talk to each other and allowing for bidirectional use of new and legacy classes. Everything from here is completely doable. I am elated.

emmiegit:
Yeah, this is a tremendous change and it gives me a lot of hope and excitement despite the quantity of Wikidot PHP we have yet to go through.

PR #216, opened 2021-04-21
Exclude cli calls (artisan) from Ozone init

danieltharp:
Hotfix to allow CI/CD to work again. It runs `php artisan key:generate` during image building which was inadvertently starting an Ozone session, which goes and looks for memcached, which isn't present when you're building an image.


emmiegit (approved):
:wrench:

PR #218, opened 2021-04-21
Add basic documentation for ftml-wasm package

rossjrw:
This is something closer to what I'd like to see for each new package coming into the `client/` subrepo -

- A brief description of what it is
- Any special notes about it that affect development

That's pretty much it. More detailed stuff like usage, build instructions, etc can come a little later except where they're a) imminently important b) very hard to work out.

~~Merges into #214.~~ oops

### DIFF
--- a/client/modules/ftml-wasm/README.md
+++ b/client/modules/ftml-wasm/README.md
@@ -1,0 +1,13 @@
+# ftml-wasm
+
+This package contains the WebAssembly compilation of FTML.
+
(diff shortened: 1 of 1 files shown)


Monkatraz (approved):

emmiegit:
This diff is a bit weird, it seems one of these branches has a different basis. I believe the easiest way to fix is to adjust this branch using `git rebase ftml-wasm-worker`.

emmiegit (approved):

PR #219, opened 2021-04-21
Reorganize some client/* folders into misc/*

Monkatraz:
This cleans up the root level of `client/` a little bit. Also, linting tools will ignore the `misc/` folder.


emmiegit:
Does "dev" here mean "local"? I remember discussing using the latter term to avoid confusion with the "dev" environment (https://www.wikijump.dev/)

Monkatraz:
Uh, no, in this context it means "development resources". If you have a better name, I'm welcome to hear it.

emmiegit:
Maybe `developers/`? Or `develop/` similar to the branch name?

Monkatraz:
I'm not sure if lengthening the name adds anything substantive - plus it just makes typing out directories a little more verbose, lol. It would have to be something like `resources/`, although that implies assets, which is why I didn't choose that.

emmiegit:
The other name I was thinking of was `install/`, but like `resources/` that doesn't really fit.

Maybe `templates/` would be better here? At least for the stuff you have to copy, then `docs/` remains where it is.

Monkatraz:
Well, `types/` goes in there too - they don't fit into any other directory (except maybe as a weird, hardcoded module).

Monkatraz:
I do expect more to eventually go into this folder, maybe `examples/`, or maybe even configuration settings. It's kind of a dumping ground for misc. monorepo stuff that isn't directly part of packages.

emmiegit:
Then `misc/`? lol

I just don't really want `dev` confusion as discussed, basically any other name would be fine with me.

Monkatraz:
yeah you know what `misc/` does fit. we've successfully talked the solution out of myself

emmiegit (approved):

rossjrw (approved):

PR #220, opened 2021-04-22
Client monorepo: Change to Vite, fix workers, and more (sorry)

Monkatraz:
This PR is a nightmare. I'm sorry. Basically, this is about... let's say, three different branches merged into one due to the nightmare that it was to figure all of this out?

Here is what the PR does, in total:
* Adds the most basic form of the FTML language package, which is just the grammar.
* Switches `web/dev-sandbox` to use Vite instead of Snowpack
* Fixes up some dependencies
* Changes how workers work, they're now inlined rather than imported with a URL reference
* References to Snowpack are removed, website documentation is removed

The removal of the `web/` documentation is because `client/web/` is likely to go unused in the near future. Instead, the repo will likely connect with `./web/` (root) and treat that as a Lerna package, and that'll be the frontend. This is due to the experiences learned in the `laravel-vite` branch.

What do we get out of all of this?
* Workers work correctly.
* `dev-sandbox` builds very quickly, with no fuss, works perfectly with the monorepo
* A solution has been found for integrating with Laravel
* I can finally return back to work on the damn editor


emmiegit:
Does it make sense to treat something in the `..` as a lerna module? Are we able to make a subdirectory within `client` that has whatever resources are needed in `web`?

Monkatraz:
> 
> 
> Does it make sense to treat something in the `..` as a lerna module? Are we able to make a subdirectory within `client` that has whatever resources are needed in `web`?

Does it make sense? Uh... maybe? Does it work? Yes.

About the subdirectory, yes, it's totally possible - however, by treating `../web/` as a module, you get the benefits of `laravel-vite`, such as HMR for blade template files, aka live-reload on save. The way this would all fit together really depends on Vite being in the Laravel directory.

emmiegit (approved):

client/modules/cm-lang-ftml/src/index.ts
```diff
@@ -0,0 +1,615 @@
...
+
+        [/\d+/, "t.unit"],             // numbers
+        [/@texSymbols/, "t.operator"], // operators
+        [/@texBrackets/, "@BR"]        // brackets
+      ]
+    }
+  })
+})
+
+export const FTMLLanguage = new TarnationLanguage({
```
rossjrw, line 64: FTML is canonically spoken in a *drawl*

rossjrw (approved):

client/modules/cm-lang-ftml/src/index.ts
Monkatraz: 'wot in tarnation

Monkatraz:
thanks @ammongit , @rossjrw

PR #221, opened 2021-04-22
[WJ-474] Create a component library

rossjrw:
This PR is for the creation of a library of Svelte components for use by the frontend. [WJ-474](https://scuttle.atlassian.net/browse/WJ-474)

- [x] Create a component library
- [x] Set up a build process for the components
- [x] Document stub replacement components
- [x] ~~Set up a process for creating stub replacement components and maybe think of a better name for them~~ Deferring for later (WJ-476)
- [x] Set a baseline for separate functions that are not components e.g. portal/teleport, and also document them & their usage
- [x] Test that components work (with actual tests)
- [x] ~~Test that stub replacements work~~ Deferring for later (WJ-476)

This PR will not actually add any components beyond what is necessary to test that the library and the build process work.

Issues that need creating in the JIRA at some point:
 
- [x] Component library (WJ-474)
- [x] Working out how to make the components play nice with [Storybook](https://storybook.js.org/) (WJ-475)
- [x] Stub components (WJ-476)


Monkatraz:
just leaving some notes here:
* consider using [testing-library-dom](https://testing-library.com/docs/dom-testing-library/intro) - intended for Jest, but doesn't have to use it
* figure out a nice way of cleaning the `tests/dist` folder

Monkatraz:
Okay, @rossjrw I think this is mostly up to you now. I added a basic `portal` function in the `lib` folder. I don't think it really needs to get anymore complicated than that. With that added, all that is left is handling stub components.

This blocks a few things I wanna do, mainly with Sheaf's various components, so if we wanna fully nail down stubs later that would be fine with me. It's going to be a while before we can use them anyways.

Monkatraz (approved):

PR #222, opened 2021-04-22
[FTML-73] Update paragraph behavior

emmiegit:
Paragraphs have always been a difficult part to deal with, but after years of work on ftml, this finally brings them in line with a sensible subset of Wikidot's behavior.

One issue with the previous implementation was that, while it properly separated paragraphs based on `Token::ParagraphBreak` (`\n\n`), it wrapped _everything_ in a paragraph. This is not correct (although is what Wikidot technically does, before HtmlPurifier fixes it).

`<p>` elements can only contain inline elements, such as text or `<span>`s. They may _not_ contain `<div>`s, `<blockquote>`s, or other `<p>`s.

This modifies `ParseResult` to introduce the concept of "paragraph safe". If this is true, it means it can be included in a `<p>`. Otherwise, the `ParagraphStack` will treat it as its own "top-level" object during paragraph construction. Combined with existing anti-empty paragraph logic, this properly replicates what should happen with paragraph-expectant syntactical elements within our wikitext. The `ok!()` macro was also modified to allow passing paragraph safety explicitly (if it cannot be inferred, see below). It is marked with a `;` divider instead of `,` to avoid syntactical ambiguities.

To assist with this, `Element::paragraph_safe()` and friends exist to output if this element type is safe to output. It however, does _not_ check this deeply, only at a surface level. To avoid O(n^2) behavior from repeated deep checks, as mentioned above, we have `ParseResult` return whether all of the elements inside are paragraph safe or not. The helper methods on `Element` etc. exist to assist with the `ok!()` macro, to avoid trivial cases where it's known to be safe or not, and collecting paragraph safety from leaf elements.

This also modifies all of the AST tests, as they all assumed everything would be naïvely wrapped in a `<p>`. They no longer do, and output much more sensible output. Likewise, nested divs, blockquotes, etc., only emit paragraphs for elements that can accept them, similar to well-formed Wikidot behavior. We also see the same with `[[div_]]` and other contexts without internal paragraphs, where child elements are not wrapped and appear as if `<div>` was used directly.

I know paragraphs are a strange and complicated topic, so I'm happy to answer any questions you may have about them in chat.


Monkatraz (approved):
well this was a lot of work

emmiegit:
thanks @Monkatraz

PR #223, opened 2021-04-23
[WJ-401] Enable FFI in our PHP containers

emmiegit:
As a pre-requisite to [WJ-401](https://scuttle.atlassian.net/browse/WJ-401) (creating an FFI for ftml for PHP), we need to enable FFI within PHP.


Monkatraz (approved):
This isn't my realm at all, but this _seems_ simple and my approval would unblock merging :P

emmiegit:
thanks @Monkatraz

PR #224, opened 2021-04-23
[FTML-113] Change [[hidden]] and [[invisible]] to return divs, not spans

emmiegit:
This way they can contain modules, blockquotes, or other divs inside without causing HTML structure issues.


Monkatraz (approved):

emmiegit:
In updating the tests for this PR, I've discovered some additional complexity that prevents this from being a simple issue like I intended. Closing for now, may revisit.

PR #226, opened 2021-04-23
Export render_text()

emmiegit:
These `pub use`s aren't actually utilized, but as convention here I'm making anything being exposed to the wasm `pub` and used, as if a Rust crate was depending on this. This PR adds `render_text()` as an export, as it is presently missing.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #228, opened 2021-04-25
[WJ-401] C FFI for ftml

emmiegit:
This PR adds an FFI for ftml. This C FFI (foreign function interface) allows any language which permits C API-compatible binding to run ftml functions within the same process. It only exposes the main rendering methods, to avoid excess FFI memory copying for each step. Consequently, it exposes parse warnings in the HTML or text output in addition to the regular fields.

This does not add any FFI interaction from PHP, it is purely for ftml itself.


ftml/README.md
```diff
@@ -36,8 +36,16 @@ You can use this as a dependency by adding the following to your `Cargo.toml`:
 ftml = "0.4"
```
Monkatraz, line 36: Should probably update this in the readme, lol

Monkatraz (approved):
This looks like it wasn't fun to do.

ftml/README.md
emmiegit: I haven't deployed any new versions of ftml to crates.io

Monkatraz: Oops, nevermind then.

emmiegit:
thanks @Monkatraz

PR #230, opened 2021-04-26
[WJ-478] HTML attribute safety

emmiegit:
This PR removes `sandbox`, an unsafe HTML attribute. It also modifies how handling works for boolean HTML5 attributes which base their logic off of the presence/absence of a solo key rather than `key="value"`. There is also a test for this behavior.


Monkatraz (approved):
Looks good to me.

emmiegit:
thanks @Monkatraz

PR #231, opened 2021-04-26
[WJ-456] Add FTML CM6 language module

Monkatraz:
Considering this is at +2000 lines, I don't think I can really add more to this and not feel dirty.

This PR adds quite a few things - but in general these things focus around the functionality of the editor. In particular:
* Editor theme defined in SCSS - not JS
* FULL AUTOCOMPLETE. Autocomplete is available for blocks, modules, block and module arguments, HTML attributes, values of HTML attributes, and values of block and module arguments.
* Linting, using FTML's native warnings

Additionally, this PR adds functionality for using Svelte components inside of CodeMirrors DOM. This is used for the tooltips, and will probably be _very handy_ in the future.

There is also quite a few general tweaks and changes around the codebase - I was resolving issues as I found them.

What's next? Probably the live-preview component and then a large codebase cleanup.


client/package.json
```diff
@@ -46,27 +46,28 @@
   "bugs": "https://scuttle.atlassian.net/servicedesk/customer/portal/2",
   "homepage": "https://github.com/scpwiki/wikijump#readme",
   "devDependencies": {
+    "@ltd/j-toml": "^1.12.2",
```
rossjrw, line 49: We already have `toml`, why this as well?

client/modules/wj-css/src/components/_sheaf.scss
```diff
@@ -0,0 +1,219 @@
...
+  50% { opacity: 0; }
+  100% {}
+}
+
+// prettier-ignore
+@keyframes cm-blink2 {
+  0% {}
+  50% { opacity: 0; }
+  100% {}
+}
```
rossjrw, line 217: This seems to be the same as cm-blink1

client/modules/wj-css/src/components/_sheaf.scss
```diff
@@ -0,0 +1,219 @@
...
+
+.cm-editor .cm-panel {
+  &.cm-panel-lint ul {
+    max-height: 16rem;
+    outline: none;
+    padding-right: 0.5rem;
+
+    // use duplicate selector so we get more specificity
+    // the CodeMirror base theme overrides otherwise
+    &:focus [aria-selected][aria-selected] {
```
rossjrw, line 184: Wow I've literally never thought to do this

rossjrw (approved):

client/modules/wj-css/src/components/_sheaf.scss
emmiegit: What does the duplicate selector do?

client/package.json
Monkatraz: This replaces the `toml` package in the dependencies. `toml` can only parse up to (iirc) TOML v0.4 rather than v1.0 like this one.

client/modules/wj-css/src/components/_sheaf.scss
Monkatraz: Sorry, I should've left a comment here. This is a CodeMirror trick - it uses two identical animations so that it can restart and pause CSS animations (by alternating them) and not have weird effects happen. It's weird technical stuff.

client/modules/wj-css/src/components/_sheaf.scss
Monkatraz: CSS selectors have a certain way that they cascade and conflict (override) each other. This is their _specificity_, and it's a source of eternal headache. If two selectors have the same specificity, they'll go off of declaration order. Otherwise, the higher specificity wins. Usually, the more complex a selector is, the higher specificity it has. So by repeatedly selecting `[aria-selected]`, we increase the specificity and override the CodeMirror base theme.

emmiegit: Interesting

Monkatraz:
I'm going to merge - @ammongit if you want to review after merge I'll make a PR with your requested changes in them. Unfortunately everything with the monorepo rn is tangled up because the infrastructure isn't done, so nearly every PR blocks every other PR lol

thanks @rossjrw

PR #234, opened 2021-04-26
Change ftml crate type to dylib

emmiegit:
When building on Docker:
```
warning: dropping unsupported crate type `cdylib` for target `x86_64-unknown-linux-musl`

warning: 1 warning emitted
```

### DIFF
--- a/ftml/Cargo.toml
+++ b/ftml/Cargo.toml
@@ -14,7 +14,7 @@ edition = "2018" # this refers to the Cargo.toml version
 
 [lib]
 name = "ftml"
-crate-type = ["lib", "dylib, "cdylib"]
+crate-type = ["lib", "dylib", "cdylib"]
 
 [features]
 default  = ["ffi", "log"]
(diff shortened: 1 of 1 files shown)


danieltharp:
If this works under musl that would be ideal, since that's all that's available to the php-fpm (alpine) container presently. If we end up needing glibc the mainline php-fpm container offers it.

emmiegit:
Yeah, I'm trying to build in Docker and apparently there's a meaningful distinction between `dylib` and `cdylib`.

Unfortunately CI is catching a regression when building to wasm:

```
Error: crate-type must be cdylib to compile to wasm32-unknown-unknown. Add the following to your Cargo.toml file:

[lib]
crate-type = ["cdylib", "rlib"]
```

So that's fun. Time to look into the specifics of how crates are built so I can get this to support both.

emmiegit:
Compiling to target both musl and produce a cdylib is seemingly rather difficult, due to some contradictions between the two.

It's a bit annoying since having a cdylib target is primarily because we need to target wasm, which is a separate build from anything that targets musl...

emmiegit:
I think this can be resolved by making a subcrate for wasm.

It might also make to make ffi a separate crate? But there's some internal stuff that will need consideration.

PR #235, opened 2021-04-26
[WJ-464] Build ftml in php-fpm containers

emmiegit:
This modifies the `Dockerfile`s for php-fpm such that they target Debian rather than Alpine. This way we can build ftml against glibc and copy the artifacts into the php-fpm container for use. Additionally, the repo is cloned in the rust step and `web` is copied into the final php-fpm container.


rossjrw (approved):

danieltharp (approved):

emmiegit:
thanks @rossjrw @pxdnbluesoul

PR #238, opened 2021-04-29
[WJ-481] Add modules.toml

emmiegit:
Add static module information, as well as minor tweaks to `modules.schema.toml`.

### DIFF
--- a/ftml/conf/modules.toml
+++ b/ftml/conf/modules.toml
@@ -6,7 +6,7 @@ page = { type = "string" }
 [Categories]
 body = "none"
 [Categories.arguments]
-includeHidden = { type = "bool" }
+includeHidden = { type = "bool", default = false }
 
 [CSS]
 body = "raw"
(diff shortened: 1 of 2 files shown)


ftml/conf/modules.toml
```diff
@@ -0,0 +1,28 @@
...
+body = "none"
+html-attributes = true
+[Join.arguments]
+button = { type = "string", default = "" } # default message depends on language
+
+[PageTree]
+body = "none"
+[PageTree.arguments]
+root = { type = "string" }
+depth = { type = "int", min-value = 1, default = 10000 } # really unlimited, but not sure how to convey that as an int
```
rossjrw, line 24: -1?

rossjrw (approved):

ftml/conf/modules.toml
Monkatraz: I could see the -1 having issues? While it's okay here because the minimum value is above 0, if it wasn't above 0 it would be a valid default. I think having a weird check for that is kind of meh. Maybe just have a special "inf" value or something? Or just leave it like this, I'm not sure how big a deal this is.

Monkatraz (approved):

emmiegit:
thanks @rossjrw @Monkatraz

PR #241, opened 2021-04-30
[FTML-103] Map parse warnings to UTF-16 indices

emmiegit:
One complication with ftml-wasm is that Javascript uses UTF-16 while Rust uses UTF-8. To make warning indices more useful, I have added functionality that maps one to another, and then exports the UTF-16 versions to Javascript.


Monkatraz (approved):
Looks good to me.

Only thing I can think of is that it may be handy to expose these functions to WASM, just for future-proofing. I don't know currently where else they would be used practically, but this could be used to remap the AST or the token stream. I know the AST doesn't have positional information currently but it may later if some basic form of source mapping is added. Speaking of, there probably should be an FTML issue for that.

emmiegit:
Okay that sounds fine. In this case I'll just expose this structure publicly, regardless of target arch.

emmiegit:
```typescript
export class Utf16IndexMap {
  free(): void;
/**
* @param {string} text
*/
  constructor(text: string);
/**
* @returns {Utf16IndexMap}
*/
  copy(): Utf16IndexMap;
/**
* @param {number} index
* @returns {number}
*/
  get_index(index: number): number;
/**
* @param {number} start
* @param {number} end
* @returns {[number, number]}
*/
  get_span(start: number, end: number): [number, number];
}
```
generated interface for `Utf16IndexMap`

emmiegit:
thanks @Monkatraz

PR #242, opened 2021-04-30
[FTML-10] Normalize URLs in rendering

emmiegit:
Make URLs more clear, especially for text. Since we don't have percent encoding, some browsers might not properly handle these non-normal URLs.

This creates a helper function to perform this properly, and updating the HTML/text tests.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #243, opened 2021-05-01
[FTML-110] Add image block

emmiegit:
This adds support for [images](https://www.wikidot.com/doc-wiki-syntax:images) in ftml. Bumps version to v0.10.1.

List of changes:
* Adds `Element::Image`
* Adds `BLOCK_IMAGE`
* Adds HTML and text rendering for `Element::Image`
* Changes the `PageInfo` data for tests
* Remove `data` module:
  * `PageInfo` is moved to a root-level module
  * Alignment structures are moved to `tree` along with other syntax tree-related definitions
* Adds tests for alignment parsing
* Adds `ImageSource` as a way of representing and parsing the source of images
* Adds `get_image_link()` method to handle for getting full file URLs
* Spruce up `ParseCondition` a bit
* Remove duplicated / unused conditions from the block parser
* Adds tests for the new syntax.


ftml/src/render/handle.rs
```diff
@@ -81,9 +81,9 @@ impl Handle {
 
         let (site, page, file): (&str, &str, &str) = match source {
             ImageSource::Url(url) => return Cow::clone(url),
-            ImageSource::File { file } => (&info.site, &info.page, &file),
-            ImageSource::OtherFile { page, file } => (&info.site, &page, &file),
-            ImageSource::RemoteFile { site, page, file } => (&site, &page, &file),
+            ImageSource::File1 { file } => (&info.site, &info.page, &file),
+            ImageSource::File2 { page, file } => (&info.site, &page, &file),
+            ImageSource::File3 { site, page, file } => (&site, &page, &file),
```
rossjrw: This seems like a regression - FileX doesn't have any inherent meaning, but the old naming scheme at least had some indication of the context. Why the change?

rossjrw (approved):

Monkatraz (approved):

ftml/src/render/handle.rs
emmiegit: I was writing out the JSON, and I thought it looked more clear to have the variants, since the fields are descriptive as to what this source is. "OtherFile" in isolation seemed a bit weird to me.

I can always change this later, in case there are strong feelings on this subject.

emmiegit:
thanks @rossjrw @Monkatraz

PR #245, opened 2021-05-02
[FTML-120] Add alignment blocks

emmiegit:
This PR adds alignment blocks (e.g. `[[>]]`, `[[==]]`) to ftml.

Apparently I had separate tokens for for each of those alignment block heads, but this is unnecessary since our block parser logic already handles non-alphanumeric block names just fine.

I make them separate rules generated via macro rather than a single rule that parses out the name because then you can mix and match ending blocks, which isn't very sensible.


ftml/src/tree/align.rs
```diff
@@ -39,6 +39,15 @@ impl Alignment {
             Alignment::Justify => "justify",
         }
     }
+
+    pub fn html_class(self) -> &'static str {
+        match self {
+            Alignment::Left => "alignleft",
+            Alignment::Right => "alignright",
+            Alignment::Center => "aligncenter",
+            Alignment::Justify => "alignjustify",
```
rossjrw, line 48: Just a thought, how come these are in tree instead of render/html?

rossjrw (approved):

ftml/src/tree/align.rs
emmiegit: Because they're methods on AST structures. See also `HeadingLevel::html_tag()`, `AnchorTarget::html_attr()`, etc.

emmiegit:
thanks @rossjrw

PR #246, opened 2021-05-02
[FTML-117] Implement iftags

emmiegit:
This PR implements the `[[iftags]]` block, which conditionally adds the contents of the block depending on what tags are present.

This is implemented by adding `Element::IfTags`, which describes the conditions and contents, and the renderer uses `PageInfo.tags` to determine if the contents should be rendered, or simply omitted. 
It also adds some tags to the test data so we can test this functionality.

The `[[ifcategory]]` block is coming after this, since it is very similar.


rossjrw (approved):

emmiegit:
thanks @rossjrw

PR #248, opened 2021-05-02
[WJ-452] Add crossorigin to images and iframes

emmiegit:
Complement to #249 

Since all images and iframes are going to be hosted on wjfiles, which is not the same site, all tags needs `crossorigin` so they will be loaded.

### DIFF
--- a/ftml/src/render/html/element/iframe.rs
+++ b/ftml/src/render/html/element/iframe.rs
@@ -33,7 +33,11 @@ pub fn render_iframe(
         "url" => url,
     );
 
-    ctx.html().iframe().attr("src", &[url]).attr_map(attributes);
+    ctx.html()
+        .iframe()
+        .attr("src", &[url])
+        .attr("crossorigin", &[])
+        .attr_map(attributes);
 }
 
 pub fn render_html(log: &Logger, ctx: &mut HtmlContext, contents: &str) {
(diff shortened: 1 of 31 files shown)


ftml/src/render/html/element/iframe.rs
```diff
@@ -33,7 +33,11 @@ pub fn render_iframe(
         "url" => url,
     );
 
-    ctx.html().iframe().attr("src", &[url]).attr_map(attributes);
+    ctx.html()
+        .iframe()
+        .attr("src", &[url])
+        .attr("crossorigin", &[])
```
Monkatraz, line 39: Out of curiosity, does Rust not have default parameter values? Or does it handle that with overloading? In JS I would make the second argument optional so you could just do `attr("crossorigin")`. Not a suggestion, just curious.

Monkatraz (approved):

ftml/src/render/html/element/iframe.rs
emmiegit: No, it does not. Rust is very big on making things explicit, and default parameter values aren't very apparent when looking at usages of a function or method. If you want to have a default value then you create a new method with a clear name. For instance, `Vec::new()` vs `Vec::with_capacity()`.

Similarly, it does not have overloading. (Honestly I think overloading as a general concept is a mistake, like in C++ it requires name mangling and type inferences to work, which massively overcomplicates things for relatively little gain).

I suppose I could add `attr_null()` or something for attributes which only have keys, but honestly I think this is clear enough.

Monkatraz: That makes sense. I suspect that `&[]` would get compiled out anyways.

Fun fact: Typescript _adds overloading_. JS doesn't have it at all, and the way TS overloading works is purely on the type level. What's funny is that I generally don't like it there too. Usually it just adds some confusion to the function signature because you don't immediately know how you're supposed to input arguments.

If JS was redesigned, it would probably have named arguments by default.

rossjrw (approved):

emmiegit:
thanks @Monkatraz @rossjrw

PR #249, opened 2021-05-02
[WJ-452] Add cross-origin headers

emmiegit:
Complement to #248 

This adds these headers:
* [`Cross-Origin-Opener-Policy`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Opener-Policy)
* [`Cross-Origin-Embedder-Policy`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Embedder-Policy)


Monkatraz (approved):
Approved but we'll have to see what this breaks  lol

rossjrw (approved):

emmiegit:
Confirmed locally using `curl -I` that the headers are being outputted on the main page

emmiegit:
thanks @Monkatraz @rossjrw

PR #250, opened 2021-05-03
[WJ-476] Stub out stub components

rossjrw:
[WJ-476](https://scuttle.atlassian.net/browse/WJ-476)

- [x] Add a schema for stub component specifications
- [x] Add a component (tabview) with a sample stub specification
- [ ] Add validation for HTML against a stub component specification
- [ ] Update tests for Svelte components that consume stub components
- [ ] Document stubs dir
- [ ] Move documentation where approrpriate from component dir

Questions that need answering, not necessarily in this PR:

- How will FTML know which spec to use where?
- How will FTML test its output against the spec?
- How will Laravel know which spec to use where?
- How will Laravel test its output against the spec?
- How will a given Svelte component declare which spec it wants?
- How will the runtime Svelte bootstrap know which query to use for which component, without bundling the stub specs?
- Will the validation script need to be reimplemented in FTML and Laravel, or can a suite of tests for them be made using the JavaScript validation script?
- Is the validation script *really* necessary?


emmiegit:
@rossjrw Given that @Monkatraz is working on [WJ-838](https://scuttle.atlassian.net/browse/WJ-838), which covers this work, should this PR be closed?

rossjrw:
@ammongit that seems to be the case, thanks for the heads-up!

PR #252, opened 2021-05-05
[WJ-462] Add FFI #defines to ftml.h

emmiegit:
For my main WJ-462 branch, so that default Rust builds have these defines.

These are used by PHP FFI to load the symbols appropriately, see [`FFI::load`](https://www.php.net/manual/en/ffi.load.php) and [`FFI::scope`](https://www.php.net/manual/en/ffi.scope.php).

### DIFF
--- a/ftml/build.rs
+++ b/ftml/build.rs
@@ -4,6 +4,11 @@ use cbindgen::Language;
 use std::env;
 use std::path::PathBuf;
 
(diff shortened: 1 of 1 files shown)


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #253, opened 2021-05-05
[WJ-462] Implement PHP integration for ftml FFI

emmiegit:
The class for handling parsing/rendering in Wikidot is `WikiTransformation` (imo a misleading name). To get us closer to having separate, configurable Text\_Wiki vs ftml parser/renderer backends, I have moved it to a new module at `web/Wikitext`. There is a subdirectory `web/Wikitext/FTML` for all ftml-related classes and implementations.

Other than the above move, this PR does *not* change the native Wikidot parser/renderer in any way. It simply introduces the necessary code to render using ftml, which is unused.

It also fixes an issue with the Rust side where the meta list wasn't freed in the `ftml_destroy_html_output()` function. The PhpStorm IDE also made some stylistic modifications.

The integration was tested locally and confirmed to work:

----

**HTML render test:**
HTML render test

**Text render test:** (in pre)
Text render test


Monkatraz (approved):
I'm not very qualified to review this, but in principle I don't see anything wrong.

well, except for the sections of the diff where it reveals just little bits of old wikidot code. true horror

web/php/Wikitext/FTML/FtmlBackend.php
```diff
@@ -0,0 +1,16 @@
+<?php
```
danieltharp: The Wikidot namespace is intended for legacy code. Any new code should go in one of the laravel-provided folders, likely the `app` folder as everything there inherits the `Wikijump` namespace.

danieltharp (changes requested):
If you move the new FTML stuff out of `web/php` and into `app` and renamespace it to `Wikijump` I think this will be fine.

web/php/Wikitext/FTML/FtmlBackend.php
emmiegit: Ah alright, I will adjust accordingly.

CODEOWNERS
```diff
@@ -5,6 +5,7 @@
 # The ftml (Foundation Text Markup Language) service provides
 # parsing and rendering of wikitext.
 ftml                                @ammongit
+web/php/Wikitext                    @ammongit
```
rossjrw:
```suggestion
web/app/Services/Wikitext           @ammongit
```

install/aws/dev/docker/php-fpm/etc/php/99-ffi.ini
```diff
@@ -0,0 +1 @@
+ffi.enable=true
```
rossjrw, line 1: What does the 99 in the filename mean?

rossjrw (approved):

install/aws/dev/docker/php-fpm/etc/php/99-ffi.ini
emmiegit: [It's a UNIX convention](https://unix.stackexchange.com/questions/134520/what-is-the-number-prefix-in-config-files-from-d-directory). We probably don't need numbering but I'm doing it for consistency.

danieltharp (approved):

emmiegit:
thanks @pxdnbluesoul @Monkatraz @rossjrw

PR #254, opened 2021-05-05
Flesh out core SCSS styling

Monkatraz:
This PR adds a fair amount of new/ported styling to the `wj-css` package. The new stuff is really just foundational work, and is a prelude to a frontend prototype.

It should be noted that I really can't test the styling very well, and I basically haven't for this PR. However, most of it is safe, ported styling from Mainframe tweaked to suit Wikijump. All of it is subject to change, but now we don't have to start from scratch.

One significant missing feature from all of this is _vertical rhythm_, which specifically describes the spacing/margins/padding between text elements like paragraphs, images, etc. Right now, due to the reset, everything that isn't directly styled is just bunched up next to each other.

The reason I haven't done that is because:
1. It can get pretty complicated
2. I don't have much to test it with - yet

My "plan", as it stands, is for anything with a "content" context, e.g. an article, forum post text box, etc. should be given a `.wikitext` class. This class would effectively scope all of the finicky styling needed for content while leaving the rest of the UI untouched, lowering the need for fiddly resets.


client/modules/wj-css/src/abstracts/_mixins.scss
```diff
@@ -19,3 +19,31 @@
...
+/// Hides content visually but not to screen readers.
+@mixin hide-visually {
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  overflow: hidden;
+  white-space: nowrap;
+  clip: rect(1px, 1px, 1px, 1px);
+  clip-path: inset(100%);
+}
```
emmiegit, line 32: General question: Right now `[[invisible]]` just does `color: transparent`. How should this interact with this styling?

client/modules/wj-css/src/components/_links.scss
```diff
@@ -0,0 +1,49 @@
+@use "sass:list";
+@use "sass:selector";
+
+a[href]:not([href*="javascript:;"]) {
```
emmiegit, line 4: lol

client/modules/wj-css/src/components/_links.scss
```diff
@@ -0,0 +1,49 @@
...
+  // prettier-ignore
+  $internal-links-list:
+    '^="/"',
+    '^=".."',
+    '^="#"',
+    '*="wikijump.com"',
+    '*="wikijump.org"',
+    '*="wikijump.test"',
+    '*="wikijump.dev"',
+    '*="localhost"';
```
emmiegit, line 20: I don't think we use `localhost` for anything.

More notably, is this won't work for any custom domains, and I don't want to have to commit and make a new `client` build to support it, it should be automatic.

Not a blocker for this PR since you have a TODO, but just a note.

client/modules/wj-css/src/components/_links.scss
```diff
@@ -0,0 +1,49 @@
...
+
+  // make our selector lists
+  $is-internal: ();
+  $not-internal: "&[href*='//']";
```
emmiegit, line 24: Would `://` be more correct?

Also technically stuff like `mailto:` exists even though it doesn't have `//` in it.

emmiegit (approved):
LGTM, though I can't comment on most of the specific CSS rules themselves

client/modules/wj-css/src/abstracts/_mixins.scss
Monkatraz: This mixin is basically like a hide nuke - it does everything it can to visually hide the element. I assume invisible text is supposed to just be invisible, but take up room and affect layout.

client/modules/wj-css/src/components/_links.scss
Monkatraz: If you run a development server and don't use any tricks you'll run it on `localhost`. All of the `client/` web dev modes currently use `localhost`.

Also, yeah, totally aware of that issue. This particular bit of the CSS is kind of a placeholder until we figure out a solution. I think the solution most likely to work is probably environment variables.

client/modules/wj-css/src/abstracts/_mixins.scss
emmiegit: Oh right, okay.

client/modules/wj-css/src/components/_links.scss
Monkatraz: I think if you do `//` at the start of a URL it means "pick a protocol for me", which is why I excluded the `:`. Although, when I test that in Firefox both `://` and `//` work.

Also, totally forgot about stuff like `mailto:` lol

client/modules/wj-css/src/components/_links.scss
Monkatraz: can't believe I had to add that. that adds at least a kilobyte to the CSS and I'm calling the police

client/modules/wj-css/src/base/_a11y.scss
```diff
@@ -0,0 +1,39 @@
...
+// Change the display on visually hidden accessible elements.
+
+[aria-hidden="false" i][hidden] {
+  display: initial;
+}
+
+[aria-hidden="false" i][hidden]:not(:focus) {
+  position: absolute;
+  clip: rect(0, 0, 0, 0);
+}
```
rossjrw, line 39: When would these situations occur?

client/modules/wj-css/src/base/_base.scss
```diff
@@ -0,0 +1,55 @@
+// Normalize html/body
+html,
+body {
+  min-width: 100%;
+  max-width: 100%;
+  min-height: 100%;
+  font-size: 1rem;
+  line-height: 1.5;
+  touch-action: manipulation;
```
rossjrw, line 9: Why `manipulation` (which according to MDN prevents double-tap-to-zoom) over `auto`?

client/modules/wj-css/src/base/_base.scss
```diff
@@ -0,0 +1,55 @@
...
+* {
+  text-decoration-skip-ink: none;
+  text-underline-offset: 0.075em;
+  text-decoration-thickness: 0.05em;
+}
+
+// Struck-out text.
+s {
+  text-decoration: line-through 0.15em;
+}
```
rossjrw, line 49: Why can't this be left to the browser? Requesting changes here

client/modules/wj-css/src/components/_scrollbar.scss
```diff
@@ -0,0 +1,35 @@
+// Avoid setting scrollbars for something that's probably a phone
+@include media(">narrow") {
+  * {
+    scrollbar-width: thin;
```
rossjrw, line 4: Comment for the file: why can't we just let browsers do scrollbars? Requesting changes

rossjrw (changes requested):
'Requesting changes' for a couple comments below, where either changes or justification would suffice

client/modules/wj-css/src/base/_base.scss
Monkatraz: Double tapping to zoom introduces a 300ms latency for _all interaction taps_. If your UI doesn't require you to zoom (because it isn't a desktop site), you can just use pinch zooming for the scant times you actually need to zoom.

client/modules/wj-css/src/base/_base.scss
Monkatraz: iirc the strikethrough text varies between browsers, and this thickens it slightly compared to default. It's a little bit easier to make out compared to the usual browser style.

Monkatraz: wait actually I keep getting mixed up by the stupid preview, you mean the bottom CSS rule there - one sec, editing comment

rossjrw: Why disable specifically skip-ink, for example? And why does making it 'easier to make out' beat consistent usage for a single user on a single browser?

client/modules/wj-css/src/components/_scrollbar.scss
Monkatraz: Which do you prefer?

image

image

The default browser scrollbars are inconsistent native implementations, and they're usually _gigantic_. Additionally, _I am_ leaving it up to the browser when I can - the `scrollbar-width: thin` property is the "official" way of doing this.

client/modules/wj-css/src/base/_a11y.scss
Monkatraz: Great question - this actually comes from `normalize.css`. If I had a guess, these selectors are for elements that can be tab focused but aren't normally visible. A good example of that is on Github, just open a new page on Github and press tab. It'll reveal a "jump to main content" button on the top left.

client/modules/wj-css/src/components/_scrollbar.scss
rossjrw: https://webaim.org/resources/contrastchecker/?fcolor=44464C&bcolor=282B34

The default implementations are consistent for a single user across whatever websites they visit. It doesn't matter if they vary between browsers.

It doesn't really matter to me if an individual site wants a special scrollbar or if Sheaf wants a special scrollbar, but this is the base theme, and it defines everything else. I'd like to defer to the defaults that the user expects as much as possible.

client/modules/wj-css/src/base/_base.scss
Monkatraz: I guess this is a difference in opinion here? I don't really care what the browser default is - honestly, they usually kind of suck. The browser default font is Times New Roman.

About skip-ink: I find that underlines generally look better and "work better" (e.g. styling things with underlined styles, like some types of highlighted text) with skip-ink off.

client/modules/wj-css/src/components/_scrollbar.scss
Monkatraz: Yes, you pointed out the color - but that wasn't really the point? We can adjust the color easily?

I'm also going to fight you on that "consistency" point - who browses a site with a single device nowadays? On Macs, Safari will often just plain _disable_ scrollbars. I think this brings them back - I'd have to test that.

I also don't think the user has much of an idea of what a "default scrollbar" is - nearly every application they will be using has a different one. Even different Windows settings apps have different scrollbars.

rossjrw: Hmmm. Alright, fair enough. I'll concede to you on that last.

rossjrw (approved):

Monkatraz:
Going to merge without additional changes for now - I have _two_ PRs following this one up, because I'm crazy.

Being serious, we're getting pretty close to the stage where I can chill with the messy, big PRs. It's roughly going to be:
* Prism Module
* CM6 FTML grammar and Sheaf's basic components
* Complex Sheaf components, such as live preview and setting up interaction bindings
* Big ol' cleanup PR where TSDoc conformity and a bunch of other stuff is done

Anyways, thanks @ammongit , @rossjrw

PR #256, opened 2021-05-06
[WJ-470] Remove Flickr integration

emmiegit:
This integration contains a significant amount of legacy code that would have to be maintained for minimal gain, as few present Wikidot uses make use of Flickr. Additionally, the integration is broken in Wikidot Mainline, so the effort to get it to work here is likely to be extensive and thus not worthwhile.


danieltharp (approved):
Looks good, and good riddance to it.

emmiegit:
thanks @pxdnbluesoul

PR #258, opened 2021-05-06
[FTML-123] Replace "locale" with "language"

emmiegit:
We're not using it as a full locale, just to know what language to target. I have renamed `locale` to `language` in ftml and across its various consumers.


Monkatraz (approved):
Looks great to me.

emmiegit:
thanks @Monkatraz

PR #259, opened 2021-05-06
[FTML-111] Rename "special" and "modifier"

emmiegit:
Originally there was just "special", the `*` in some blocks (e.g. `[[*user ...]]`), but after "modifier" was added (e.g. `[[div_]]`), it became clear these names were too generic to be useful.

The new names are "star" and "score". This PR changes all instances of these flags, both in code, tests, and documentation, to use the new terms.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #260, opened 2021-05-07
[WJ-463] Add feature flag for wikitext backend

emmiegit:
Adds a new section for feature flags, as we will doubtlessly have more in the future. This adds a new property `FEATURE_WIKITEXT_BACKEND`, which is presently unused, but exists as a precursor to an upcoming PR to allow PHP to select the wikitext backend based on this value.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #261, opened 2021-05-07
 [WJ-506] Implement generic wikitext backend interface

emmiegit:
I've created a research document describing the uses of `WikiTransformation` here: https://scuttle.atlassian.net/wiki/spaces/WD/pages/1210712065/WikiTransformation+Usages

This PR transforms those usages into calling through the `WikitextBackend` abstract class, which then delegates the actual work to one of three backends:
* `TextWikiBackend` -- sends calls to `Text_Wiki`, the original Wikidot parser/render
* `FtmlBackend` -- sends calls to libftml, our eventual replacement
* `DummyBackend` -- trivial implementation for testing

You can change which backend is used by setting `wikitext_backend` in `wikijump.ini` to one of `text_wiki`, `ftml`, or `dummy`.

This also has a lot of miscellaneous changes to refactor Wikidot code and fix issues with local deployment:
* [WJ-516](https://scuttle.atlassian.net/browse/WJ-516), which removes the social bookmarks integration
* The "smileys" parse rule, which isn't used and we don't have the assets for anyways
* The page preview module, since it has some really annoying assumptions for `WikiTransformation` that would've made `WikitextBackend` much more annoying, and we're replacing it with our new editor Sheaf anyways.

Screenshot showing the `FtmlBackend` in use as the wikitext parser/renderer. (Headers are plaintext since they aren't added yet). I've tested all three backends and they work as expected.
backend use

This implementation of `WikitextBackend` does _not_ convey parser warnings, CSS, or backlinks. These will be addressed in future PRs.


danieltharp (approved):
Sure, let's give it a shot.

Monkatraz (approved):
yikes

I'm so sorry

web/app/Services/Wikitext/HtmlMeta.php
```diff
@@ -3,42 +3,24 @@
 
 namespace Wikijump\Services\Wikitext;
 
-use \FFI;
-
 /**
- * Class HtmlMeta, representing a returned 'struct ftml_html_meta' object.
+ * Class HtmlMeta, represents one &lt;meta&gt;, to be included in a rendered HTML document.
```
rossjrw, line 7:
```suggestion
 * Class HtmlMeta, represents one <meta>, to be included in a rendered HTML document.
```

I assume PHP allows angle brackets in comments?

web/app/Services/Wikitext/HtmlOutput.php
```diff
@@ -3,27 +3,45 @@
...
+     */
     public string $html;
+
+    /**
+     * @var string The CSS output of the rendering process.
+     *
+     * This concatenates separate CSS styles specified in the tree in one string.
+     * The styles are not necessarily valid or safe, as they come from whatever the user inputted.
+     */
     public string $style;
```
rossjrw: Out of scope for this PR, but noting that I think it'd be helpful to retain Wikidot's behaviour of putting each CSS snippet into separate style elements - created an issue [WJ-570](https://scuttle.atlassian.net/browse/WJ-570)

rossjrw (approved):

web/app/Services/Wikitext/HtmlMeta.php
emmiegit: No it interprets HTML elements as elements, like bolding a section of your phpdoc. Given that having a random open `<meta>` tag is not valid, we have to escape the angle brackets.

web/app/Services/Wikitext/HtmlOutput.php
emmiegit: I guess that's fair. I originally combined them to avoid extra string buffer allocations in Rust, but that's kind of moot given all the FFI and such happening.

emmiegit:
thanks @pxdnbluesoul @rossjrw @Monkatraz

PR #263, opened 2021-05-07
Fix UTF-16 index conversion

emmiegit:
The implementation for `get_index_end()` was bugged. As it turns out, it's not necessary at all, `get_index()` works for both the start and end of a span!

To help avoid something like this again, this PR also adds a test that thoroughly checks every single codepoint subslice of both UTF-8 and UTF-16 versions of a string and ensures they match in each possible combination.


Monkatraz (approved):
image

The crab checks are hilarious.

emmiegit:
I need one character for each possible UTF-8 width, from 1 to 4 bytes. Notably emoji exist in the 4-byte range, and since this is Rust, well...

emmiegit:
Good catch finding this bug

emmiegit:
thanks @Monkatraz

PR #264, opened 2021-05-07
Add ftml-wasm test suite that runs the actual FTML tests

Monkatraz:
Basically, the test added here checks for all `.json` files in `ftml/test/`, processes them, and then actually runs each test case. This should make it very easy to check for regressions on the WASM side of things.

Additionally, this adds some very basic smoketesting/benchmarking:
image


Monkatraz:
Excluding the `include` tests, which are skipped, there is only one test failing:
```
  FAIL  ftml-wasm  "bold-italics-underline"
    Expected values to be deeply equal:  (equal)

        ··{
        ····"elements":·[
        ······{
        ········"element":·"container",
        ········"data":·{
        ··········"type":·"paragraph",
        ··········"attributes":·{},
        ··········"elements":·[
        ············{
        ··············"element":·"container",
        Expected:
        ++············"attributes":·{},
        ··············"data":·{
        ················"type":·"bold",
        ················"attributes":·{},
        ················"elements":·[
        ··················{
        ····················"element":·"text",
        ····················"data":·"Apple"
        ··················},
        ··················{
        ····················"element":·"text",
        ····················"data":·"·"
        ··················},
        ··················{
        ····················"element":·"container",
        Expected:
        ++··················"attributes":·{},
        ····················"data":·{
        ······················"type":·"italics",
        ······················"attributes":·{},
        ······················"elements":·[
        ························{
        ··························"element":·"text",
        ··························"data":·"Banana"
        ························},
        ························{
        ··························"element":·"text",
        ··························"data":·"·"
        ························},
        ························{
        ··························"element":·"container",
        ··························"data":·{
        ····························"type":·"underline",
        ····························"attributes":·{},
        ····························"elements":·[
        ······························{
        ································"element":·"text",
        ································"data":·"Cherry"
        ······························}
        ····························]
        ··························}
        ························},
        ························{
        ··························"element":·"text",
        ··························"data":·"·"
        ························},
        ························{
        ··························"element":·"text",
        ··························"data":·"Durian"
        ························}
        ······················]
        ····················}
        ··················}
        ················]
        ··············}
        ············},
        ············{
        ··············"element":·"text",
        ··············"data":·"·"
        ············},
        ············{
        ··············"element":·"text",
        ··············"data":·"Pineapple"
        ············}
        ··········]
        ········}
        ······}
        ····],
        ····"styles":·[]
        ··}
   ```
   
 This picture is probably a bit easier to read:
image

File: client/modules/ftml-wasm/tests/ftml-original-tests.ts [hunk 1]
```diff
@@ -0,0 +1,154 @@
...
+import * as lib from "../src/index"
+import type { IPageInfo, IParseWarning, ISyntaxTree } from "../vendor/ftml"
+
+/** Test categories to skip, e.g. "image" or "include". */
+const SKIP: string[] = ["include"]
+
+const PAGE_INFO: IPageInfo = {
+  alt_title: null,
+  category: null,
+  language: "en_US",
```
emmiegit: This was changed (or should be, at least) to just `en`.

client/modules/ftml-wasm/tests/ftml-original-tests.ts
```diff
@@ -0,0 +1,154 @@
...
+          const expected = await fs.readFile(pathTXT, "utf-8")
+          const text = lib.render(test.input, { mode: "text", info })
+          assert.fixture(clean(text), clean(expected))
+        }
+
+        assert.equal(result.ast, test.tree)
+
+        // for now we won't compare warnings because they get UTF16 mapped for JS
+        // this could get corrected later but for right now it's fine
+        // assert.equal(result.warnings, test.warnings)
```
emmiegit, line 88: You could copy the UTF-16 transformation routine that the WASM does. The mapper structure is exposed to make this possible in JS.

client/modules/ftml-wasm/tests/ftml-original-tests.ts
```diff
@@ -0,0 +1,154 @@
...
+
+function fileExists(path: string) {
+  return fse
+    .access(path)
+    .then(() => true)
+    .catch(() => false)
+}
+
+function clean(str: string) {
+  return str.replaceAll(/\r\n/g, "\n").trim()
```
emmiegit, line 144: f windows

Actually, does this mean `cargo test` doesn't pass on Windows? I'm pretty sure I fixed that for old ftml but apparently not this version.

client/modules/ftml-wasm/tests/ftml-original-tests.ts
```diff
@@ -0,0 +1,154 @@
...
+}
+
+function transformRanges(str: string) {
+  return str.replaceAll(
+    /"span": \[(\d+), (\d+)\],/g,
+    '"span": { "start": $1, "end": $2 },'
```
emmiegit, line 150: Actually, since the same Rust deserializer is being run, in theory this should work unmodified, right?

emmiegit:
That bold-italics-underline test is probably failing because I left an extra field in there on accident. `serde_json` will ignore extraneous fields, but here it is being checked so it fails.

emmiegit (approved):

emmiegit:
Is it possible to add these tests (or uvu tests in general) to be run as part of GitHub Actions? Doesn't have to be this PR.

client/modules/ftml-wasm/tests/ftml-original-tests.ts
Monkatraz: Yeah, you're right - I looked at an old FTML version I think.

client/modules/ftml-wasm/tests/ftml-original-tests.ts
Monkatraz: No (I think), it's running the normal WASM rendering pipeline, which means these get deserialized to objects and not arrays. I'm not sure what the "same Rust deserializer" is honestly

client/modules/ftml-wasm/tests/ftml-original-tests.ts
Monkatraz: image

this is an infernal operating system

client/modules/ftml-wasm/tests/ftml-original-tests.ts
Monkatraz: I could - but I don't have tests written for that yet in the normal ftml-wasm tests. I'd want to do those first before I mess with them.

Monkatraz:
> 
> 
> Is it possible to add these tests (or uvu tests in general) to be run as part of GitHub Actions? Doesn't have to be this PR.

Should be totally possible - the tests are compiled to run under Node. As long as Node is used I suspect any CI will be able to run the tests.

Monkatraz:
thanks @ammongit (I'm going to add the UTF16 handling when I update the `ftml-wasm` module to actually export those functions)

PR #268, opened 2021-05-08
[WJ-455] Client monorepo code coverage reporting

Monkatraz:
This PR adds Codecov support and code coverage testing in general to the client monorepo.

~~Unfortunately, the implementation is currently somewhat flawed. Due to how tests are currently built in the monorepo, NYC (the code coverage handler) can't (or refuses to) correctly identify source files that are not somehow imported into the test files. This means we're only testing the code coverage of the files that the tests are importing (recursively).~~

~~I'm not sure how to fix it (except by compiling the _entire_ monorepo to do code coverage), but this is at least a good starting point.~~

Scratch that - I'm a genius! Let's just compile the ENTIRE monorepo into a single JS file!
image

This actually has some minor performance benefits - all the tests suites get executed in one file, so they go a little faster.


client/scripts/build-tests.js
```diff
@@ -8,36 +13,67 @@ const sveltePlugin = require("esbuild-svelte")
...
-  })
-  return tests
+async function generateMegaBundleEntrypoint() {
+  const sourceFiles = await globby("modules/*/src/**/*.ts")
+  const testFiles = await globby("modules/*/tests/*.ts")
+
+  // compile the entire monorepo into a dynamic import that won't get executed
+  // we do this so that the _entire monorepo_ gets bundled, but won't break anything
+  // this makes sure that code coverage and the like works
+  let entrypoint = "if (!globalThis && 'foo' === undefined) {\n"
```
emmiegit: What is `globalThis`, and also, why `'foo' === undefined`? When I saw the generated output I figured the `undefined` was actually a templates variable for something, but I guess not.

Monkatraz: Ah yeah, I should probably explain that. `globalThis` is simple - it's just the modern standard for the global object for Node, workers, web, etc. It's equivalent to `window`, `self`, and `global`. Yes, JS has this problem.

The reason why the expression here is strange is so that optimizers, minifiers, etc. don't dead-code this section. So I chose two things I knew would always be false but was also something that couldn't be analyzed by an optimizer.

emmiegit: Okay I see. Yeah please leave a comment explaining that. I also think a variable like `foo` makes it seem like a temporary / debug measure, maybe if it's `'some-string-value'` or `'this-will-always-be-false'` or something it's more clear that it's intended to be there.

emmiegit (approved):

Monkatraz:
Alright, well, this PR got a little messier, but it's a good thing! I was working on cleaning the monorepo a little bit - and I decided to try and simplify the test building process a little. Well, that turned into compiling the tests with Vite, replacing JSDOM with a lighter-weight and faster alternative, and removing component doc-gen for now.

The biggest change is how the test files are actually instrumented - they're instrumented as they're being built by Vite, with a plugin that I cobbled together with the power of Google. It took way too long. Anyways, the code coverage is _extremely_ accurate now, to the point that I'm not even worried about it.

Additionally, the code coverage percentage has spiked because:
1. every single file in the monorepo gets imported (top level executed), so that effectively gets tested (this is a good thing, modules should be harmless if just imported)
2. Code coverage works for Svelte components now, even in the markup

Overall, this went better than expected.

client/.gitattributes
```diff
@@ -0,0 +1,3 @@
+npm-shrinkwrap.json    merge=binary
+package-lock.json      merge=binary
+*.json                 linguist-language=JSON-with-Comments
```
emmiegit: Nitpick: no endling newline

emmiegit (approved):

client/.gitattributes
Monkatraz: whoops

client/scripts/build-tests.js
```diff
@@ -1,49 +1,167 @@
 const path = require("path")
-const esbuild = require("esbuild")
+const fs = require("fs-extra")
+
+// make sure we're at root
+process.chdir(path.resolve(__dirname, "../"))
```
rossjrw, line 5: Interesting - does this still work on Windows?

client/scripts/build-tests.js
```diff
@@ -1,49 +1,167 @@
...
+          preferConst: true,
+          sourcemapExcludeSources: true
+        },
+        external: [
+          /uvu/,
+          // /vendor\/prism\.js/,
+          // codemirror has CJS exports now, so this is ok
+          /^@codemirror/,
+          "lezer-tree",
+          // breaks things if included
```
rossjrw, line 110: Will this list of external dependencies need to be updated each time another is added?

rossjrw (approved):
LGTM

> every single file in the monorepo gets imported (top level executed), so that effectively gets tested

I'm not sure what this means - what's being tested here?

client/scripts/build-tests.js
Monkatraz: Node's `path` library is sane and hides all the Windows shenanigans. Specifically, `path.resolve` tends to convert between `/` and `\\` paths on the fly as needed.

Monkatraz: Also I'm on windows so of course it works on windows lmao

client/scripts/build-tests.js
Monkatraz: I suppose? It's not too often - only `threads` actually broke something when imported. The rest of them are external'd because it speeds up the instrumentation step a lot.

rossjrw: I see - I'd assumed that they'd been externalised because the test runner thought that they were things that the tests should cover and the coverage was inaccurate as a result. What determines if a dependency needs to be here?

Monkatraz: No actually, Istanbul/`nyc` is pretty smart and only reports coverage for things that you've narrowed down as needing covered or directly marked.

What determines what goes in here is:
1. it breaks something if you bundle it (node built-ins, `threads`)
2. it's huge and works fine if it's not bundled

Monkatraz:
> I'm not sure what this means - what's being tested here?

Oops, never answered this question.

Basically, the code coverage reporter is checking what lines of code get ran - and in this instance, it does that by literally shoving thousands of functions behind nearly every line of code and tracking which ones get called. The way the bundling works here is that we're effectively blanket importing in a single file _the entire monorepo_, and so every single module has it's top level scope executed and any side effects it has applied.

That increases the code coverage, because every single module is getting imported, and at least those top level lines are being ran, and thus counted as covered. This is a good thing in my opinion - it tests whether or not randomly importing the module breaks something.

EDIT: Here, have an example:
```ts
  static variableForString(variable) {
    cov_1pfhrxmqz3().f[8]++;
    cov_1pfhrxmqz3().s[44]++;
    if (isString(variable)) {
      cov_1pfhrxmqz3().b[18][0]++;
      cov_1pfhrxmqz3().s[45]++;
      return variable;
    } else {
      cov_1pfhrxmqz3().b[18][1]++;
    }
    cov_1pfhrxmqz3().s[46]++;
    if (isRegExp(variable)) {
      cov_1pfhrxmqz3().b[19][0]++;
      cov_1pfhrxmqz3().s[47]++;
      return variable.source;
    } else {
      cov_1pfhrxmqz3().b[19][1]++;
    }
    cov_1pfhrxmqz3().s[48]++;
    return null;
  }
```

Monkatraz:
I also want to point out that I tried _many times_ to get `c8`, a native, V8 runtime based coverage tracker, to work here. However, it just refused to cooperate - probably because of how weird this situation is (coverage testers weren't really meant for bizarre untranspiled code). If it did work, we wouldn't have needed to use Babel or anything - and the tests would be just as fast as they were before. Alas, it was not to be.

Monkatraz:
thanks @ammongit , @rossjrw

PR #272, opened 2021-05-10
Testing coverage: Add tests for almost of all of wj-utils

Monkatraz:
The other few functions not tested I didn't want to bother with yet. This is mostly just me playing with code coverage and making sure everything is working - and some of it wasn't! So I fixed that, mostly.

### DIFF
--- a/client/scripts/build-tests.js
+++ b/client/scripts/build-tests.js
@@ -66,7 +66,8 @@ async function build() {
         compilerOptions: { cssHash: () => "svelte" },
         preprocess: [
           sveltePreprocess({
-            sass: { sourceMapEmbed: true, sourceMapContents: true, sourceMap: true }
+            sourceMap: true,
+            scss: { sourceMapEmbed: true, sourceMapContents: true, sourceMap: true }
           })
         ]
       })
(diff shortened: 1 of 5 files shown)


client/scripts/build-tests.js
```diff
@@ -88,80 +89,101 @@ async function build() {
...
+              // null map informs rollup to preserve the current sourcemap
+              return { code, map: null }
+            }
+          },
           istanbul({
             babelrc: false,
             babelHelpers: "bundled",
-            extensions: [".js", ".jsx", ".es6", ".es", ".mjs", ".ts", ".tsx", ".svelte"]
+            sourceType: "module",
+            extensions: [".ts", ".tsx", ".svelte"],
```
emmiegit, line 110: Why aren't the previous extensions needed anymore?

emmiegit (approved):

client/scripts/build-tests.js
Monkatraz: Those are the default Babel extensions, and I figured we'd probably only want to do code coverage for TS since there isn't really a reason to write the client code in anything else. It also helps to avoid expensively instrumenting vendor code and the like.

PR #273, opened 2021-05-10
Update recommended ftml wasm build command

emmiegit:
Since `wasm-pack --dev` builds pretty fast without all the optimizations, we can just use that instead of `cargo check --target wasm32-unknown-unknown`, which doesn't produce `ftml.js` or anything.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #274, opened 2021-05-10
Remove the get_span() method from ftml wasm 

emmiegit:
This method previously existed instead of `get_index_end()` to allow retrieving the start and end of a span in a single function.

However now that `get_index_end()` doesn't exist anymore, and that start and end indices have no special differentiation, we should remove this. Users can just call `get_span()` twice.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #275, opened 2021-05-10
Empty warnings optimization for UTF-16 transformation

emmiegit:
We have the UTF-16 index transformation structure, which exists to convert UTF-8 indices to UTF-16 ones for Javascript. However, if there are no warnings, then we can optimize a bit by not building the UTF-16 mapping at all, since it's unused.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #276, opened 2021-05-10
[WJ-444] Users Refactor

danieltharp:
This PR is a lot.

- The OzoneUser class family, database table, and methods have been removed from the codebase and replaced with a new User class, migration, seeder, and methods.
- The UserSettings class family has been replaced with a general Settings class that can accept any type of setting (User, Forum, Site etc.)
- The creation of the Wikijump database and data is now handled almost entirely within Laravel, with the postgres seed.sql script only creating the database, user, and grant, and everything else is handled via migrations and seeders.
- `ajax-module-connector.php` has been removed in favor of an `ajax--handler` endpoint handled by Laravel that passes it off to the typical flow controller, but within the context of a Laravel application.
- Laravel Fortify has been stubbed out to assist with typical user authn as well as advanced features like MFA.
- Laravel Socialite has been stubbed out to handle OAuth logins from providers like Google and Facebook.
- The removal of some classes we intend to not support, petitions and simpletodo.
- Removal of the unused Wikijump Facade classes with the exception of...the Exception facade which _is_ being used.
- Other minor occasional fixes to l10n that annoyed me enough to do in the middle of all this.


emmiegit:
Petitions removal issue: https://scuttle.atlassian.net/browse/WJ-217

danieltharp:
> 
> 
> Petitions removal issue: https://scuttle.atlassian.net/browse/WJ-217

Thanks. I took a quick look through the code and a few bits and pieces remain, I'll close them out in another branch once this is merged.

web/.env.example
```diff
@@ -44,3 +44,17 @@ PUSHER_APP_CLUSTER=mt1
 
 MIX_PUSHER_APP_KEY="${PUSHER_APP_KEY}"
 MIX_PUSHER_APP_CLUSTER="${PUSHER_APP_CLUSTER}"
+
+# Initial password for the Admin user
+ADMIN_INITIAL_PASSWORD=admin1
+
+# This var should be set to a valid locale
+DEFAULT_LANGUAGE=en
```
danieltharp, line 52: I haven't decided how much should be an environment variable and how much should be stored in the new config array. They ostensibly do the same thing so we can discuss what pattern we prefer.

install/local/dev/php-fpm/Dockerfile
```diff
@@ -137,4 +137,4 @@ RUN sed -i "s/BASEDOMAIN/${MAIN_DOMAIN}/g" conf/wikijump.ini && \
 
 COPY etc/php/ /usr/local/etc/php/conf.d/
 
-CMD ["php-fpm"]
+ENTRYPOINT ["sh", "-c", "php artisan migrate && php-fpm"]
```
danieltharp, line 140: `php artisan migrate` handles the database migration and seeding. If the migrations have already been run, nothing happens, so new containers won't inadvertently destroy or reset data.

web/app/Actions/Fortify/CreateNewUser.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace App\Actions\Fortify;
```
danieltharp: Noting that everything in `web/app/Actions/Fortify` was added by Laravel and isn't my code.

web/app/Actions/Fortify/CreateNewUser.php
```diff
@@ -0,0 +1,41 @@
...
+
+use App\Models\User;
```
danieltharp: Commit coming to use the correct namespaces everywhere in the Fortify actions.

web/app/Http/Controllers/SettingsController.php
```diff
@@ -0,0 +1,85 @@
+<?php
+
+namespace Wikijump\Http\Controllers;
+
+use Illuminate\Http\Request;
+use Wikijump\Models\Settings;
+
+class SettingsController extends Controller
```
danieltharp, line 8: Controllers are stubs that will be fleshed out alongside the API.

web/app/Providers/RouteServiceProvider.php
```diff
@@ -60,4 +60,32 @@ protected function configureRateLimiting()
             return Limit::perMinute(60)->by(optional($request->user())->id ?: $request->ip());
         });
     }
+
+    /**
+     * Used to generate routes that aren't already cached.
+     *
+     * @return void
+     */
+    public function map() : void
```
danieltharp, line 69: We add this code as we need to define some routes that aren't normally associated with a `routes/*.php` file for Socialite callbacks from the providers.

web/config/fortify.php
```diff
@@ -0,0 +1,145 @@
+<?php
+
+use Wikijump\Providers\RouteServiceProvider;
+use Laravel\Fortify\Features;
+
+return [
```
danieltharp, line 6: Laravel boilerplate.

web/config/wikijump.php
```diff
@@ -0,0 +1,115 @@
+<?php
+
+return [
```
danieltharp, line 3: Soon I'm going to want to move the entirety of the wikijump.ini values to config file values or env values, whichever makes more sense for each one. This will save a hit to disk for every php-fpm call.

web/database/migrations/2021_05_05_185149_add_foreign_key_constraints.php
```diff
@@ -0,0 +1,292 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class AddForeignKeyConstraints extends Migration
```
danieltharp, line 7: This is broken out into its own migration because otherwise the Legacy Seeder won't fire.

web/database/seeders/UserSeeder.php
```diff
@@ -0,0 +1,65 @@
...
+                'username' => 'Anonymous',
+                'unix_name' => WDStringUtils::toUnixName('Anonymous'),
+                'password' => Hash::make(bin2hex(random_bytes(32))),
+                'email' => 'anonymous@wikijump',
+                'email_verified_at' => now(),
+                'language' => env('DEFAULT_LANGUAGE'),
+            ]
+        ]);
+        if(env('APP_ENV') != 'production') {
+            DB::table('users')->insert(
```
danieltharp, line 52: I'm adding a non-admin user to the dev deployments, accessible via `user:user1`. As we begin adding RBAC it'll be a nice thing to have for testing.

web/lib/ozoneframework/php/core/Database/BaseDBObject.php
```diff
@@ -144,10 +144,7 @@ public function populate($row) {
 		foreach ($this->fieldNames as $field) {
 			$fieldType = $peer->getFieldType($field);
 
-			$val = $row[$this->tableName."___".$field];
-			if ($val == null) {
-				$val = $row[$field];
-			}
+			$val = $row[$this->tableName."___".$field] ?? $row[$field];
```
danieltharp, line 147: This broke as part of this branch as we're no longer obeying ozone's schema patterns.

web/lib/ozoneframework/php/core/RunData.php
```diff
@@ -546,6 +546,7 @@ public function handleSessionEnd() {
 			// if session storage is empty and userId = null - clear stop the session!
 			$session = $this->session;
 			$serializedData = $session->getSerializedData();
+			if($serializedData === false) { $serializedData = []; }
```
danieltharp, line 549: This was needed as we start really crossing the streams between Laravel users and Ozone sessions.

web/lib/ozoneframework/php/core/SecurityManager.php
```diff
@@ -323,32 +316,6 @@ public function getGroupAncestors($group){
 		return $out;
 	}
 
-	public function getGroupMembers($group){
```
danieltharp, line 326: I didn't feel up to rewriting this in Eloquent since it was not used.

web/php/Actions/CreateAccountStep1Action.php
```diff
@@ -235,86 +231,43 @@ public function finalizeEvent($runData)
...
+        $nuser->username = $name;
+        $nuser->email = $email;
+        $nuser->password = Hash::make($password);
+        $nuser->unix_name = $unixified;
+        $nuser->language = $lang;
         $nuser->save();
 
         // profile
 
         $profile = new Profile();
```
danieltharp: Removing this.

web/php/Actions/LoginAction.php
```diff
@@ -32,27 +34,33 @@ public function loginEvent($runData)
...
+        if ($runData->sessionGet("login_seed") == null) {
             throw new ProcessException(_("Your session has expired. Please log in again."), "no_seed");
         }
 
-        if ($userId && is_numeric($userId) && $userId >0) {
-            $user = OzoneUserPeer::instance()->selectByPrimaryKey($userId);
-            if (!$user or password_verify($upass, $user->getPassword())) {
+        if (
+            $uname != null
+            && strtolower($uname) != 'automatic'
```
danieltharp, line 43: We won't attempt to process logins for the service accounts, this does the same thing as the previous version but is more clear in its intention.

web/php/Modules/ManageSite/ManageSitePrivateSettingsModule.php
```diff
@@ -19,15 +19,7 @@ public function build($runData)
         $runData->contextAdd("settings", $site->getSettings());
         $runData->contextAdd("superSettings", $site->getSuperSettings());
 
-        // get the viewers
```
danieltharp, line 22: This was another one I didn't feel like trying to make work in Eloquent as we'll have this replaced with RBAC before 1.0.

web/php/Utils/GlobalProperties.php
```diff
@@ -4,8 +4,6 @@
 
 use Exception;
 
-class GlobalPropertiesException extends Exception {}
```
danieltharp, line 7: Only one class should be defined in a file, and this one didn't actually do anything different from its base class.

web/phpunit.xml
```diff
@@ -6,10 +6,10 @@
 >
     <testsuites>
         <testsuite name="Unit">
-            <directory suffix="Test.php">./tests/Unit</directory>
+            <directory suffix="Test.php">tests/php/Unit</directory>
```
danieltharp, line 9: PHPUnit was broken before this.

web/tests/php/CreatesApplication.php
```diff
@@ -13,7 +13,7 @@ trait CreatesApplication
      */
     public function createApplication()
     {
-        $app = require __DIR__.'/../bootstrap/app.php';
+        $app = require __DIR__.'/../../bootstrap/app.php';
```
danieltharp, line 16: Needed for the `Tests\TestCase` class to function, which allows for Laravel's custom assertions and application booting

.

web/web/ozone.php
```diff
@@ -4,8 +4,6 @@
 use Ozone\Framework\OzoneLogger;
 use Wikidot\Utils\WikiFlowController;
 
-require ('../php/setup.php');
```
danieltharp, line 7: No longer needed here as it's provided by the autoloader, and the inclusion of this breaks Laravel booting.

install/local/dev/php-fpm/etc/php/99-xdiff.ini
```diff
@@ -1 +0,0 @@
-extension=xdiff
```
emmiegit, line 1: Is this unneeded or is it stored somewhere else?

web/.env.example
emmiegit: The convention I've been using in ftml for language is to have a language `C` (capitalized, after the standard locale of the same name), whose language has "debug-like" strings for its l10n. This way we can avoid hardcoded English, as we are native English speakers.

Alternatively, if we want to keep everything lowercase (e.g. `en`, `zh`, `pig`), we can use `default` instead.

web/app/Actions/Fortify/CreateNewUser.php
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Validate and create a newly registered user.
+     *
+     * @param  array  $input
+     * @return \Wikijump\Models\User
+     */
+    public function create(array $input)
+    {
+        Validator::make($input, [
+            'name' => ['required', 'string', 'max:255'],
```
emmiegit, line 24: Bikeshedding: I think we've also discussed having the minimum set at 2 or 3 characters.

web/app/Actions/Fortify/CreateNewUser.php
```diff
@@ -0,0 +1,41 @@
+<?php
```
emmiegit, line 1: General comment, not for this PR specifically (it's long enough as-is):

I think we should use `declare(strict_types = 1);` for new PHP files we create.

web/app/Http/Middleware/VerifyCsrfToken.php
```diff
@@ -12,6 +12,6 @@ class VerifyCsrfToken extends Middleware
      * @var array
      */
     protected $except = [
-        //
+        '/ajax--handler'  // TODO: Add the csrf token to requests to ajax--handler.
```
emmiegit, line 15: CSRF tokens sound like a good idea. We should give the cookie a distinctive name, like `wikijump_token7`!

emmiegit:
Use positive user IDs: https://scuttle.atlassian.net/browse/WJ-425 ([Changes](https://github.com/scpwiki/wikijump/pull/276/files#diff-3730fd46563ce3b4d82820c6637e2d23de9f45eaa825a9c7ac309cfafb990e89R33-R34))

web/app/Models/User.php
```diff
@@ -30,14 +66,26 @@ class User extends Authenticatable
     protected $hidden = [
         'password',
         'remember_token',
+        'two_factor_secret',
+        'two_factor_recovery_codes'
```
emmiegit, line 70: Bikeshedding: Unless these names are declared in some parent that we can't change, I would suggest changing these to `multi_factor_*`, as MFA is a more accurate term than 2FA.

web/app/Models/User.php
```diff
@@ -30,14 +66,26 @@ class User extends Authenticatable
...
         'email_verified_at' => 'datetime',
     ];
+
+    /**
+     * Retrieve the list of default values for user settings.
+     * @return array
+     */
+    public static function defaults() : array
+    {
+        return Config::get('wikijump.defaults.user');
```
emmiegit, line 88: I'm assuming `Config::get()` caches this?

web/app/Traits/HasSettings.php
```diff
@@ -0,0 +1,43 @@
...
+     */
+    public function set(array $settings) : bool
+    {
+        return $this->settings()->firstOrNew()->modify($settings);
+    }
+
+    /**
+     * Get a single setting back, or the default if unset.
+     * Note: If you want *all* the settings, use something like
+     * $user->settings()->firstOrNew() instead.
```
emmiegit, line 35: Isn't this what this method does?

web/config/fortify.php
```diff
@@ -0,0 +1,145 @@
...
+    | Some of the Fortify features are optional. You may disable the features
+    | by removing them from this array. You're free to only remove some of
+    | these features or you can even remove all of these if you need to.
+    |
+    */
+
+    'features' => [
+        Features::registration(),
+        Features::resetPasswords(),
+        // Features::emailVerification(),
```
emmiegit:
```suggestion
        // Features::emailVerification(), TODO: https://scuttle.atlassian.net/browse/WJ-511
```

web/database/factories/SettingsFactory.php
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * Define the model's default state.
+     *
+     * @return array
+     */
+    public function definition()
+    {
+        return [
+            //
+        ];
```
emmiegit, line 26: Why is this stubbed? Shouldn't it be derived automatically from `defaults.php` above?

web/database/factories/UserFactory.php
```diff
@@ -22,12 +24,21 @@ class UserFactory extends Factory
...
-            'name' => $this->faker->name,
-            'email' => $this->faker->unique()->safeEmail,
+            'username' => $username,
+            'unix_name' => WDStringUtils::toUnixName($username),
+            'email' => $this->faker->unique()->freeEmail,
             'email_verified_at' => now(),
-            'password' => '$2y$10$92IXUNpkjO0rOQ5byMi.Ye4oKoEa3Ro9llC/.og/at2.uheWG/igi', // password
+            'password' => Hash::make('password'),
             'remember_token' => Str::random(10),
+            'language' => env('DEFAULT_LANGUAGE', 'en'),
```
emmiegit, line 35: Same note re language as above

web/database/migrations/2021_04_22_013230_create_users_table.php
```diff
@@ -0,0 +1,55 @@
...
+            $table->text('two_factor_recovery_codes')->nullable();
+            $table->rememberToken();
+            $table->string('language',10)->nullable();
+            $table->unsignedMediumInteger('karma_points')->default(0);
+            $table->unsignedTinyInteger('karma_level')->default(0);
+            $table->string('real_name', 30)->nullable();
+            $table->string('pronouns', 30)->nullable();
+            $table->date('dob')->nullable();
+            $table->string('bio', 2000)->nullable();
+            $table->string('about_page', '80')->nullable();
```
emmiegit: Is this meant to say:
```suggestion
            $table->string('about_page', 80)->nullable();
```

web/database/migrations/2021_04_30_144146_add_legacy_tables.php
```diff
@@ -0,0 +1,739 @@
...
+           $table->boolean('flag_proxy')->default(false);
+           $table->string('reason', 200000)->nullable();
+           $table->timestamp('date_blocked')->nullable();
+        });
+
+        Schema::create('license', function (Blueprint $table) {
+           $table->id('license_id');
+           $table->string('name', 100)->nullable()->unique();
+           $table->string('description', 200000)->nullable();
+           $table->unsignedInteger('sort')->default(0);
```
emmiegit, line 280: Off-topic to this PR: The `licenses` table in Wikidot is essentially constant data. I've created [WJ-512](https://scuttle.atlassian.net/browse/WJ-512) for replacing the table with configuration instead.

web/database/migrations/2021_05_09_161419_create_settings_table.php
```diff
@@ -0,0 +1,35 @@
...
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('settings', function (Blueprint $table) {
+            $table->id();
+            $table->json('settings');
+            $table->unsignedInteger('setter_id');
+            $table->string('setter_type', 128);
```
emmiegit, line 20: Nitpick: Can `setter_type`s really get this long?

web/database/seeders/UserSeeder.php
emmiegit: Sounds good. When we have RBAC we can create a number of users named `test-xyz` that have various prefilled permissions. (Admin, guest, banned, etc)

web/database/seeders/UserSeeder.php
```diff
@@ -0,0 +1,65 @@
...
+                // ID 3
+                'username' => 'Anonymous',
+                'unix_name' => WDStringUtils::toUnixName('Anonymous'),
+                'password' => Hash::make(bin2hex(random_bytes(32))),
+                'email' => 'anonymous@wikijump',
+                'email_verified_at' => now(),
+                'language' => env('DEFAULT_LANGUAGE'),
+            ]
+        ]);
+        if(env('APP_ENV') != 'production') {
```
emmiegit, line 51: What will our official environment names be? We've been using `dev`, but then using the long form instead of `prod` seems inconsistent.

web/app/Models/User.php
```diff
@@ -2,26 +2,62 @@
...
+    use Notifiable;
+    use SoftDeletes;
+    use HasSettings;
+
+    /**
+     * These are service accounts added by the UserSeeder. They're used during
+     * the generation of various pages, threads, and so on. This is their ID.
+     * @see UserSeeder
+     */
+    public const AUTOMATIC_USER = 2;
```
emmiegit, line 33: Why isn't `ADMIN_USER = 1;` here?

web/php/Actions/AccountMembershipAction.php
```diff
@@ -282,7 +289,7 @@ public function restoreSiteEvent($runData)
         } else {
             $unixName = WDStringUtils::toUnixName($unixName);
 
-            if (!$runData->getUser()->getSuperAdmin()) {
+            if (!$runData->getUser()->id == 1) {
```
emmiegit, line 292: Shouldn't this compare against `User::ADMIN_USER`?

web/php/Actions/CreateAccountStep2Action.php
```diff
@@ -227,74 +222,31 @@ public function finalizeEvent($runData, $skipEvcode = false)
...
+
+        if ($u > 0) {
             $runData->resetSession();
             throw new ProcessException(_("Account creation failed: A user with this name (or very similar) already exists."), "user_exists");
         }
 
         // add new user!!!
 
-        $nuser = new OzoneUser();
-        /* email as the username!!! */
```
emmiegit, line 250: End of an era.

web/php/Actions/LoginAction.php
```diff
@@ -32,27 +34,33 @@ public function loginEvent($runData)
 
         // decrypt! woooohhooooo!!!!!!!!
```
emmiegit, line 35: I wish I had this level of enthusiasm for basic cryptography.

web/php/Actions/LoginAction.php
```diff
@@ -32,27 +34,33 @@ public function loginEvent($runData)
...
             throw new ProcessException(_("Your session has expired. Please log in again."), "no_seed");
         }
 
-        if ($userId && is_numeric($userId) && $userId >0) {
-            $user = OzoneUserPeer::instance()->selectByPrimaryKey($userId);
-            if (!$user or password_verify($upass, $user->getPassword())) {
+        if (
+            $uname != null
+            && strtolower($uname) != 'automatic'
+            && strtolower($uname) != 'anonymous'
```
emmiegit, line 44: Maybe these should be constants, similar to `ANONYMOUS_USER`?

web/php/Modules/Users/UserInfoWinModule.php
```diff
@@ -18,7 +19,7 @@ public function build($runData)
         $pl = $runData->getParameterList();
         $userId = $pl->getParameterValue("user_id");
 
-        $user = OzoneUserPeer::instance()->selectByPrimaryKey($userId);
+        $user = User::find($userId);
         $avatarUri = '/common--images/avatars/'.floor($userId/1000).'/'.$userId.'/a48.png';
```
emmiegit:
```suggestion
        $avatarUri = '/common--images/avatars/'.floor($userId/1000).'/'.$userId.'/a48.png'; // TODO https://scuttle.atlassian.net/browse/WJ-249
```

web/php/Screens/Feed/WatchedForumPostsFeed.php
```diff
@@ -129,10 +130,10 @@ public function build($runData)
             $content .= _('Author of the post').': '.WDRenderUtils::renderUser($post->getUserOrString()).'<br/>';
 
             $item['content'] = $content;
-            if ($post->getUserId()>0) {
+            if ($post->getUserId() != User::ANONYMOUS_USER && $post->getUserId() != User::AUTOMATIC_USER) {
```
emmiegit, line 133: Since this logic is appearing in a couple places, it may be worth it to add a helper method `isSystem()` or such to the user model.

web/php/Utils/KarmaCalculator.php
```diff
@@ -39,85 +47,60 @@ public function calculate($user)
...
+         * Calculate karma level.
+         *
+         * Note that the behavior of a switch in PHP is to execute every command
+         * after the first true condition without a `break` statement. If it
+         * were set so if $p >= $minPointsLevelX, everyone would end up with karma
+         * level 5.
+         */
+        switch($p) {
+            case ($p < $this->minPointsLevel1):
+                $user->karma_level = 0;
```
emmiegit, line 62: One thing I've done in C is add an explicit comment `/* FALLTHROUGH */` each time I intentionally omit a `break`.

web/php/Utils/WDPermissionManager.php
```diff
@@ -717,14 +639,14 @@ public function canBecomeAdmin($user)
...
     public function canBecomeMaster($user)
     {
-        if ($user->getSuperAdmin()) {
+        if ($user->id == 1) {
             return true;
         }
         $us = $user->getSettings();
         if ($us->getMaxSitesMaster()) {
             $c = new Criteria();
             ;
```
emmiegit: lol what

emmiegit:
```suggestion
```

web/routes/web.php
```diff
@@ -40,6 +41,28 @@
     return view('welcome');
 });
 
+/**
+ * Socialite route, null until I"m ready to begin work there.
```
emmiegit:
```suggestion
 * Socialite route, null until I'm ready to begin work there.
```

emmiegit (approved):
Most of my comments here are minor issues. Overall this is a really tremendous PR, with a lot of work put into it. This is really great for getting us closer to Laravel and away from Wikidot's nonsense.

install/local/dev/php-fpm/etc/php/99-xdiff.ini
danieltharp: Since xdiff is only used by the legacy `ODiff` class and the current docker build doesn't actually seem to register the extension, I'm honestly not super interested in keeping it around. We know we're going to replace it with something else.

web/app/Actions/Fortify/CreateNewUser.php
danieltharp: I'm not remotely married to the language styling, it is expected in a few places in the code but nothing that can't be modified. You may have to show me what you have in mind and we'll make it work.

web/app/Actions/Fortify/CreateNewUser.php
danieltharp: I'll add this in the inevitable cleanup branch and ensure the new code still works.

web/app/Http/Middleware/VerifyCsrfToken.php
danieltharp: So, there is a `csrf_token` that can be passed to the frontend view either in Smarty or Blade, but I might need some assistance from others in getting the front and back ends working together on validation.

web/app/Actions/Fortify/CreateNewUser.php
emmiegit: I'm not sure what you mean by "language styling". I just meant that usernames are required to be at least two or three characters, so a username like `Z` isn't valid. Presumably you can do that here by just `min:3`?

web/app/Models/User.php
danieltharp: They are expected to be as such by Fortify, and I don't believe there's an override available in our non-vendor classes.

web/app/Models/User.php
danieltharp: Not caching in the typical sense, but all the arrays under the `web/config` directory are loaded into RAM, and PHP-FPM's architecture is such that those arrays are always available. If anything, it's faster than calling memcached.

web/app/Traits/HasSettings.php
danieltharp: No, this only returns a single key-value pair from the array of settings. If we determine a need to get back all the values for a user (or forum or whatever) including the defaults, I'll write that up and add it to the trait.

web/config/fortify.php
danieltharp: I'm skipping the verification for now as we need a working email stack. Laravel offers the ability to "send" emails to a log file, so you can see how they'd come out, and also retrieve things like confirmation links. I'll commit the suggestion and it'll be part of a larger effort to enable email functionality, that's in the scope of the schema refactor as Emails are a legacy class.

emmiegit: Oh I just meant adding the `TODO` comment, with the relevant issue, not that you'd need to implement email verification in this PR.

web/database/factories/SettingsFactory.php
danieltharp: A factory is only used in mocking up instances of an object. I don't yet have a use case to build the factory as this class relies on the existence of a parent class, which could just as easily set the properties itself.

web/database/migrations/2021_05_09_161419_create_settings_table.php
danieltharp: It looks like it will be `Wikijump\Models\WhateverModel` so probably not. 64 is probably a reasonable length.

web/database/seeders/UserSeeder.php
danieltharp: ¯\_(ツ)_/¯

web/app/Models/User.php
danieltharp: My intent was originally just to handle the negative number users since they were the most relevant. I can probably do that as well, but all of the legacy code that is looking for user ID 1 will also end up being replaced with model policies fairly soon.

web/php/Actions/LoginAction.php
danieltharp: Possibly but we're also going to be handing off authn to Fortify soon and the legacy module and action will go away.

web/php/Screens/Feed/WatchedForumPostsFeed.php
danieltharp: Not a bad idea. I'll do that in the follow-up.

web/app/Models/User.php
emmiegit: Fair point, it probably isn't worth the time to replace it all.

web/app/Actions/Fortify/CreateNewUser.php
danieltharp: Sorry, for whatever reason GitHub's not letting me comment on the actual thing I was trying to respond to which was `en` for `DEFAULT_LANGUAGE`.

emmiegit: Oh alright, I see.

emmiegit: So for the language stuff, I meant that the default language has debug strings. So instead of
```php
_("This user does not exist! ;-)")
```
You have
```php
_("error-user-not-exist")
```
So any areas where we would otherwise have hardcoded English / the language parameter is missing, you'll see `error-user-not-exist` on the web page, which is clearly an error.

I want to be vigorous about avoiding hardcoded English, since Wikidot plays things pretty loose and the experience for international users is pretty inconsistent and atrocious, especially if they aren't fluent in English.

web/config/wikijump.php
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * How many username changes a user is allowed.
+     * A farm admin can force changes above this limit.
+     */
+    'username_change_limit' => 3,
+
+    /**
+     * Forbidden usernames.
+     */
+    'forbidden_usernames' => [
```
Monkatraz, line 21: Should we think about adding a basic curse word filter here too? I'm sure there is a decent library for it - I'd rather make it harder than easier for people to use the n-word for their name.

Monkatraz (commented):
Totally reviewed!!!!! I mostly just wanted to say: Holy crap dude, this is a ton of work! Awesome stuff.

web/config/wikijump.php
emmiegit: There is a separate issue for that: [WJ-260](https://scuttle.atlassian.net/browse/WJ-260).

To allow sites to add their own restriction lists (for instance, for non-English languages) it should be a feature that hits the database. We can add some general blacklists here, but the comprehensive filtering is separate.

web/app/Actions/Fortify/CreateNewUser.php
danieltharp: My only reservation with that pattern is that the locale translation file is _much_ easier for a translator to read when they have the original English meaning side-by-side with their translation. See https://laravel.com/docs/8.x/localization#using-translation-strings-as-keys

emmiegit: That's true. I guess one of my concerns is that when you update the English copy you have to update all the translation files as well, something Wikidot has seemingly forgotten to do all over its codebase.

danieltharp: I thought about this some more and I think we'll try it your way. Translators can copy the English file and be able to do the same sort of thing, and an obvious placeholder string will almost certainly be reported to us where a bit of stray English may not.

PR #277, opened 2021-05-11
Update dependencies, in particular Prettier

Monkatraz:
Also rerformats the test build file a little bit to make it cleaner. The emitted output of the test bundler is also nicer.


Monkatraz:
Some weird issues with the CI there - so I made it use `npm ci` instead of `npm install`. That means caching dependencies is useless - however, it seems to take the same amount of time between caching dependencies and not caching dependencies (because `npm ci` is very fast), so this seems fine...

rossjrw (approved):

Monkatraz:
thanks @rossjrw

PR #278, opened 2021-05-11
Add enum abstract class

emmiegit:
Based on the solution [here](https://stackoverflow.com/questions/254514/enumerations-on-php/17045081#17045081), it will allow us to make enums in PHP using class constants.

An example enumeration using this code:

```php
use \Wikijump\Common\Enum;

final class ResponseStatusCode extends Enum
{
    const OK                         = 200;
    const CREATED                    = 201;
    const ACCEPTED                   = 202;
    // ...
    const SERVICE_UNAVAILABLE        = 503;
    const GATEWAY_TIME_OUT           = 504;
    const HTTP_VERSION_NOT_SUPPORTED = 505;
}
```

### DIFF
--- a/web/app/Common/Enum.php
+++ b/web/app/Common/Enum.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace Wikijump\Common;
+
+use http\Exception\RuntimeException;
+
+/**
+ * Implements an enumeration abstract class for the creation of enums within PHP.
+ *
+ * Children should make a final class that extends Enum, and then add all the
+ * variants as constants for enum values.
+ *
+ * The static methods values() and isValue() are available for use.
+ *
+ * See https://stackoverflow.com/questions/254514/enumerations-on-php/17045081#17045081
+ *
+ * @package Wikijump\Common
+ */
+abstract class Enum
+{
+    final private function __construct()
+    {
+        throw new RuntimeException("You may not create Enum class instances");
+    }
+
+    final private function __clone()
+    {
+        throw new RuntimeException("You may not clone Enum class instances");
+    }
+
+    final public static function values(): array
+    {
+        $class = new ReflectionClass(static::class);
+        return $class->getConstants();
+    }
+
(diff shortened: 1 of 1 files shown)


Monkatraz (approved):
```php
final class ResponseStatusCode extends Enum
{

    const OK                         = 200;
    const CREATED                    = 201;
    const ACCEPTED                   = 202;
    // ...
    const SERVICE_UNAVAILABLE        = 503;
    const GATEWAY_TIME_OUT           = 504;
    const HTTP_VERSION_NOT_SUPPORTED = 505;

}
```
It's wonderful to know that PHP is dealing with the _exact same issues_ TypeScript deals with! this is a really nice and slick solution for this problem though

EDIT: Just for kicks, here is how TypeScript transforms its enums:
```ts
enum Enum {
  A,
}

let a = Enum.A;
let nameOfA = Enum[a]; // "A"
```
```js
var Enum;
(function (Enum) {
    Enum[Enum["A"] = 0] = "A";
})(Enum || (Enum = {}));
let a = Enum.A;
let nameOfA = Enum[a]; // "A"
```

It's kind of horrifying. If it's a `const` enum they'd actually get compiled out, but with `isolatedModules` (which is nigh-required nowadays due to build tools) `const` enums don't do anything.

emmiegit:
I thought Typescript did have enums though, I could've sworn I've used them before.

Monkatraz:
> I thought Typescript did have enums though, I could've sworn I've used them before.

Yeah it does, but they're fake. You often don't see them used because people don't trust them - and the fact that you can implement 90% of their functionality using the following:
```ts
const Enum = {
  Type1: 1,
  Type2: 2
} as const

function takesEnum(input: keyof Enum) {
  // blegh
}
```

Now is that verbose and not as nice as the actual `enum` syntax? Totally - but people just don't use them even though they really should...

emmiegit:
thanks @Monkatraz

PR #280, opened 2021-05-11
Experimental proposal: Use PNPM instead of Lerna/NPM

Monkatraz:
Link to PNPM: https://pnpm.io/

This switches Lerna out for using PNPM, an alternative to the NPM client. It uses the same registry and everything - it's just a different way of using NPM packages.

PNPM offers the following advantages to us:
* Native monorepo handling, which removes the Lerna dependency (this gets rid of about 400+ dependencies).
* It's fast - it uses an external cache at the root of your disk (or somewhere configurable) and can cleanly install cached modules even if `node_modules` is entirely deleted.
* It has a much tighter, smaller lock-file format which should reduce on the hilarious line counts in PRs.
* It can handle caching in CIs a bit better than NPM can, because it doesn't reuse the `node_modules` folder.
* It has some really unique and powerful ways of handling how packages are symlinked, installed, etc. that avoids a lot of the issues NPM has.
* You can just do `pnpm add <package>` in a monorepo subpackage and not have to care about bootstrapping or anything.

The advantages above are the reason why are lot of repos use PNPM nowadays, including most of the Svelte repos and pretty much the _entirety_ of Microsoft's monorepos.

But it does have issues:
* It has to replace your NPM client - not literally, but you shouldn't be using `npm run <cmd>` or `npm install` in the monorepo. You need to use `pnpm <cmd>`, `pnpm install`, `pnpm add <package>`, etc.
* It's not familiar, and it's something you have to install. (`npm install pnpm --global`)
* I don't think it's as intuitive as `npm` is.

I haven't updated the documentation for this PR - I don't really want to put the effort in for that until I know how everyone feels about this.


client/modules/cm-lang-ftml/package.json
```diff
@@ -6,7 +6,9 @@
   "keywords": [
     "wikijump"
   ],
-  "scripts": {},
+  "scripts": {
+    "build": ""
```
Monkatraz, line 10: These `"build": ""` commands are added everywhere so that `ultra-runner` can find and cache the "build results" of these packages. This enables the incremental rebuilding feature for anything that does actually need a build-step.

client/package.json
```diff
@@ -71,12 +65,12 @@
     "globby": "^11.0.3",
     "happy-dom": "^2.19.0",
     "include-media": "^1.4.10",
-    "lerna": "^4.0.0",
     "nabs": "^0.3.0",
     "nyc": "^15.1.0",
     "postcss": "^8.2.15",
     "prettier": "^2.3.0",
     "prettier-plugin-svelte": "2.3.0",
+    "rollup": "^2.47.0",
```
Monkatraz, line 73: Added `rollup` to satisfy some peer dependency complaints - the only ones left are now because we're on the TypeScript beta.

client/.npmrc
```diff
@@ -1,2 +1,7 @@
 # TODO: turn this off at some point
 legacy-peer-deps = true
+
+# PNPM
+link-workspace-packages = deep
+ignore-workspace-root-check = true
+shamefully-hoist = true # wish this wouldn't be required, but build tools won't cooperate
```
emmiegit, line 7: What does this do?

client/pnpm-workspace.yaml
```diff
@@ -0,0 +1,4 @@
+packages:
+  - 'modules/*'
+  - 'web/*'
+  - '!**/tests/**'
```
emmiegit: Missing newline at end of file

emmiegit (approved):

client/.npmrc
Monkatraz: It's hard to explain, but I'll try:

NPM (and/or Lerna) will hoist a workspaces dependencies all the way to the root. If a package depends on `xyz@1.0.0`, the root will get a package installed in `node_modules` named `xyz`. Hoisting!

But, PNPM doesn't directly do that, because that opens up the issue of _spoooky phantom dependencieeessss_ which basically means that another package which depends on `xyz` can just nab it from `node_modules` at the monorepo root, _without actually depending on it inside of the package's `package.json`_.

PNPM instead shoves all of the hoisted dependencies into `.pnpm` or something in `node_modules` and then symlinks the hard, actually listed dependencies for the root monorepo package into `node_modules`. This preserves "what you expect", but sometimes build tools don't resolve a subpackage's dependencies, for some god damn reason, and instead try to resolve them from the monorepo root. By "shamefully hoisting", PNPM won't make a strict `node_modules` and the naughty build tools will find the packages they're looking for in the root `node_modules`.

Clear as mud?

rossjrw (commented):
Noting that to get this to work with Webstorm I had to change the ESLint package to `wikijump\client\node_modules\.ignored_eslint` and the TypeScript interpreter to `wikijump\client\node_modules\.ignored_typescript` - no idea if that's the _correct_ way to do that but it is _a_ way

Overall this seems fine to me.

Monkatraz:
> Noting that to get this to work with Webstorm I had to change the ESLint package to `wikijump\client\node_modules\.ignored_eslint` and the TypeScript interpreter to `wikijump\client\node_modules\.ignored_typescript` - no idea if that's the _correct_ way to do that but it is _a_ way
> 
> Overall this seems fine to me.

okay then

Apparently WebStorm has some notorious issues with symlinks - and combined with WSL and other potential issues that might just be the best way to handle it.

Monkatraz:
Alright, I've significantly revamped the documentation of the repo to go along with this PR. I'm fairly confident about my docs here - mainly because the repo is just less complicated to explain. If there is issues with the docs, just go ahead and make a PR. I don't want to wait on big tooling changes due to the sheer blocking-factor.

thanks @ammongit , @rossjrw

PR #281, opened 2021-05-12
[WJ-444] Follow-up cleanup and improvements

danieltharp:
This PR puts a bit of polish on #276 and adds some QOL improvements.
- `declare(strict_types=1);` has been added to everything in `web/app` including the Laravel-provided boilerplate.
- A new `LegacyCompatibility` trait has been created to help keep new and old pieces of the code talking to each other. New models should implement this trait until we're no longer using classes that extend `BaseDBPeer` anywhere, which will be a little bit.
- The database migration files now include the proper information for destroying the legacy tables and foreign key constraints, so `php artisan migrate:reset` works as expected.
- The legacy migration files include the new `startingValue()` attribute to not clash with the hardcoded IDs set by the seed data.
- The `fts_entry` legacy table has been re-added as it's apparently still hooked into the page and post creation process.
- The remaining usages of Petitions are gone, closing [WJ-217](https://scuttle.atlassian.net/browse/WJ-217).
- An apparently long-lived bug with saving pages with `[[[wikilinks]]]` in them has been fixed.
- The PHPUnit configuration has been further improved to allow the tests to run in an ephemeral SQLite environment. This will allow us to run PHPUnit as a GitHub Action.
- Many small fixes to modules and actions, allowing the Site Manager and User Panel to work at nearly 100%. I may get the remaining issues fixed before merging this, but the rest of the PR is ready for review now.


web/app/Helpers/LegacyTools.php
```diff
@@ -15,19 +17,28 @@ class LegacyTools
...
     {
         $offset = strlen(dirname(__FILE__, 3)); // Get the length of the string of the absolute path 3 levels up from this file.
         $unique_path = substr($path, $offset, -4); // Chop off that length and the last 4 characters. (.php)
         $unique_path = str_replace('/', '\\', $unique_path);
-        $translations = array(
+        $translations = [
             "\\php\\" => "Wikidot\\",
             "\\lib\\ozoneframework\\php\\core\\" => "Ozone\\Framework\\",
             "\\lib\\ozoneframework\\php\\Template\\" => "Ozone\\Framework\\Template\\"
-        );
```
emmiegit, line 27: General question: There isn't any difference between `array(...)` and `[...]`, right?

web/app/Models/Settings.php
```diff
@@ -89,9 +96,10 @@ public function modify(array $setting = [], array $options = []) : bool
             }
             /**
              * If the model *did* exist and all the settings are no defaults,
-             * destroy the object. Returns true
+             * destroy the object. Return true.
```
emmiegit: Nitpick: as a complete sentence, shouldn't this be "Returns true."?

web/php/Utils/GlobalProperties.php
```diff
@@ -74,6 +74,8 @@ class GlobalProperties
     public static $UI_SLEEP;
     public static $DEFAULT_LANGUAGE;
 
+    public static $FEATURE_WIKITEXT_BACKEND;
```
emmiegit, line 77: I should've just made a small PR to fix this, I guess I had more confidence I'd be finished with my main PR than I should've :sweat_smile:

web/php/Utils/WDPermissionManager.php
```diff
@@ -134,7 +134,7 @@ public function __construct()
 
     public function hasPermission($action, $user, $site = null)
     {
-         if ($user->id == 1) {
+         if ($user->id === 1) {
```
emmiegit, line 137: Oof, do I want to know the story behind this one?

emmiegit (approved):
:broom:

web/app/Helpers/LegacyTools.php
danieltharp: `array()` is deprecated in favor of `[]` by most coding standards, and PHPStorm considers it a warn-level inspection problem.

web/php/Utils/GlobalProperties.php
danieltharp: No worries, I knew you had done it somewhere but it was causing the php-fpm container to crash-loop.

web/php/Utils/WDPermissionManager.php
danieltharp: No story, just favoring stricter equivalence when I'm touching old code.

web/app/Helpers/LegacyTools.php
danieltharp: To be absolutely clear though, there is no functional difference and I'd be surprised if there ever was, given the ubiquity of both.

PR #290, opened 2021-05-15
[WJ-456] [WJ-459] Add basic FTML live-preview output

Monkatraz:
We move ever closer to having a working editor.

This PR is fairly rough - I didn't want to include _too much_ in it, but it was getting kinda big quickly. Ultimately, all this does is this:
image

The preview on the right updates automatically.

@ammongit if you could create an FTML issue for sourcemapping of some type (there is multiple ways you could do it), that'd be great.


client/modules/wj-css/src/abstracts/_media.scss
```diff
@@ -1,5 +1,5 @@
 // Import `include-media` package from NPM
-@import "include-media";
+@import "../../node_modules/include-media/dist/include-media";
```
Monkatraz, line 5: I had really weird resolution issues when trying to import `_abstracts.scss` inside of a Svelte component, so for now all of this pathing is just done manually, unfortunately.

client/modules/components/src/Card.svelte
```diff
@@ -0,0 +1,50 @@
+<script lang="ts">
```
emmiegit, line 1: What is a "card" here?

client/modules/wj-css/src/abstracts/_media.scss
emmiegit: I would port this comment to the SCSS file itself, so future readers of the code understand why this is path is direct

emmiegit (approved):

client/modules/components/src/Card.svelte
Monkatraz: "Card" is a common name for a UI panel of some kind. They're very generic.

https://open-ui.org/components/card.research/

Monkatraz:
thanks @ammongit

PR #292, opened 2021-05-16
[WJ-505] Remove xsend fields from wikijump.ini

emmiegit:
We no longer use lighttpd so these headers don't do anything, and they're disabled anyways.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #293, opened 2021-05-16
[WJ-564] Increase ftml code coverage

emmiegit:
Since installing code coverage, we can look at the results and find areas where coverage is lacking, and add tests for them.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #294, opened 2021-05-16
[WJ-459] [WJ-456] [WJ-491] Significant UI expansion, particularly Sheaf

Monkatraz:
this is a chunky boi (SORRY)

There is a lot in this PR, but it's 90% UI related. Lots of component helper functions, fundamental components, etc.

Main bits:
* Component helper functions, such as `focusGroup`, `anim`, `Media` singleton, and a few others
* Major UI expansion for Sheaf, with tabbed view and optimized preview rendering scheme
* Tabview and tabs, although these are Mainframe ports. These will probably be expanded upon so that they can be used for FTML in the future
* Repacked FTML and updated the interfaces, API, etc.
* Buncha fundamental components like Button, Icon, and Spinny (a loading spinner)

The Icon component uses Iconify, which may be a bit presumptuous of me but the utter simplicity of its usage I think will convince everyone.


Monkatraz:
Okay, because I was annoyed, I made this change to Codecov:
```yaml
coverage:
  status:
    patch: false
    project:
      default:
        informational: true
```
What this does is make it so that Codecov basically does nothing but report coverage. It won't fail CI, it won't put annotations in GitHub diffs. I think this early in development this is 100% fine, but let me know if it isn't.

emmiegit:
@Monkatraz 	

> Okay, because I was annoyed, I made this change to Codecov:
> 
> ```yaml
> coverage:
>   status:
>     patch: false
>     project:
>       default:
>         informational: true
> ```
> 
> What this does is make it so that Codecov basically does nothing but report coverage. It won't fail CI, it won't put annotations in GitHub diffs. I think this early in development this is 100% fine, but let me know if it isn't.

At least for me, I'm definitely going to forget that we added that as a temporary thing. Could you add a comment explaining why it was added and calling for it to be eventually removed?

Monkatraz:
> At least for me, I'm definitely going to forget that we added that as a temporary thing. Could you add a comment explaining why it was added and calling for it to be eventually removed?

Donezo:
```yaml
coverage:
  # TODO: Remove this when the project stabilizes
  # This section sets it so that Codecov does basically nothing but report coverage.
  # It won't fail CI, and it won't put annotations in diffs.
  # This is probably "bad" for the long run, but for now it's set like this because
  # test coverage may be inconsistent across the project for a long time.
  # Early in development this is fine, but this should get removed in the future.
  status:
    patch: false
    project:
      default:
        informational: true
```

client/modules/wj-util/src/decorators.ts
```diff
@@ -0,0 +1,39 @@
...
+ * ```
+ * let perf = 0
+ * class foo {
+ *   ;@measure(time => perf = time)
+ *   method() { return "some-expensive-calculation" }
+ * }
+ * ```
+ * @param callback - Callback to fire with the performance measurement taken.
+ */
+export function measure(callback: (perf: number) => void) {
```
Monkatraz, line 15: This decorator is unused currently (because you can only really use decorators on classes and class methods) but it serves as an example of how to make one, so I felt it was worth keeping around.

client/scripts/tests-shim.js
```diff
@@ -1,3 +1,12 @@
 const { GlobalRegistrator } = require("@happy-dom/global-registrator")
 
 GlobalRegistrator.register()
+
+// this is a filthy hack
```
Monkatraz, line 5: i'd like to point out that it hurt my soul to write this hack

it works though lmao. `happy-dom` apparently doesn't have `matchMedia` implemented, probably because that would require implementing a ton of the CSS spec

client/modules/components/src/Spinny.svelte
```diff
@@ -0,0 +1,216 @@
+<!--
+  @component Versatile inline or overlay based spinner.
+-->
```
rossjrw, line 3: imagine thinking we need this when we have progressbar

client/modules/components/src/Spinny.svelte
```diff
@@ -0,0 +1,216 @@
...
+  /** Sets the styling so that the spinner can be displayed inline along with text. */
+  export let inline = false
+  /** CSS `top` offset. */
+  export let top = inline ? "0" : "50%"
+  /** CSS `left` offset. */
+  export let left = inline ? "0" : "50%"
+  /** Sets the width of non-inline spinners, otherwise sets font-size. */
+  export let size = inline ? "1em" : "120px"
+  /** Sets a delay until the spinner appears. Only works for non-inline spinners. */
+  export let wait = 300
```
rossjrw, line 20: Why would this be desirable?

client/modules/components/src/Wikitext.svelte
```diff
@@ -0,0 +1,147 @@
+<!--
+  @component Generic wikitext container.
+-->
```
rossjrw, line 3: Seems like this component is more like a wikitext preview in Sheaf et al than a 'generic wikitext container' (which to me implies e.g. the main block of content on a page). Is this the case?

client/modules/components/src/Icon.svelte
```diff
@@ -0,0 +1,52 @@
...
+  let icon: IconifyIcon | null = null
+  let viewBox = "0 0 0 0"
+
+  $: if (i) {
+    if (Iconify.iconExists(i)) {
+      icon = Iconify.getIcon(i)
+    } else {
+      Iconify.loadIcons([i], () => {
+        icon = Iconify.getIcon(i)
+      })
```
emmiegit, line 27: What does this do if the icon can't be found or loaded?

client/modules/components/src/Spinny.svelte
```diff
@@ -0,0 +1,216 @@
+<!--
```
emmiegit, line 1: Shouldn't a spinner be called "spinner"?

client/modules/components/src/Spinny.svelte
```diff
@@ -0,0 +1,216 @@
...
+  /** Sets the text, if any, to be displayed alongside the spinner. */
+  export let description = ""
+
+  let label = ""
+  // prettier-ignore
+  $: switch (status) {
+    case "active":  label = "Active loading indicator" ; break
+    case "error":   label = "Error indicator"          ; break
+    case "success": label = "Success indicator"        ; break
+    case "warning": label = "Warning indicator"        ; break
```
emmiegit, line 38: Hardcoded english. You can just add a TODO for later, but do note it.

client/modules/components/src/Wikitext.svelte
```diff
@@ -0,0 +1,147 @@
...
+
+  let rendering = false
+
+  let perfRender = 0
+
+  const render = createMutatingLock(async (wikitext: WikitextInput) => {
+    const displayIndicatorTimeout = setTimeout(() => (rendering = true), 100)
+    const measure = perfy()
+    if (typeof wikitext === "function") wikitext = wikitext()
+    wikitext = await wikitext
```
emmiegit, line 66: `wikitext = wikitext()` and then `wikitext = await wikitext`?
This is kind of weird, I don't really understand what's going on here.

client/modules/components/src/Wikitext.svelte
```diff
@@ -0,0 +1,147 @@
...
+      ? typeof wikitext !== "string"
+        ? wikitext
+        : await FTML.render(wikitext)
+      : { html: "", styles: [""] }
+    perfRender = measure()
+    clearTimeout(displayIndicatorTimeout)
+    return result
+  })
+
+  // TODO: Security audit of this - how much should we trust FTML output right now?
```
emmiegit, line 77: ftml should only produce safe HTML elements (at least as far as we accept on Wikidot, like through our attribute filter).

Unless you're referring to something else? What would this code do if it was fully untrusted?

client/modules/components/src/tabs/Tabview.svelte
```diff
@@ -0,0 +1,109 @@
...
+  export let conditional = false
+  export let noborder = false
+  export let contained = false
+  export let compact = false
+
+  let ready = false
+
+  let buttons: HTMLElement | undefined
+
+  let key = writable<any>(null)
```
emmiegit, line 29: What does this do?

client/modules/ftml-wasm-worker/src/worker/ftml.worker.ts
```diff
@@ -2,9 +2,49 @@
...
+/**
+ * Ad-hoc HTML formatting.
+ *
+ * Creates newlines between any non-obvious inline-formatting tags.
+ * Uses `indent.js`, a tiny library, to handle indentation of the resulting HTML.
+ */
+function formatHTML(html: string) {
+  try {
+    html = html
+      .replaceAll(/<\/?([^\s<>]+)([^]*?)>/g, (match, tag, extra) =>
```
emmiegit, line 37: Regex to match HTML?

client/modules/sheaf/src/render-handler.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  /**
+   * Renders the document's combined stylesheet.
+   * @decorator `@Memoize`
+   */
+  @Memoize()
+  async style() {
+    const { styles } = await this.result()
+    return styles.join("\n")
```
emmiegit: As [Croque notes here](https://github.com/scpwiki/wikijump/pull/261#discussion_r633123357), stylesheets should be separate because some constructs (e.g. `@import`s) need to be the first thing in the style.

client/modules/wj-util/src/decorators.ts
```diff
@@ -0,0 +1,39 @@
...
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  return (_target: Object, _propertyKey: string, descriptor: PropertyDescriptor) => {
+    const method = descriptor.value
+    const async = method.constructor.name === "AsyncFunction"
+
+    if (!async) {
```
emmiegit: If there is both an if and else branch, I find having the main condition be positive (i.e. not negated) to be more readable.

client/modules/wj-util/src/index.ts
```diff
@@ -357,3 +401,15 @@ export function html(strings: TemplateStringsArray, ...subs: (string | string[])
   })
   return toFragment(src)
 }
+
+/**
+ * Performs a modulo operation.
+ * This differs from JavaScript's `%` operator,
+ * which is more of a remainder operator.
```
emmiegit, line 408: I assume this is with respect to negative values?

emmiegit (approved):

client/modules/components/src/Icon.svelte
Monkatraz: Fairly certain it just does nothing - I suppose I could add a placeholder icon.

client/modules/components/src/Spinny.svelte
Monkatraz: Spinny is a very fun name. It also differentiates this component from any other generic spinner, as many things can be called a spinner.

client/modules/components/src/Wikitext.svelte
Monkatraz: Yeah, it's a bit JS wizardary heavy. Basically, it's very quickly "decomposing" the wikitext variable into its final form, as it may or may not be a Promise, and it may or may not be a function, and that function may return a promise, etc. One thing to note about JS `await` is that you can await anything and it's fine. If it's not a promise or thenable it just resolves to the value.

client/modules/components/src/Wikitext.svelte
Monkatraz: Well, in the early stages if Wikijump gets used anywhere it may be a good idea to not trust FTML until we get more time with it.

If the code was fully untrusted, I'd probably hook up DOMPurify to it, along with a bunch of rules/hooks so that DOMPurify doesn't sanitize out some normally blacklisted syntax.

client/modules/components/src/tabs/Tabview.svelte
Monkatraz: Makes the key into an observable. Observables can be subscribed to and updated, which informs all subscribers. This key is listened to to figure out which tab is selected.

client/modules/ftml-wasm-worker/src/worker/ftml.worker.ts
Monkatraz: Summoning zalgo is a past time of mine.

It just matches tags so it seems to be fine lmao

client/modules/wj-util/src/index.ts
Monkatraz: Correct

client/modules/ftml-wasm-worker/src/worker/ftml.worker.ts
emmiegit: :/

client/modules/components/src/Wikitext.svelte
emmiegit: Well note that for rendered page output, PHP puts any output through an HTML purifier, so at least those can be trusted.

client/modules/components/src/Wikitext.svelte
emmiegit: I was more wondering why it would be in different forms like that to begin with.

client/modules/components/src/Wikitext.svelte
Monkatraz: Oh also I think with DOMPurify I could hook into it and have it report "violations", which may be interesting to log. I'm not sure how often it's just going to report minor nitpicks or not but I could play with it

client/modules/ftml-wasm-worker/src/worker/ftml.worker.ts
Monkatraz: If it's any consolation, if it breaks, all you get is a  weird (but probably valid) output for the HTML Output preview. Literally all it's here for is to make it not impossible to read

client/modules/components/src/Wikitext.svelte
Monkatraz: Makes it very easy to use. If you have wikitext already, you can provide it. If you have a Promise for wikitext, you can just provide it directly without shoving any async into markup or into your scripts. If it's a function, you can use that for performance reasons.

Because the component is also supposed to be highly reusable and versatile, it makes sense to have a very flexible input

client/modules/components/src/Wikitext.svelte
Monkatraz: For right now I'm going to trust FTML's output - it's locked down pretty well I think. We can maybe hedge our bets that the browser native sanitizer API will be somewhat available by the time we start getting users, and we could use that. Or just be cautious, idk.

client/modules/sheaf/src/render-handler.ts
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * Heavily memoized FTML-render-emit handler.
+ * Designed to invoke the FTML renderer as little as possible.
+ */
+export class RenderHandler {
+  private declare fragmentNode?: DocumentFragment
+
+  /** @param doc - The CodeMirror `state.doc` to render with. */
+  constructor(public doc?: Text) {}
```
rossjrw, line 14:
```suggestion
  constructor(private doc?: Text) {}
```
Changing it would do nothing, assuming the wanted method has already been called at least once - might be better to prevent potentially inconsistent behaviour? Or will mutability be important here?

rossjrw (approved):

client/modules/components/src/Spinny.svelte
Monkatraz: you joke but we'll need a progress bar gdi

client/modules/components/src/Spinny.svelte
Monkatraz: So that you don't have to add spinner delay logic yourself everywhere. Having the spinner instantly appear is strange so this helps

client/modules/components/src/Wikitext.svelte
Monkatraz: In this circumstance yes, but there may be other uses for this component which is why there is so many input options. I have no idea what those scenarios are because we haven't discovered them yet.

client/modules/sheaf/src/render-handler.ts
Monkatraz: You're correct about mutability but rather than making it private it should be made `readonly`. I'll add that in another PR because I'd rather not deal with merge conflicts

client/modules/components/src/Spinny.svelte
rossjrw: Why would that be strange?

Monkatraz: It's a big honking wavy spinner with an intro animation, and if it appeared instantly and an operation took less time than the intro took you'd just set a spinner scale up like 5% and then immediately disappear again. And you'd see it constantly if the operation occurred frequently. It's kind of like a debounce

PR #295, opened 2021-05-16
[WJ-570] Output styles as a vec

emmiegit:
This modifies `HtmlOutput` such that `styles` is now a `Vec<String>` rather than a `String`. This also modifies the WASM and FFI usages of ftml.


rossjrw (approved):

emmiegit:
thanks @rossjrw

PR #296, opened 2021-05-17
[ftml] Rename "html" to "body"

emmiegit:
This is a more accurate way of describing which part of the HTML document is being produced.

It also has a minor change that uses a more complete crate version in the generated HTML meta fields.


rossjrw (approved):

emmiegit:
thanks @rossjrw

PR #297, opened 2021-05-17
Add sample C program for testing the ftml FFI

emmiegit:
Allows simple verification that the right symbols are being exported, the header is being generated appropriately, and there aren't any memory issues on the Rust FFI side.

Just adding this to the `misc` directory because I kept finding myself recreating a simple version of this every time I needed to verify this, and having it premade would assist with development.

### DIFF
--- a/ftml/misc/ffi_test.c
+++ b/ftml/misc/ffi_test.c
@@ -22,15 +22,15 @@ int main(int argc, char **argv)
 {
 	struct ftml_html_output output;
 	struct ftml_page_info page_info = {
-		"my-page",
-		NULL,
-		"www",
-		"Test page!",
-		NULL,
-		69.0,
-		NULL,
-		0,
-		"C",
+		.page = "my-page",
+		.category = NULL,
+		.site = "www",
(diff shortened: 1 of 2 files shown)


ftml/misc/ffi_test.c
```diff
@@ -0,0 +1,93 @@
...
+	struct ftml_page_info page_info = {
+		"my-page",
+		NULL,
+		"www",
+		"Test page!",
+		NULL,
+		69.0,
+		NULL,
+		0,
+		"C",
```
lambda-fairy: Can we use C99 struct initializers here? I think the field names would make it more readable.

https://stackoverflow.com/a/330867/617159

emmiegit: Oh yeah that's what that syntax was. I usually use ANSI C which would mean filling out the fields individually, but I don't really care about the standard here since it's just for a simple test.

Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #299, opened 2021-05-17
[WJ-573] Remove _ concatenation

emmiegit:
In preparation for [WJ-573](https://scuttle.atlassian.net/browse/WJ-573) eventually adding actual support for soft line breaks, we should just remove the `_` concatenation for now to avoid regressions.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #301, opened 2021-05-17
Move coverage generation for client monorepo to a new workflow job

rossjrw:
I noticed that our sunburst on [CodeCov](https://app.codecov.io/gh/scpwiki/wikijump/) only contained files from `ftml/` and `web/` but was missing `client/`.

I think I've managed to track this down to the [`pnpm test`](https://github.com/scpwiki/wikijump/blob/f6f8b6e0ebeb5b4ec42b4373205da2a80f07c946/.github/workflows/client.yaml#L39-L40) step. Tests are currently failing, so that step errors out, which results in subsequent steps not being executed, therefore coverage is never uploaded.

I'm not sure if this PR will fix it or not, but I'm hoping that it might prompt CodeCov to update our report and that might give us a definite answer. If this doesn't work it would probably be a good idea to split coverage generation out into another command (that doesn't fail), and see if that works instead.

### DIFF
--- a/.github/workflows/client.yaml
+++ b/.github/workflows/client.yaml
@@ -56,7 +56,7 @@ jobs:
             ~/.pnpm-store
           key: ${{ runner.OS }}-pnpm-coverage-${{ hashFiles('**/pnpm-lock.yaml') }}
           restore-keys: |
-            ${{ runner.OS }}-pnpm-
+            ${{ runner.OS }}-pnpm-coverage-
             ${{ runner.OS }}-
 
       - name: Install Node
--- a/.github/workflows/client.yaml
+++ b/.github/workflows/client.yaml
@@ -23,7 +23,6 @@ jobs:
           key: ${{ runner.OS }}-pnpm-${{ hashFiles('**/pnpm-lock.yaml') }}
           restore-keys: |
             ${{ runner.OS }}-pnpm-
-            ${{ runner.OS }}-
 
       - name: Install Node
         uses: actions/setup-node@v2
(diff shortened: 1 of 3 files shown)


Monkatraz:
I think this will fix it but it means that a test failure won't..., uh, fail? I think the proper fix would be to setup coverage to run in another workflow again, and to use the `|| true` thing there instead.

rossjrw:
Yeah - looks like this does work but _at great cost_

I'll revamp this PR to split out coverage generation.

rossjrw:
Alright, looks we're back in business - coverage uploads correctly (so we're down to 63% from 74%), but tests still fail.

`pnpm cover` generates coverage; `pnpm test` just runs tests.

image

.github/workflows/client.yaml
```diff
@@ -42,6 +42,37 @@ jobs:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: PNPM Store Cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.pnpm-store
+          key: ${{ runner.OS }}-pnpm-${{ hashFiles('**/pnpm-lock.yaml') }}
```
Monkatraz: Cache collision here - set this to use something like `-pnpm-coverage-` for the cache key.

Monkatraz (approved):
One change needed but approving just to get things moving.

I fixed tests in https://github.com/scpwiki/wikijump/pull/294/files btw

.github/workflows/client.yaml
rossjrw: Aha, good catch - thanks

```suggestion
          key: ${{ runner.OS }}-pnpm-coverage-${{ hashFiles('**/pnpm-lock.yaml') }}
```

rossjrw: That's a shame, honestly - I would've hoped that the cache used for validation could also be used for coverage. I'm sure GitHub's servers can handle it lmao

PR #302, opened 2021-05-17
[WJ-585] Convert file newlines so tests work on Windows

emmiegit:
If an individual is using Windows and has git checkout with CRLF newlines, this test logic should ensure that the tests still pass like they would on a UNIX-like platform.

### DIFF
--- a/ftml/src/test.rs
+++ b/ftml/src/test.rs
@@ -70,6 +70,17 @@ macro_rules! file_name {
     };
 }
 
+#[cfg(not(target_os = "windows"))]
+fn process_newlines(_: &mut String) {}
+
+#[cfg(target_os = "windows")]
+fn process_newlines(text: &mut String) {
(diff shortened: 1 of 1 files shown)


Monkatraz (approved):
yayyy lmao

emmiegit:
thanks @Monkatraz

ftml/src/test.rs
```diff
@@ -70,6 +70,17 @@ macro_rules! file_name {
     };
 }
 
+#[cfg(not(target_os = "windows"))]
+fn process_newlines(_: &mut String) {}
+
+#[cfg(target_os = "windows")]
+fn process_newlines(text: &mut String) {
+    while let Some(idx) = text.find("\r\n") {
```
lambda-fairy, line 78: `text.replace("\r\n", "")`

?

PR #303, opened 2021-05-17
[WJ-591] Add user block

emmiegit:
This PR adds `[[user xxx]]` and `[[*user xxx]]` to ftml. Exact structure of user display isn't decided and can be adjusted later.


ftml/src/data/user_info.rs
```diff
@@ -0,0 +1,62 @@
...
+
+        UserInfo {
+            user_id: 0,
+            user_name: cow!("michal-frackowiak"),
```
Monkatraz, line 55: how dare ye bring such accursed names to our repository?

ftml/src/render/html/element/user.rs
```diff
@@ -0,0 +1,86 @@
...
+                        .contents(|ctx| {
+                            if show_avatar {
+                                ctx
+                                    .html()
+                                    .img()
+                                    .attr("class", &["small"])
+                                    .attr("src", &[&info.user_avatar_data]);
+                            }
+
+                            ctx.push_escaped(&info.user_name);
```
Monkatraz: For the sake of CSS selectors it would be helpful to shove the username into a `<span>`.

ftml/src/render/html/element/user.rs
```diff
@@ -0,0 +1,86 @@
...
+        log,
+        "Rendering user block";
+        "name" => name,
+        "show-avatar" => show_avatar,
+    );
+
+    ctx
+        .html()
+        .span()
+        .attr("class", &["printuser avatarhover"])
```
Monkatraz: If maintaining Wikidot DOM is the goal this is fine. But otherwise I'd change it to a single class, maybe "userblock" or something like that.

Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #304, opened 2021-05-18
ESLint, Prettier plugin changes and upgrades

Monkatraz:
Depends on https://github.com/scpwiki/wikijump/pull/294 to be merged first

This PR adds three things, all linter related:
* `prettier-plugin-organize-imports`, so that you don't have to think about it and you let Prettier handle it
* `prettier-plugin-jsdoc`, which is slightly finicky but has utterly magical automatic comment formatting
* `eslint-plugin-regexp`, which replaces the old regex plugin which was deprecated in favor of this one (this is a minor upgrade)

Of all of these, the most significant is the second, `pretter-plugin-jsdoc`. This plugin is so magical because it _auto reflows/rewraps comments_. You just don't have to think about the comment wrapping at all, because it just does it for you. It _can_ mess this up, but I had to fix only one or two problems after it had formatted the entire codebase. It works in Svelte, too.

It has some other neat tricks, like running Prettier on example code blocks. Ultimately, using this plugin is a bit of an experiment, but I think the automatic rewrapping is too useful to pass up. You can literally change the comment width of the entire codebase with a single config parameter.


emmiegit (approved):

client/.eslintrc.js
```diff
@@ -113,38 +113,34 @@ const rules = {
...
+        "no-useless-two-nums-quantifier",
+        "no-zero-quantifier",
+        "prefer-predefined-assertion",
+        "hexadecimal-escape",
         "prefer-character-class",
-        "prefer-predefined-character-set",
-        "prefer-predefined-quantifiers",
-        "simple-constant-quantifier",
-        "sort-flags"
+        "prefer-d",
```
rossjrw, line 136: One concern with this is that `\d` will match digits of any language such as `۱۲۳۴۵۶۷۸۹` (Persian) when the regex's author may have intended solely for `0-9`. I think I'd prefer to have the regex author decide what they want to match in this case

rossjrw (approved):

client/.eslintrc.js
Monkatraz: well that's wacky but does make sense, sure. Again I'll do this in another PR, just so I can avoid merge conflicts

PR #305, opened 2021-05-18
[WJ-638] Add paragraph block

emmiegit:
This does normal paragraph construction based on its contents (ensuring paragraph safety), and applies any attributes to the paragraphs generated.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #306, opened 2021-05-18
Hotfix: block rule mapping

emmiegit:
The merge on this file didn't produce a conflict, even though both PRs edit it and increase the count by 1. It's supposed to be increased by 2 because two blocks were added in two PRs.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #307, opened 2021-05-18
Additional client utilities for future components

Monkatraz:
Depends on https://github.com/scpwiki/wikijump/pull/304

This adds:
* The `wj-state` package, which contains or will contain a lot of misc. state-related utilities.
* The `PreferenceHandler` class, which wraps around `localStorage` and makes it extremely easy to handle persistent settings
* A key handling Svelte `use` function
* A popover utility using the same framework that `tippy.js` does, meaning it doesn't add any dependencies
* The `UserAgent` namespace/object/utility, which exports normalized information about the current user's state (in a technical sense)


emmiegit (approved):

client/modules/wj-state/src/useragent.ts
```diff
@@ -0,0 +1,31 @@
...
+   * with caution.
+   */
+  export const isMobile = /Mobi|Android/i.test(navigator.userAgent)
+
+  // set up our listeners
+
+  window.addEventListener("mousemove", evt => {
+    mouseX = evt.clientX / window.innerWidth
+    mouseY = evt.clientY / window.innerHeight
+  })
```
rossjrw, line 24: ??????

why??????

Monkatraz: you won't need these values for precise location, instead it's for rough location with no effort involved.

e.g. if you want a navbar to pop down if the mouse is near the top of the screen, you can just check if mouseY is less than 0.10 or something like that. otherwise you'd have to kinda guestimate a pixel range, which isn't reliable for different device sizes

rossjrw: Huh, alright. I'm down with that.

rossjrw (approved):

PR #309, opened 2021-05-20
Build ftml in php-fpm without logging

emmiegit:
I've thought about this for a bit, but because ftml emits _so much_ logging, and this causes the performance on a CPU-bound task to degrade below the WASM implementation (which has no logging), I think it is better that for our production environment, we disable logs.

We can investigate bugs locally by finding the relevant wikitext that failed and running it, inspecting log output.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #310, opened 2021-05-20
[WJ-456] [WJ-482] Translation/I18n basics, further Sheaf enhancements

Monkatraz:
This PR kind of became a mess, like my previous ones. Unfortunately, the translation issue became pretty blocking to me so I decided to just go ahead and "fix" it.

Primarily, this PR adds:
* Translation/I18n basics, such as a `locales/` folder at root, and a JS library setup to use the locales
* Some enhancements to the structure of Sheaf
* A topbar for Sheaf, which lets you toggle the preview and to see stats
* Converted most of the existing UI to use the new translation system


client/misc/templates/vscode-template/settings.json
```diff
@@ -10,5 +10,13 @@
     "**/.ultra.cache.json": true,
     "**/pnpm-lock.yaml": true
   },
-  "typescript.tsdk": "node_modules\\typescript\\lib"
+  "typescript.tsdk": "node_modules\\typescript\\lib",
```
emmiegit: shouldn't this be `/`? for cross-platform compatibility

client/modules/sheaf/src/PaneEditorTopbar.svelte
```diff
@@ -0,0 +1,172 @@
...
+
+  let chars = 0
+  let bytes = 0
+  let words = 0
+  let lines = 0
+
+  function kilobyteCount(str: string) {
+    // this is a really _spicy_ way of getting a rough kilobyte count
+    // it doesn't need to be precise at all, so this is fine.
+    return Math.round((encodeURI(str).split(/&..|./).length - 1) / 1000)
```
emmiegit, line 26: I don't really understand why this is being done vs just, string length.

client/modules/wj-state/src/i18n.ts
```diff
@@ -0,0 +1,140 @@
...
+import type { MessageFormatter } from "svelte-i18n/types/runtime/types"
+import type { Readable } from "svelte/store"
+
+// -- INIT
+
+// we use `en` as a fallback language, so rather than dynamically importing it
+// we will bundle it with the JS. this may be inefficient but it means that
+// a user with a spotty internet connection won't see every single string as garbage.
+// this should also prevent some edge case errors, hopefully
+import langEN from "../../../../locales/en.yaml"
```
emmiegit, line 14: Do we really want this? It's kind of the hardcoded english issue, but for people with poor connections. Is there some alternative for say, french users?

client/modules/wj-state/src/media.ts
```diff
@@ -1,9 +1,14 @@
-import { writable } from "svelte/store"
+import { Readable, writable } from "svelte/store"
+
+export type BreakpointName = "narrow" | "small" | "normal" | "wide"
+export type BreakpointString = `${">" | "<" | ">=" | "<=" | ""}${BreakpointName}`
```
emmiegit, line 4: I don't think I understand what "breakpoints" mean here.

client/scripts/vite-plugin-yaml.js
```diff
@@ -0,0 +1,25 @@
...
+      if (fileRegex.test(id)) {
+        const obj = yaml.load(src)
+        if (!obj || typeof obj !== "object") {
+          throw new Error(
+            "Invalid YAML provided to plugin. YAML provided must resolve to an object."
+          )
+        }
+
+        return {
+          code: `export default JSON.parse(${JSON.stringify(JSON.stringify(obj))});`,
```
emmiegit, line 21: This is interesting, why does it need to be stringified twice? I guess because the entire file is a JSON string that has a JSON object? Either way, I would suggest leaving a comment explaining this here.

locales/en.yaml
```diff
@@ -0,0 +1,58 @@
...
+    WARNING_SOURCE: >
+      ftml({rule}: {kind} at {token}) [{from, number, :: ,_}, {to, number, :: ,_}]
+    # Warnings
+    RECURSION_DEPTH_EXCEEDED: Too much recursion in markup.
+    NOT_IMPLEMENTED: The syntax ''{slice}'' hasn't been implemented in FTML yet.
+    INVALID_INCLUDE: This include is invalid and won't be rendered.
+    LIST_DEPTH_EXCEEDED: This list is nested too deeply, and can't be rendered.
+    BLOCKQUOTE_DEPTH_EXCEEDED: This blockquote is nested too deeply, and can't be rendered.
+    NO_SUCH_BLOCK: Unknown block ''{slice}''.
+    INVALID_SPECIAL_BLOCK: Block ''{slice}'' doesn't have a special invocation. (starting '*' character)
```
emmiegit: In [WJ-523](https://scuttle.atlassian.net/browse/WJ-523) (#259) "special" was renamed to "star" and "modifier" was renamed to "score". The warnings were also adjusted, so both the key and the message will need adjusting.

client/modules/sheaf/src/PaneEditorTopbar.svelte
Monkatraz: That won't work for a file composed of non Latin characters, like Japanese. It would report a potentially much smaller file size than it should

client/scripts/vite-plugin-yaml.js
Monkatraz: I actually don't know precisely why it's like this, because this technique is copied from how Vite does it internally for JSON files.

client/modules/sheaf/src/PaneEditorTopbar.svelte
emmiegit: Javascript really doesn't have a way of getting bytes?

Or you cannot get UTF-16 codepoints and calculate based on whether they are `> 0xff`?

I'm not sold on `encodeURI()` being faster, but if that's what performance tests say then we should use it.

client/modules/wj-state/src/media.ts
Monkatraz: Breakpoint is a common term for the pixel sizes (e.g. screen which is 400 pixels across, which isn't real pixels but weird CSS relative pixels), where the CSS changes when that size is crossed.

emmiegit: Ohh right, I was thinking about parsing and wondering if this was some editor thing I didn't understand. I was thinking about syntax and forgot `@media` queries existed.

client/modules/sheaf/src/PaneEditorTopbar.svelte
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  function wordCount(str: string) {
+    return str.trim().split(/\s+/).length
+  }
```
rossjrw, line 32: It's going to be a pain to get this to ignore css blocks and stuff lmao

client/modules/sheaf/src/PaneEditorTopbar.svelte
Monkatraz: You think JS has a decent standard library? lmao

Joking aside, `encodeURI` is almost certainly native (as in C or C++ land) and I doubt I could come up with something faster if I tried. I should've put in the comment that I'm not this clever and I didn't come up with this trick, and it's been around a while

client/modules/wj-state/src/i18n.ts
Monkatraz: I can't think of a decent alternative, because the issue is that you'd need to serve a different JS file depending on locale which sounds messy.

Although there actually is a great alternative, using top level await, but that's a really modern language features and isn't supported everywhere yet. It means the JS execution would synchronously pause while the i18n code fetches the translation file. Also this would only work for languages where translation is complete, as otherwise the fallback might be missing while a non translated string is requested

client/modules/wj-state/src/media.ts
```diff
@@ -15,39 +20,68 @@ class MediaQueryHandler {
...
+    return mediaQuery
+  }
+
+  private activeBreakpoint() {
+    // setting the type normally causes TS to over-simplify the type to just "narrow"
+    // for some reason... an assertion fixes that, apparently
+    let current = "narrow" as BreakpointName
+    this.breakpointQueries.forEach((query, idx) => {
+      if (query.matches) current = this.breakpointNames[idx]
+    })
```
rossjrw, line 56: Why forEach and not find?

emmiegit (approved):

client/modules/sheaf/src/PaneEditorTopbar.svelte
emmiegit: Okay yeah, please leave a comment explaining that this trick has been used before. Saying it was "really spicy" made me think it was an original creation, which had me concerned.

client/modules/wj-state/src/i18n.ts
emmiegit: I don't really like how this privileges `en` as a language, vs our approach of the default debug language. While I don't like the idea users seeing debug strings, I'd rather all users see the same debug strings rather than non-english users seeing unexpected english. I think that fits our ideal of "no hardcoded english" better.

rossjrw (approved):

client/modules/sheaf/src/PaneEditorTopbar.svelte
emmiegit: That's not really the point though, it's just a general count of number of space-separated chunks in the work.

client/modules/sheaf/src/PaneEditorTopbar.svelte
Monkatraz: Will do

Yeah JS has a bunch of these "tricks" that you kinda just get used to and they're kind of magical when you see them.

example, removing duplicates from an array:
```ts
arr = [...new Set(arr)]
```

emmiegit: I'm not saying you have to leave exhaustive comments to every weird trick, like I think the set trick above makes sense if you stare at it for a bit, I just think this one is fringe enough to warrant a mention.

client/modules/wj-state/src/i18n.ts
Monkatraz: Unfortunately there is a difference between "the debug language" and synchronously loading `en`. In the mean time while no language is loaded, the translation service has no locales whatsoever and that did cause me some issues, for example if you call formatMessage too quickly it'll error out for a bizarre reason.

So we'd have to synchronously load an actual debug language and not just show the key names. Personally I'd rather just accept the EN fallback, at least if a string is untranslated you have something you can copy paste into Google Translate

emmiegit: I was imagining that we'd have a specific localization file, automatically generated, that just copied key files as translation strings. Does that not make sense in this context?

Though while having english appear when another language should is not ideal, you are true that it is possible for international users to work around it.

Monkatraz: Well if you just copied all the key names that would work, but that's kind of a lot of kilobytes for ultimately still English text, just technical and bizarre looking English text.

emmiegit: Right...

Oh well.

client/modules/sheaf/src/PaneEditorTopbar.svelte
Monkatraz: although I totally didn't think about this.

... I do have access to an AST or token stream................ yes.... very accurate word count....

I'll actually think about solving the word count properly, as I'm sure it'll get used as a point of comparison and I'd like for it to be close to correct.

client/modules/wj-state/src/media.ts
Monkatraz: cuz im real smart, and better than you jeez

I forgot that find existed apparently

client/modules/sheaf/src/PaneEditorTopbar.svelte
Monkatraz: Alright I've thought about it and this isn't so bad to do, because I have access to the editors syntax tree, and there are efficient methods to iterate across the entire tree. All I'd need to do for a 95% accurate word count is go through every `content` leaf and just sum up the word counts. 

This won't catch any text you put inside, say, the argument of a collapsible, but as long as you're not a nutjob and put most of your article in arguments it should be fine. This will work for includes btw

client/modules/wj-state/src/media.ts
Monkatraz: actually find won't work because we need the _last valid breakpoint_, not the first valid one. the list would have to be reversed and I think at that point it would be less clear than just doing this

Monkatraz:
@ammongit I significantly revamped the way the lint rules work, FYI. if you add more warnings they can fairly easily be added yourself now, check https://github.com/scpwiki/wikijump/pull/310/commits/24caa7f7980e927306b03581fbc3eb8258710315 for details

Monkatraz:
thanks @ammongit , @rossjrw 

glad this is finally figured out mostly lol. also, I added an `en-gb` locale with a dummy theme picker tooltip string that uses the word "colour". I made the assumption that `en` is "american" or something, which might be a bad idea but I'm not sure if a "neutral" English exists anywhere

PR #311, opened 2021-05-21
[WJ-444] Yeet Profiles

danieltharp:
This removes the Profile legacy class family and table, as well as any method calls the Profile classes would handle.


web/php/Actions/AccountProfileAction.php
```diff
@@ -226,28 +225,25 @@ public function saveAboutEvent($runData)
...
 
         // check date
         $d = getdate();
         if (checkdate((int) $birthdayMonth, (int) $birthdayDay, (int) $birthdayYear) && $birthdayYear < $d['year']) {
-            $profile->setBirthdayDay($birthdayDay);
-            $profile->setBirthdayMonth($birthdayMonth);
-            $profile->setBirthdayYear($birthdayYear);
+            // Pad with zeroes if needed.
+            if($birthdayMonth < 10) { $birthdayMonth = "0".$birthdayMonth; }
+            if($birthdayDay < 10) { $birthdayDay = "0".$birthdayDay; }
```
emmiegit, line 236: You can use [`str_pad`](https://www.php.net/manual/en/function.str-pad.php) here instead:
```php
$birthdayMonth = str_pad($birthdayMonth, 2, '0', STR_PAD_LEFT);
$birthdayDay = str_pad($birthdayDay, 2, '0', STR_PAD_LEFT);
```
It works even if the input is an integer. Unfortunately the default is to pad right rather than left, which makes the call rather long.

web/templates/modules/Account/Profile/APAboutModule.tpl
```diff
@@ -13,79 +13,19 @@
...
-						<option value="4" {if $profile->getBirthdayMonth() == 4}selected="selected"{/if}>{t}Apr{/t}</option>
-						<option value="5" {if $profile->getBirthdayMonth() == 5}selected="selected"{/if}>{t}May{/t}</option>
-						<option value="6" {if $profile->getBirthdayMonth() == 6}selected="selected"{/if}>{t}Jun{/t}</option>
-
-						<option value="7" {if $profile->getBirthdayMonth() == 7}selected="selected"{/if}>{t}Jul{/t}</option>
-						<option value="8" {if $profile->getBirthdayMonth() == 8}selected="selected"{/if}>{t}Aug{/t}</option>
-						<option value="9" {if $profile->getBirthdayMonth() == 9}selected="selected"{/if}>{t}Sep{/t}</option>
-						<option value="10" {if $profile->getBirthdayMonth() == 10}selected="selected"{/if}>{t}Oct{/t}</option>
-						<option value="11" {if $profile->getBirthdayMonth() == 11}selected="selected"{/if}>{t}Nov{/t}</option>
-						<option value="12" {if $profile->getBirthdayMonth() == 12}selected="selected"{/if}>{t}Dec{/t}</option>
```
emmiegit, line 80: good lord

emmiegit (approved):

web/php/Actions/AccountProfileAction.php
danieltharp: Nice, that sounds better.

Monkatraz (approved):

rossjrw (approved):

PR #312, opened 2021-05-21
[WJ-655] Add CI check for ftml configuration files

emmiegit:
To prevent `blocks.toml` and `modules.toml` from collecting too much drift, this script compares this with scraped information from the block rule source files and errors if discrepancies are found.

In order to make the check pass, this PR also implements existing issues: fixing some fields and adding missing blocks. This also found a bug in `ifcategory` and `iftags`, which for some reason were set to permit score flags, even though the block rule does not handle them.


danieltharp (approved):

emmiegit:
thanks @pxdnbluesoul

PR #316, opened 2021-05-22
Tarnation, CodeMirror FTML grammar significant improvements

Monkatraz:
This brings the CodeMirror grammar much closer to what FTML actually parses. Some of the work involved to do this, is to be frank, disgusting, but I'm working with what I have. In order to massively improve the cleanliness of some sort of parser like this, there would need to be a handwritten one made. If a handwritten one was made, that would be fairly pointless, as we already have FTML.

This also adds two new tabs for the live preview, which is the FTML tokens and editor AST previews. I made my own tree printer to avoid the issues with the last one.

Two important features added by all of this work was code block highlighting, which works wonderfully and nearly took my soul, and CSS highlighting inside style strings. The latter was achieved by creating a fork of the official CSS Lezer grammar and vendoring it in. I wasn't about to reinvent the wheel, especially when doing it this way allows for autocomplete to work for CSS.

Finally, the amount of lines added by this PR is a tad misleading, as about 1800 of them are from an FTML editor syntax test file that consists of a bunch of labeled FTML test cases, which were taken from FTML's actual test cases.


client/modules/cm-lang-ftml/src/grammars/ftml.ts
```diff
@@ -21,112 +19,67 @@ const data = {
...
   ),
 
   // currently empty
   // blk_valmap_el: lkup(
   //   blockEntries
   //     .filter(([, { head, body }]) => head === "value+map" && body === "elements")
-  //     .flatMap(aliases)
+  //     .flatMap(aliases),
+  //  { ignoreCase: true }
   // ),
```
emmiegit, line 69: Commented code?

emmiegit (approved):

client/modules/cm-lang-ftml/src/grammars/ftml.ts
Monkatraz: The `lkup` function isn't intended to take an empty array, so I've commented out this particular combination until a block type fulfills it.

I could make the function work with an empty array, sure, but I've found this kinda useful to keep track of things. Almost certainly a block of this type will get added at some point and uncommenting this will be my verification that everything is working.

Monkatraz:
thanks @ammongit 

the rest of the commits I added are mostly bugfixes, reshuffling about code, or arcane regex tweaks, so there isn't much to review about it

next up: block documentation, hovertips!

PR #317, opened 2021-05-24
[WJ-653] Build gettext localization files 

emmiegit:
This adds a build process to take the locale files in `locales/`, ensure their schemas match, and then construct `.po` files from them, which are then compiled into `.mo` message catalogs suitable for use in deployment.

The process handles parenting, so any `lang_NATION` file inherits from the base language, overriding relevant rules and avoiding code duplication. You can see the existing example with `en` vs `en_GB`.

Related, but `en-gb.yaml` was renamed `en_GB.yaml` to match standard naming conventions for locales.

This also takes the doc comments in the YAML and adds them to the `.po` files. However the way that [ruamel.yaml[(https://yaml.readthedocs.io/en/latest/) handles these means that comments must be one the same line as or after the item they reference. The existing locale files have been adjusted appropriately.

CI checks to both build the message catalog, and lint/format the scripts are in place.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #320, opened 2021-05-27
[WJ-458] [WJ-302] Supercharged FTML block documentation, wikitext Prism highlighting

Monkatraz:
There is still some left to do regarding this topic (that is, Sheaf pseudo-LSP-documentation), but this PR is getting _lorge_, so we'll just review this for now.

This PR adds:
* Hover tooltips for block names, showing their docs as a tooltip
* Supercharged FTML block documentation for Sheaf, which shows up in the autocomplete and hovering tooltips
* A Prism.js grammar for FTML/Wikitext
* Vertical rhythm for wikitext blocks, along with some other styling changes
* Some other misc. fixes and tweaks that were encountered along the way

What's missing?
* Documentation for modules (only works for blocks, currently)
* Translation strings for `pig`, which I did not do because it's a bit of a hassle and the documentation strings are likely to undergo a bunch of revisions (will do after modules are added, for sure)


client/modules/cm-lang-ftml/src/grammars/ftml.ts
```diff
@@ -597,7 +597,7 @@ export const FTMLLanguage = new TarnationLanguage({
         { style: {
           BlockNameInclude: t.keyword,
           IncludeValue: t.link,
-          IncludeParameterProperty: t.propertyName
+          IncludeParameterProperty: t.special(t.propertyName)
```
emmiegit, line 600: Special meaning star, right?

client/modules/cm-lang-ftml/src/tips/BlockTip.svelte
```diff
@@ -0,0 +1,242 @@
...
+  // looks messy but this just assembles a reasonable looking block string
+  let ftmlString = `[[${name}${
+    block.head === "map"
+      ? ' arg="value"'
+      : block.head === "value"
+      ? " value"
+      : block.head === "value+map"
+      ? ' value arg="value"'
+      : ""
+  }]]`
```
emmiegit, line 51: Would a switch be more readable here?

client/modules/cm-lang-ftml/src/tips/BlockTip.svelte
```diff
@@ -0,0 +1,242 @@
...
+  ftmlString = Prism.highlight(ftmlString, "ftml")
+</script>
+
+<div class="cm-ftml-block-tip">
+  <div class="cm-ftml-block-tip-header">
+    <h5 class="cm-ftml-block-tip-title">
+      {docs.title}
+    </h5>
+    <TippySingleton let:tip>
+      <!-- In both types we won't display type "NONE" as that would just be confusing -->
```
emmiegit, line 62: Are these comments passed on to the final HTML?

client/modules/wj-state/src/i18n.ts
```diff
@@ -18,7 +18,7 @@ i18n.addMessages("en", langEN as any)
 // register other languages here
 const localeLoaders = {
   "en": async () => langEN, // dummy import but useful for type information
-  "en-gb": () => import("../../../../locales/en-gb.yaml"),
+  "en-GB": () => import("../../../../locales/en_GB.yaml"),
```
emmiegit: shouldn't the key be `en_GB`?

locales/en.yaml
```diff
@@ -161,3 +160,473 @@ cmftml:
...
+        INFO: This block accepts a body, but interprets that body as raw text.
+      elements:
+        TITLE: ELEMENTS
+        INFO: This block accepts a body, and can nest additional elements within it.
+      other:
+        TITLE: OTHER
+        INFO: This block has a special syntax that isn't easily categorized.
+
+    BLOCK_UNDOCUMENTED: >
+      This block currently does not have any documentation, but it is a valid block.
```
emmiegit:
```suggestion
      This block currently does not have any documentation, but it is valid.
```
?

locales/en.yaml
```diff
@@ -161,3 +160,473 @@ cmftml:
...
+      TITLE: Deletion
+      INFO: >
+        Denotes that the enclosed text has been "deleted" from the document.
+        The text is not literally deleted, instead, it is styled or otherwise
+        set to appear as deleted.
+
+        Depending on context, this may have different meanings. In the context
+        of a story, the deleted text may be a storytelling element, representing
+        something diagetically deleted from a document. In a programming context,
+        it may represent code that was deleted from a file, e.g. in a diff.
```
emmiegit, line 335: Really like the explanations you have here.

emmiegit (approved):

client/modules/cm-lang-ftml/src/grammars/ftml.ts
Monkatraz: This is a detail regarding how CM6 handles styling, nothing to do with the actual grammar. Certain tags, like `special`, and `function`, aren't tags by themselves, but instead wrap around other tags to mark them in some way. The simplest example is `t.function(t.variableName)`, which denotes a function variable.

client/modules/cm-lang-ftml/src/tips/BlockTip.svelte
Monkatraz: Maybe? Switch isn't an expression, so I'd have to have a temp variable or assemble the string the same way in every case, or something like that. I tend to avoid switch because its syntax sucks.

client/modules/cm-lang-ftml/src/tips/BlockTip.svelte
Monkatraz: afaik, no. Svelte assembles HTML "manually", rather than using something like a template string. So it would be pointless to add code to generate the comment nodes.

client/modules/wj-state/src/i18n.ts
Monkatraz: whoops, yeah.

Monkatraz:
thanks @ammongit 

TypeScript type-check is failing due to a weird issue with CodeMirror's view module - says that the types are incompatible due to a difference in private class members, but nothing is _actually_ incompatible. So I'm just ignoring it.

PR #321, opened 2021-05-27
Much better worker-module class/handler thingy

Monkatraz:
In anticipation of adding WASM-based Hunspell checking to Sheaf, I've elected to redo the `worker-module` concept that I first made in the `ftml-wasm-worker` package.

This new one is simply much better. New stuff includes:
* Can be configured to import the worker source only when it needs to
* Can actually be strongly typed
* Fixes a bunch of bugs
* Is a bit faster
* Has a way less confusing interface

On the last point, I'd like to share a demonstration of this. Old, vs new:
```ts
// old
return decode(await invoke<ArrayBuffer>(() => module.worker.renderText(transfer(str))))
// new
return decode(await this.invoke("renderText", transfer(str)))
```

That `invoke` method is fully typed, btw. Typing in `"renderText"` as the string causes the ensuing arguments to take on the correct types for the worker's `renderText` method.


emmiegit (approved):

Monkatraz:
thanks @ammongit

PR #322, opened 2021-05-29
[WJ-660] Fix iftags behavior

emmiegit:
Wikidot's documentation on `[[iftags]]` is incorrect. `+` means that all tags are required (AND logic), and no prefix means that some tags are required (OR logic). When combined, the results are ANDed together (with default being true).

This necessitated some changes to `ElementCondition`, but also allowed simplifying how ifcategory and iftags do their condition checking.

See the [WJ-660](https://scuttle.atlassian.net/browse/WJ-660) issue for information on testing how exactly these tag conditions work.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #323, opened 2021-06-01
Ginormous Tarnation rewrite

Monkatraz:
This PR is for a significant rewrite of Tarnation, excluding the actual grammar definition and compiling system. This rewrite adds an enormously important new feature: incremental parsing/reuse of data _ahead_ of the edit position. That means the tokenizer only tokenizes what actually got changed, at least roughly. This is an UTTERLY ENORMOUS speed gain, it wouldn't be at all unrealistic to say this is about 20 times faster.

Additionally, I took the time to organize and document the code much better this time. There are a lot of doc comments, nearly every new class and function has doc-comments for it.

This still has some bugs, but I doubt me fixing them will change the way the code looks very much. Thus, because of how big this PR is, I'm going to open it up for review now.

Note: there is actually a few minor changes to the grammar system, mainly adding predicates and refactoring the `Matcher` class a little bit.


Monkatraz:
I also want to say that you should not be surprised if you see code in here that could've been done "simply". Like, using `forEach`, `for (const x of y)`, etc. I did not use them, intentionally so. Iterators are basically wrappers around _a ginormous amount of function calls_. They are just slower. So, I used a crap ton of for loops, `arr.slice`, direct access (no destructuring), etc. This is all for the sake of performance, and not an insignificant amount of it either.

leo60228:
> They are just slower.

For the loops, the difference is definitely there, but it's *extremely* small.

http://jsbench.github.io/#315fa2a7e8ac3a4b7db53ff017f1383e

image

For reference, this is about the same performance boost I can get by closing a few tabs.

Monkatraz:
> > They are just slower.
> 
> For the loops, the difference is definitely there, but it's *extremely* small.
> 
> http://jsbench.github.io/#315fa2a7e8ac3a4b7db53ff017f1383e
> 
> image
> 
> For reference, this is about the same performance boost I can get by closing a few tabs.

An example like this doesn't really do what I did justice, to be honest. I specifically profiled the functions involved, primarily the tokenizer and parse step functions. I improved their speed significantly by avoiding destructuring and using only for loops.

I suspect the slowdown comes from the sheer amount of objects involved -- if you use iterators in isolation like this, it gives the code a chance to "get hot" and run at a normal speed.

leo60228:
Did you profile loops and destructuring separately? I definitely do get far worse performance with destructuring compared to indexing, and I'm currently investigating why that is.

leo60228:
Destructuring generates extremely complex bytecode. I really have no idea why that's the case.

Monkatraz:
No, I used Firefox's profiler, which doesn't go into enough detail to figure that distinction out. My strategy was just to repeatedly profile different strategies until I got the fastest one. (I'm on a really good computer so I have to make sure I am optimizing for low power devices)

Monkatraz:
> Destructuring generates extremely complex bytecode. I really have no idea why that's the case.

The destructuring syntax is really complex and flexible, so I'm wondering if that has anything to do with it? For example:

```ts
const [, x, y, ...z] = [...]
```

Like, that has an implicit slice in it. Often newer language features like this have... _interesting_ performance quirks. I think generator functions used to be pretty bad?

client/modules/cm-tarnation/src/host.ts
```diff
@@ -0,0 +1,324 @@
...
+  advance(): Tree | null {
+    if (!this.measurePerformance) this.measurePerformance = perfy()
+    switch (this.stage) {
+      case Stage.Tokenize: {
+        // try to reuse ahead state
+        const reused =
+          this.tokenizerPreviousRight &&
+          this.tokenizer.tryToReuse(this.tokenizerPreviousRight)
+
+        // can't reuse the buffer more than once (pointless)
```
emmiegit, line 256: Why is this? Out of curiosity?

client/modules/cm-tarnation/src/parser/parser.ts
```diff
@@ -0,0 +1,257 @@
...
+          // this doesn't affect the buffer at all, but now we can watch for
+          // when another node closes one of the open nodes we added
+          // if (open) {
+          if (t[3]) {
+            // for (let i = 0; i < open.length; i++) {
+            for (let i = 0; i < t[3].length; i++) {
+              // const [id, inclusive] = open[i]
+              const o = t[3][i]
+              // ctx.stack.push(id, inclusive ? from : to, type ? (inclusive ? 0 : -1) : 0)
+              ctx.stack.push(o[0], o[1] ? t[1] : t[2], t[0] ? (o[1] ? 0 : -1) : 0)
```
emmiegit: Might be more readable if it was split onto multiple lines:
```suggestion
              ctx.stack.push(
                o[0],
                o[1] ? t[1] : t[2],
                t[0] ? (o[1] ? 0 : -1) : 0),
              )
```

client/modules/cm-tarnation/src/parser/parser.ts
```diff
@@ -0,0 +1,257 @@
...
+   * Fully advances the parser, using the tokenizer's emitted chunks
+   * directly with no additional parsing. This is for debug purposes, as it
+   * allows distinguishing between bugs that orginate from either the
+   * tokenizer or the parser.
+   */
+  advanceFullyRaw(): { buffer: number[]; reused: Tree[] } {
+    const buffer: number[] = []
+    for (let idx = 0; idx < this.pending.length; idx++) {
+      const tokens = this.pending[idx].compile()
+      if (tokens.length === 0) console.error("what")
```
emmiegit: Can you add a more useful error message here?

emmiegit (approved):

client/modules/cm-tarnation/src/host.ts
Monkatraz: The reasoning is really simple, actually. The tokenizer emits to a buffer which holds a bunch of chunks. If we reuse another buffer on the right (as in, we concatenate it), we just move all the way to the end of the newly concatenated buffer. We reuse the entire thing, so there isn't any point in checking if we can still reuse it. We already did!

client/modules/cm-tarnation/src/parser/parser.ts
Monkatraz: good idea, made a commit where I did the same to the commented out destructured version

client/modules/cm-tarnation/src/parser/parser.ts
Monkatraz: whoops missed this console log. removed it in a commit

Monkatraz:
While there is still some things I could do on this, I think I should stop for now. It does work - 95% of the time, anyways. It's really stupid fast though, so it's worth it.

thanks @ammongit

PR #324, opened 2021-06-03
Massively improve font consistency by adding CJK Noto fonts, Roboto fallback font

Monkatraz:
I couldn't sleep, so I did this, as it was bugging me.

For the 5000th commit, this PR adds proper CJK font support in the CSS for the new frontend. It's dynamic based on locale, and has full support for Japanese, Korean, and Chinese (Simplified, Traditional, HK). Additionally, Roboto and Roboto Mono have been added as fallback fonts. This improves the appearance of Cyrillic text and other related scripts due to Roboto's wide Unicode support.

Now, there are some issues to think about:
* With the CJK fonts, in what order should they be in? If you're Japanese, would you prefer a fallback Chinese character to be in traditional or simplified script? This extends to pretty much every CJK locale so we'll probably need to ask around.
* Are the huge sizes of the CJK fonts a problem? The variable fonts (not mono) are around 5MB, but the mono ones are 10MB+.  Personally, I don't think it's too big of an issue, as the font downloads are non-blocking and have system-based fallbacks.
* What should the default CJK font be for a non-CJK locale? I picked Japanese, but that's because I just arbitrarily picked it.

I can't answer these questions on my own, so there isn't much I can do other than get this PR out there.

Screenies:
image
image


emmiegit (approved):
LGTM, though we should consult some international people to figure out specifics

leo60228:
> What should the default CJK font be for a non-CJK locale? I picked Japanese, but that's because I just arbitrarily picked it.

Just to record a short discussion on Discord:
* Firefox tries the system locale if it's CJK, followed by the hardcoded order `ja`, `zh-CN`, `zh-HK`, `zh-TW`, `ko`
* Chromium uses the system locale if it's CJK, and hardcodes `zh-Hans` as a fallback.

Note that the handling here is slightly different even beyond the fallback order, as Chromium determines a single fallback locale while Firefox tries several locales in order.

Monkatraz:
thanks @ammongit , @leo60228 , and Jochoi on the discord

I elected to follow Firefox's fallback order, and I also added some language selectors for the `zh-Hans` and `zh-Hant` codes, which should mostly account for these issues anyways

PR #325, opened 2021-06-07
[WJ-457] Sheaf spellchecking, other significant improvements

Monkatraz:
This PR adds spellchecking to Sheaf, along with some other significant changes. Here is the general list of changes:
* Sheaf spellchecking, using the `spellchecker-wasm` package
* Dictionaries for said spellchecker (this is where the huge line count comes from)
* Worker for getting stats and info from big chunks of wikitext
* Refactor/clean-up of the `sheaf-core` package
* Made the `wj-prism` package a little more advanced, languages load async now, a vendored version of the minified Prism library with no languages has been added for use in workers
* Added a Prism grammar for Svelte
* Added the `EditorField` class to `sheaf-core` for making easier to add extensions based off state fields
* and last but not least, added a path alias for `@root`


client/modules/sheaf-core/src/extensions/spellcheck/extension.ts
```diff
@@ -0,0 +1,59 @@
...
+import { SpellcheckState } from "./state"
+
+// this is apparently how CodeMirror does underlines,
+// I figured it was just a text underline, but no, it's actually this
+// kind of interesting
+function underline(color: string) {
+  if (typeof btoa !== "function") return "none"
+  let svg = `<svg xmlns="http://www.w3.org/2000/svg" width="6" height="3">
+    <path d="m0 3 l2 -2 l1 0 l2 2 l1 0" stroke="${color}" fill="none" stroke-width=".7"/>
+  </svg>`
```
emmiegit, line 16: lol wow

client/modules/sheaf-core/src/extensions/spellcheck/spellchecker/dicts.ts
```diff
@@ -0,0 +1,49 @@
...
+  }),
+  "it": async () => ({
+    dict: await url(import("../../../../../../../locales/spellcheck/it.txt?url"))
+  }),
+  "ru": async () => ({
+    dict: await url(import("../../../../../../../locales/spellcheck/ru.txt?url"))
+  }),
+  "zh": async () => ({
+    dict: await url(import("../../../../../../../locales/spellcheck/zh.txt?url"))
+  })
```
emmiegit, line 41: No dictionary for `pig`? :stuck_out_tongue:

client/modules/sheaf-core/src/extensions/spellcheck/spellchecker/normalize.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * Pairs of strings, with the first string being text to replace with the
+ * latter. Used for normalizing text.
+ */
+const REPLACEMENTS: [string, string][] = [
```
emmiegit, line 5: Add a comment/todo about how this is latin-centric?

client/modules/sheaf-core/src/extensions/spellcheck/spellchecker/normalize.ts
```diff
@@ -0,0 +1,42 @@
...
+  ["’", "'"] // normalize right single quotation marks into apostrophes
+]
+
+/** Returns a list of patterns to filter out as "not words", depending on locale. */
+function getFilters(locale: string): RegExp[] {
+  switch (locale) {
+    case "en":
```
emmiegit, line 12: Does this capture `en_GB` etc.?

client/modules/wj-prism/src/index.ts
```diff
@@ -17,8 +15,18 @@ const encode: (src: string) => string = Prism.util.encode as any
...
-prismFTML(Prism)
+// asynchronously import languages to prevent the large prism JS file
+// from blocking the page from rendering
+// only the base library is synchronously imported and everything else is
+// asynchronously imported
+// prettier-ignore
+async function importLanguages() {
+  ;(await import("../vendor/prism-langs")).prismBase(Prism)
+  ;(await import("../vendor/prism-svelte")).prismSvelte(Prism)
+  ;(await import("./ftml")).prismFTML(Prism)
```
emmiegit, line 26: why `;(`? why not `;-)`?

client/modules/wj-util/src/index.ts
```diff
@@ -430,3 +434,42 @@ export function html(strings: TemplateStringsArray, ...subs: (string | string[])
...
+  return str.substr(0, from) + sub + str.substr(to)
+}
+
+/**
+ * Capitalizes (or lowercases, if set) a string. Does not uppercase the
+ * entire string, just the first character.
+ *
+ * @param str - The string to capitalize.
+ * @param state - Whether to capitalize or lowercase. Defaults to capitalize.
+ */
```
emmiegit, line 457: All of these only work on ASCII and should be noted as such.

If we want to check case-insensitively, you will need to do a unicode normalize/case fold.

client/modules/wj-util/src/index.ts
```diff
@@ -430,3 +434,42 @@ export function html(strings: TemplateStringsArray, ...subs: (string | string[])
...
+export function capitalize(str: string, state = true) {
+  if (!state) return str.toLowerCase()
+  return replaceRange(str.toLowerCase(), 0, 1, str[0].toUpperCase())
+}
```
emmiegit, line 460: Python uses "title" for this operation, to avoid confusion. When I first saw this I thought it meant upper-case.

emmiegit (approved):

client/modules/wj-util/src/index.ts
Monkatraz: I thought `toLowerCase` and `toUpperCase` accounted for the characters it was checking, but if it doesn't I think JS has another function for that.

one google search later:

Okay, so `toLowerCase` and `toUpperCase` do correctly handle non-ASCII text, however certain locales handle the same text in different ways, such as the `TR` locale. For that, there is `toLocaleLowerCase` and `toLocaleUpperCase`, which about 99% of the time is going to give the same result as the ordinary functions. I could hook that up into these functions, using an optional `locale` argument, I think.

client/modules/wj-prism/src/index.ts
Monkatraz: When you're writing JS without semicolons at the end of the line, it's a convention to stick semicolons at the start of the line. There are probably many ways to argue this, but for me this is helpful because it makes the semicolon really obvious.

client/modules/sheaf-core/src/extensions/spellcheck/spellchecker/normalize.ts
Monkatraz: It's intended to only handle the language code. You're supposed to feed into the spellchecker just a locale's language, so by the time it gets to this function it shouldn't have any region codes attached.

That could cause problems later, but honestly, I don't really know. It adds quite a bit of complexity if regions are supported, because then we start talking about having to merge dictionaries (probably) or worse.

client/modules/sheaf-core/src/extensions/spellcheck/spellchecker/dicts.ts
Monkatraz: I mean, we could autogenerate one lol

client/modules/sheaf-core/src/extensions/spellcheck/spellchecker/normalize.ts
Monkatraz: My hope with this specific object is that it's just for replacing unusual substitutes for characters - I don't think there should be any language specific replacements, or if there is, it's language-specific enough that it doesn't conflict with anything.

client/modules/wj-util/src/index.ts
Monkatraz: good idea

client/modules/wj-util/src/index.ts
emmiegit: Okay, cool. Can you leave this as a comment here? Since it looks like a naive ASCII implementation at first glance.

client/modules/wj-prism/src/index.ts
emmiegit: yeah, I figured, but I wanted to make a dumb joke

Monkatraz: i'm an idiot lmfao

client/modules/wj-util/src/index.ts
Monkatraz: donezo, added with locale support

Monkatraz:
thanks @ammongit 

next up: uh... probably the `wj-codemirror`, `sheaf-core`, `sheaf` reorganization crap, and then hopefully maybe probably I'll get started on adding drafts, which should be pretty fun.

PR #326, opened 2021-06-07
Consolidate CodeMirror dependencies, other repo cleanup stuff

Monkatraz:
This:
* Consolidates all CodeMirror packages (e.g. `@codemirror/view`) into a single file that exports _everything_. This is found in the new `wj-codemirror` package, which will also contain generic extensions (like the spellchecker), and allow for me to eventually merge the `sheaf-core` and `sheaf` packages together. Additionally, having all CodeMirror packages depended on by only a single package makes it much easier to handle the versioning and updates.
* Renames the `components` package to `wj-components`, to match with other generically named packages
* Renames `worker-module` to `threads-worker-module` as `worker-module` is already an NPM package and I'd rather the names not conflict
* Cleans up some dependencies in various packages, like Svelte not being in `devDependencies` like it should be


emmiegit (approved):

Monkatraz:
thanks @ammongit

PR #328, opened 2021-06-13
[WJ-457] (WIP) Use NSpell for spellchecker, new cm-nspell package

Monkatraz:


emmiegit (approved):

Monkatraz:
thanks @ammongit 

I'm going to probably immediately start working on the Spylls port - it might take a while. There is technically "better things I could be doing", but all of those are pretty much blocked by something else right now. A lot of the progress that needs to be made with Wikijump is on the backend, unfortunately.

PR #333, opened 2021-06-18
Remove profiles from recommended hosts entries

emmiegit:
Because we have removed the profiles site, we don't need those hosts entries anymore.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #339, opened 2021-06-23
[WJ-657] Add line centering

emmiegit:
We already have the alignment blocks `[[=]]` etc., but this PR adds `=` centering. For instance:

```
= This text is centered.

This text is not.
```

Similar to blockquotes, because it is newline-dependent, it is attached to the input start and line break rules in the mapping.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz @Yossipossi1

PR #344, opened 2021-07-08
[WJ-580] Record backlinks in ftml, export to PHP

emmiegit:
[WJ-580](https://scuttle.atlassian.net/browse/WJ-580).

This adds a `Backlinks` structure to ftml, which stores links as they're encountered. The structure is added to FFI and wasm. Then in the PHP wikitext service, the FFI data is extracted and saved in an instance of the `Backlinks` class, where it's consumed by existing Wikidot logic.

Additional changes are made to support this addition: for instance the legacy database seeder's sequence value is updated to prevent primary key collisions.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #345, opened 2021-07-19
[WJ-574] Add headings

emmiegit:
This PR adds support for headings, H1 through H6, to ftml. These are done using a `+` to `++++++` at the beginning of a line, each corresponding to its respective heading level. It also adds tests for headings, including failure cases.

The text renderer is also modified to strip all leading and trailing newlines from its output.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #346, opened 2021-07-20
[WJ-693] Add property testing to ftml

emmiegit:
This PR adds property testing [property testing](https://altsysrq.github.io/proptest-book/proptest/getting-started.html) to ftml. This acts similar to a fuzzer, in that input data is generated randomly based on a specification, and then tests are run using it to ensure there are no crashes or unexpected results.

Tests are added for:
* Preprocessing
* UTF-16 index mapping
* Depth list parsing
* Tokenization
* Rendering to HTML
* Rendering to text

The majority of code here generates a fuzzed abstract syntax tree by recursively generating elements.


ftml/src/preproc/whitespace.rs
```diff
@@ -26,11 +26,6 @@
 //! * Convert tabs to four spaces
 //! * Convert null characters to regular spaces
 //! * Compress groups of 3+ newlines into 2 newlines
-//!
-//! Note on the first item:
-//! It was originally implemented in the parser, however it was moved here
-//! to prevent typography from converting the `--` in `[!--` and `--]` into
-//! em dashes.
```
emmiegit, line 33: These are minor documentation changes, since while making this PR I noticed they were out of date.

ftml/src/tree/attribute/safe.rs
```diff
@@ -176,7 +179,7 @@ pub fn is_safe_attribute(attribute: UniCase<&str>) -> bool {
     }
 
     for prefix in &SAFE_ATTRIBUTE_PREFIXES {
-        if attribute.starts_with(prefix) {
+        if attribute.starts_with(prefix) && ATTRIBUTE_SUFFIX_SAFE.is_match(&attribute) {
```
emmiegit, line 182: This was a bug that was found that could affect `data-xxx` keys, where a user could create an attribute which would result in an assertion tripping.

emmiegit:
This has been outstanding for a while, and is blocking newer work.

PR #347, opened 2021-07-22
[WJ-593] List item blocks

emmiegit:
This PR adds block versions of the list elements. `[[ul]]`, `[[ol]]`, and `[[li]]`, along with score support (e.g. `[[ul_]]`). It also has new AST tests to ensure functionality.


emmiegit:
This has been outstanding for a while, and is blocking newer work.

PR #349, opened 2021-07-27
[WJ-576] Move base ftml CSS to separate file

emmiegit:
This PR adds `misc/ftml-base.css`, which contains basic styling for generated HTML, and methods which output the CSS for convenience. These are behind the `css` feature flag, so as to be excludable when binary size is a concern. All outputted CSS classes are prefixed with `wj-` for namespacing.


emmiegit:
Merging for now.

PR #351, opened 2021-07-28
Fix invalid method call in UserInfoWinModule template

danieltharp:
Brain fart on a conditional causes the info panel to fail to load. Fix is tested working.


emmiegit (approved):
:wrench:

PR #355, opened 2021-07-30
Fix ftml configuration for list blocks

emmiegit:
The CI build didn't fail when the list blocks were added (`[[li]]`, `[[ol]]`, `[[ul]]`) when it should've. This changes the build trigger to activate when any block updates, and adds them to the configuration.

It also updates dependencies and updates the stable Rust version.


danieltharp (approved):

emmiegit:
thanks @pxdnbluesoul

PR #356, opened 2021-08-01
Fix locale build

emmiegit:
The localization build was failing because `pig` had untranslated strings, which are added here.

To assist with identifying untranslated strings, I added a step in the build process that reports which fields are missing / extra:
```
Loading localizations...
+ en
+ pig
+ en_GB

The following localizations do not match the schema:
- pig
    Missing Fields:
     - sheaf.spellcheck.MISSPELLED_WORD
     - sheaf.spellcheck.SOURCE
     - sheaf.spellcheck.ADD_TO_DICTIONARY
     - sheaf.spellcheck.FORBIDDEN_WORD
     - sheaf.spellcheck.WARNED_WORD
     - sheaf.spellcheck.tooltips.ADD_TO_DICTIONARY
     - sheaf.spellcheck.tooltips.ACCEPT_SUGGESTION
```

After this was fixed, several of the strings containing newlines or quotes was causing issues with parsing the generated `.mo` files. This was ameliorated by addressing the string escape mechanism. Unfortunately, Python's `unicode_escape` functionality doesn't escape quote characters, and hacking with `repr()` (which prefers single-quote strings) or replace would be a mess, so instead JSON string representation is used.


danieltharp (approved):

emmiegit:
thanks @pxdnbluesoul

PR #358, opened 2021-08-04
[WJ-274] Contacts, Interactions, and Followers

danieltharp:
This is a comprehensive PR around some new core concepts to the Wikijump platform. The centerpiece of it is Interactions, which is a dual-polymophic table associating two objects of similar or disparate types on a particular type of relation.

Interaction types are defined by the `InteractionTypes` class which exposes a set of constants we're treating as `ENUM`s. A `HasInteractions` trait allows a model class to interact with helper functions around the Interaction class static methods, and returns the actual target Eloquent models rather than the intermediate Interaction object.

This concept is initially used to replace the `Contact` classes with new contact interactions, and also adds an intermediate step of contact requests. (We are also removing code around exposing email addresses for bidirectional contacts.) This PR also adds the ability for a user to follow another user, a feature that will later be augmented through interactions on a page update.

Feature tests are provided for Follows and Contacts.


web/app/Http/Controllers/InteractionController.php
```diff
@@ -0,0 +1,85 @@
+<?php
+
```
emmiegit, line 3: Some files are missing strict types:
```suggestion
declare(strict_types = 1);

```

web/app/Models/Interaction.php
```diff
@@ -0,0 +1,184 @@
...
+     * @return Interaction
+     */
+    public static function add($setter, int $relation, $target, ?array $metadata = []) : ?Interaction
+    {
+
+        /**
+         * We have to do some validation here as we're accepting generics.
+         */
+        if(is_subclass_of($setter, 'Illuminate\Database\Eloquent\Model')
+            && is_int($relation)
```
emmiegit: You can use [`Enum::isValue()`](https://github.com/scpwiki/wikijump/blob/develop/web/app/Common/Enum.php#L48) to check if it's a valid object.

web/app/Models/Interaction.php
```diff
@@ -0,0 +1,184 @@
...
+                    'metadata' => $metadata
+                ]
+            );
+
+            try {
+                $interaction->save();
+                return $interaction;
+            }
+            catch(QueryException $e) {
+                if($e->errorInfo[0] == 23505) {
```
emmiegit, line 99: Where does "23505" come from?

web/app/Models/User.php
```diff
@@ -80,6 +84,24 @@ class User extends Authenticatable
...
+    public function avatarSm(): string
+    {
+        return '/common--images/avatars/'.floor($this->id/1000).'/'.$this->id.'/a16.png';
+    }
+
+    /**
+     * Retrieve the path for the user's large avatar.
+     * @return string
+     */
+    public function avatarLg(): string
```
emmiegit: Personal preference: I think `avatarSmall`/`avatarLarge` is more readable.

web/app/Models/User.php
```diff
@@ -80,6 +84,24 @@ class User extends Authenticatable
...
+    {
+        return '/common--images/avatars/'.floor($this->id/1000).'/'.$this->id.'/a48.png';
```
emmiegit, line 102: My suggestion is `"/avatars--common/$id/large.png"`.

web/tests/php/Feature/UserFollowUserTest.php
```diff
@@ -0,0 +1,73 @@
...
+ * @package Tests\Feature
+ */
+class UserFollowUserTest extends TestCase
+{
+    use RefreshDatabase;
+
+    /**
+     * @var User $user
+     */
+    private $user;
```
emmiegit: Does `private User $user;` work?

emmiegit (approved):

web/app/Models/User.php
danieltharp: This will be in another PR, it is set in about 30 places in the code.

PR #359, opened 2021-08-05
[WJ-444] User-to-user Blocks and new Interaction methods

danieltharp:
This PR refactors away the `PrivateUserBlock` methods and table in favor of the `USER_BLOCKS_USER` Interaction and includes full test coverage of the new methods, and introduces some interactivity between interactions: A blocked user can't send you a contact request. 

It refactors the Interactions class to have all methods return booleans, except for the `retrieve` class which will return an `Interaction` object directly. It also adds an `Interaction::set()` method to update metadata attached to an interaction.

It also includes some small bugfixes around the `Enum` helper class.


web/app/Models/Interaction.php
```diff
@@ -61,73 +61,148 @@ public function target() : MorphTo
...
+                'target_id' => $target->id,
+                'metadata' => $metadata
+            ]
+        );
+
+        try {
+            return $interaction->save();
+        }
+        catch(QueryException $e) {
+            /** Postgres unique constraint violation: */
```
emmiegit, line 116: Is there a way we can wrap `QueryException`? Having to remember magic postgres error codes doesn't seem like great ergonomics long term.

web/app/Models/Interaction.php
```diff
@@ -61,73 +61,148 @@ public function target() : MorphTo
...
+     */
+    public static function retrieve($setter, $relation, $target) : ?Interaction
+    {
+        if(self::isInvalid($setter, $relation, $target))
+        {
+            Log::error("Interaction::retrieve was given an invalid set of params. 
             Setter: $setter Relation: $relation Target: $target");
-        abort(500);
-        return null;
+            abort(500);
```
emmiegit, line 142: What does `abort()` do?

web/app/Models/User.php
```diff
@@ -120,34 +123,33 @@ public function followingUsers() : Collection
...
-     * @return Interaction
+     * @return bool
      */
-    public function followUser(User $user_to_follow) : Interaction
+    public function followUser(User $user_to_follow) : bool
     {
-        return Interaction::add($this, InteractionType::USER_FOLLOWS_USER, $user_to_follow);
+        if ($this->isFollowingUser($user_to_follow) === false)
+        {
+            return Interaction::create($this, InteractionType::USER_FOLLOWS_USER, $user_to_follow);
```
emmiegit, line 141: For this an several other interaction methods: Should these be in transactions to avoid data races?

web/app/Models/User.php
```diff
@@ -157,133 +159,259 @@ public function unfollowUser(User $user_to_unfollow) : int
...
+     * @param string $reason
+     * @return bool
      */
-    public function approveContactRequest(User $user_to_approve): ?Interaction
+    public function blockUser(User $user_to_block, string $reason = '') : bool
     {
-        Interaction::remove($user_to_approve, InteractionType::USER_CONTACT_REQUESTS, $this);
-        return $this->addContact($user_to_approve);
+        /** Do not add a second block for the same target user. */
+        if($this->isBlockingUser($user_to_block)) { return false; }
```
emmiegit, line 339: This makes me think, are there any checks for any interaction methods that prevents interactions with oneself? e.g. following or blocking yourself

web/php/Actions/AccountSettingsAction.php
```diff
@@ -156,59 +154,41 @@ public function saveReceiveMessagesEvent($runData)
...
-
-        // check max
-        $c = new Criteria();
-        $c->add("user_id", $runData->getUserId());
-        $blockCount = PrivateUserBlockPeer::instance()->selectCount($c);
-
-        $maxBlocks = 30;
-
-        if ($blockCount>$maxBlocks) {
-            throw new ProcessException("Sorry, you can only block $maxBlocks users max.", "max_block");
```
emmiegit, line 188: Max blocks? How arbitrary and terrible for high-traffic users.

emmiegit (approved):

web/app/Models/User.php
danieltharp: There are, they're in the Event controller right now, when I get to refactoring the actual controllers (requires API) we'll make some determinations on whether it makes sense to have the validation in the controller or the model method, I'm leaning towards the controller but could be convinced either way.

web/app/Models/User.php
danieltharp: Yes. I had it in my head that I couldn't mix transactions and Eloquent but that doesn't look to be the case, I will add these where it makes sense to before I merge.

web/app/Models/Interaction.php
danieltharp: `abort()` returns a response to the caller with the provided HTTP status code instead of a `die()` statement which does not offer that. This is a stub at the moment, when the controller gets added to the mix we'll throw various exceptions to be provided to the user.

web/app/Models/Interaction.php
danieltharp: I believe there is already a specific exception for unique constraint violations, I'll try and find it.

web/app/Common/PGError.php
```diff
@@ -1,8 +1,196 @@
...
+    public const READING_SQL_DATA_NOT_PERMITTED = '2F004';
+    public const INVALID_CURSOR_NAME = '34000';
+    public const EXTERNAL_ROUTINE_EXCEPTION = '38000';
+    public const CONTAINING_SQL_NOT_PERMITTED = '38001';
+    public const MODIFYING_SQL_DATA_NOT_PERMITTED_38002 = '38002';
+    public const PROHIBITED_SQL_STATEMENT_ATTEMPTED_38003 = '38003';
+    public const READING_SQL_DATA_NOT_PERMITTED_38004 = '38004';
+    public const EXTERNAL_ROUTINE_INVOCATION_EXCEPTION = '39000';
+    public const INVALID_SQLSTATE_RETURNED = '39001';
+    public const NULL_VALUE_NOT_ALLOWED_39004 = '39004';
```
emmiegit, line 116: Why do some of these end in numbers?

web/app/Models/User.php
```diff
@@ -230,8 +231,12 @@ public function requestContact(User $user_to_request): bool
      */
     public function approveContactRequest(User $user_to_approve): bool
     {
-        Interaction::remove($user_to_approve, InteractionType::USER_CONTACT_REQUESTS, $this);
-        return $this->addContact($user_to_approve);
+        DB::transaction(function () use ($user_to_approve) {
+            Interaction::remove($user_to_approve, InteractionType::USER_CONTACT_REQUESTS, $this);
+            return $this->addContact($user_to_approve);
+        });
```
emmiegit, line 237: How do returns from `DB::transaction()` work? Why is it just `return false` at the end now?

web/app/Common/PGError.php
danieltharp: They are duplicate keys otherwise, they have the same error code twice in here, and I'm not familiar enough with the particulars of these errors to come up with a better name. We'll have to see if they're ever relevant and then maybe we refactor around it.

web/app/Models/User.php
danieltharp: DB::transaction returns whatever return was invoked from the call within, which is already a boolean. The `return false;` at the end is largely to get PHPStorm to shut up because it will either return true or false within, or throw an exception, but PHPStorm complained about no return. I could also add a phpdoc telling it to ignore that inspection, but it's already in a few places here.

web/app/Common/PGError.php
emmiegit: Oh that's strange. Well this workaround makes sense to me: you may want to consider adding a comment like above to explain it to future versions of ourselves.

web/app/Models/User.php
emmiegit: Interesting, so `DB::transaction();` can act as a `return` in a function, even though it's a call? Or is it because `return` in the lambda returns from the outer function?

danieltharp: I should just be returning the result of the call, you're right. https://github.com/laravel/framework/blob/4.1/src/Illuminate/Database/Connection.php#L417 shows how it works, a new commit has it return the result from `DB::transaction()` and it passes all tests.

emmiegit: Ah okay, this makes more sense now.

PR #360, opened 2021-08-08
[WJ-705] Allow ARIA attributes in ftml

emmiegit:
This adds [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) attributes as allowed in ftml, which permits power users to improve accessibility on wikitext they create.


danieltharp (approved):

emmiegit:
thanks @pxdnbluesoul

PR #363, opened 2021-08-09
Check blocks/modules docs automatically

emmiegit:
This updates `check_conf.py` to do a loose search of the documentation files to prompt the human to update them accordingly when blocks or modules are added or changed.

The PR also adds missing entries (so that the check passes), and fixes a discovered bug around the table of contents not having a prefixed ID (`toc` instead of `wj-toc`).


danieltharp (approved):

emmiegit:
thanks @pxdnbluesoul

PR #364, opened 2021-08-11
[WJ-444] User Messages, Bitmasks, UUIDs, and Policies

danieltharp:
This PR is the last in WJ-444's scope of refactoring the classes and tables that are first-level relations of `User`, this time focusing on Private Messages. 

- The new `UserMessages` class makes use of [Ordered UUIDs](https://laravel.com/docs/8.x/helpers#method-str-ordered-uuid) for reduced predictability of messages while still offering a nicely sortable and indexable primary key. A new `UsesUUIDs` trait has been built to facilitate this for any table that will use a UUID in lieu of an incrementing PK.
- The class is also the first to use bitwise flags (bitmasks) for managing independent boolean values in a single INT field, and so a `UsesBitmasks` trait has also been added along with some helpers, `getFlag()`, `setFlag()`, and `clearFlag()`. For PMs, the `flags` field is tracking is a message is unread, a draft, starred, archived, or a sent copy of a message, all independently in a single field (without using our predecessors design pattern of concatenating random letters together and checking for the existence of one of those letters.
- This class is the first to make use of an [Authorization Policy](https://laravel.com/docs/8.x/authorization#writing-policies). These will be used heavily going forward for authorization, as a part of role-based access control. As we get closer to the API and using controllers, the policy checks will likely be moved there so the response surfaces faster.
- Full unit and feature testing has been added for bitmasks, UUIDs, authorization policies, and the messages themselves.


web/app/Models/UserMessage.php
```diff
@@ -0,0 +1,356 @@
...
+        if($permission->denied())
+        {
+            throw new AuthorizationException($permission->message());
+        }
+
+        return DB::transaction(function() {
+            $this->save();
+
+            $sent = $this->replicate();
+            $sent->setFlag(UserMessage::MESSAGE_SENT);
```
emmiegit, line 70: Is there a reason you can't do `$this->setFlag(UserMessage::MESSAGE_SENT)` and then `$this->save()`?

web/app/Models/UserMessage.php
```diff
@@ -0,0 +1,356 @@
...
+
+    /**
+     * Generate a preview of a message.
+     * @param int $length
+     * @return string
+     */
+    public function preview(int $length = 200) : string
+    {
+        $preview = substr(strip_tags($this->body), 0, $length);
+        if (strlen($preview) == $length) {
```
emmiegit: Even though it's the same here, I think `>=` is more readable since it conveys "truncation" as a programmer intent better.

web/app/Models/UserMessage.php
```diff
@@ -0,0 +1,356 @@
...
+        return $this->getFlag(UserMessage::MESSAGE_ARCHIVED);
+    }
+
+    /**
+     * Is this message not archived?
+     * @return bool
+     */
+    public function isNotArchived() : bool
+    {
+        return ! $this->getFlag(UserMessage::MESSAGE_STARRED);
```
emmiegit, line 335: Are these `isNotXXX()` methods really necessary? We can just do `!message.isArchived()`

web/app/Policies/UserMessagePolicy.php
```diff
@@ -0,0 +1,80 @@
...
+     * Global allow and deny for User Policies.
+     * @see https://laravel.com/docs/8.x/authorization#policy-filters
+     * @param User $user
+     * @param string $ability
+     * @return bool|null
+     */
+    public function before(User $user, string $ability) : ?bool
+    {
+        /** The farm admin account can do anything. */
+        if($user->id == 1) { return true; }
```
emmiegit: There isn't a `User::ADMINISTRATOR_USER` in https://github.com/scpwiki/wikijump/blob/develop/web/app/Models/User.php#L42-L43, but I think there should be. This way we don't have a naked ID value like this, which I think causes more confusion down the road (though you do have a comment here so I think it's _okay_)

web/database/factories/UserMessageFactory.php
```diff
@@ -0,0 +1,35 @@
...
+     * Define the model's default state.
+     *
+     * @return array
+     */
+    public function definition(): array
+    {
+        return [
+            'subject' => $this->faker->sentence,
+            'body' => $this->faker->paragraph,
+            'flags' => 0 // Unread, non-draft, unstarred, unarchived message.
```
emmiegit, line 32: :+1: for the comment

web/database/migrations/2021_08_06_003806_create_user_messages_table.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        Schema::create('user_messages', function (Blueprint $table) {
+            $table->uuid('id')->primary();
+            $table->foreignId('from_user_id')->index();
+            $table->foreignId('to_user_id')->nullable()->index();
+            $table->string('subject', 80)->nullable();
+            $table->string('body', 50000);
+            $table->unsignedTinyInteger('flags');
+            $table->timestamps();
+            $table->softDeletes();
```
emmiegit, line 25: If we have soft deletes on for PMs, will there be a "trash" concept, or is it only soft deletes in the code but there's no practical way for users to recover them? (Or something else)

web/php/Actions/PMAction.php
```diff
@@ -1,326 +1,218 @@
...
+        $message = UserMessage::find($runData->get('message_id'));
 
-        $selected = $runData->getParameterList()->getParameterValue("selected");
-        $json = new JSONService(SERVICES_JSON_LOOSE_TYPE);
-        $selected = $json->decode($selected);
+        $nextMessage = UserMessage::drafts($runData->user())
+                ->whereTime('id', '<', $message->created_at)->first()
+            ??
+            UserMessage::drafts($runData->user())
+                ->whereTime('id', '>', $message->created_at)->first();
```
emmiegit, line 208: I feel like this is non-obvious _enough_ that it's it's worth making into an explicit if statement instead of null coalesce. Or leave a comment explaining why this is happening I guess.

Alternatively, make an inline helper function that takes an argument `'<'` or `'>'` and then call `helperNameHere('<') ?? helperNameHere('>') // Get the next message, or previous if it's the last one`

web/php/Modules/Account/PM/PMComposeModule.php
```diff
@@ -1,61 +1,77 @@
...
-            $subject = preg_replace("/^Re: /", '', $subject);
-            $runData->contextAdd("subject", "Re: ".$subject);
-        } elseif ($continueMessageId) {
-            $message = PrivateMessagePeer::instance()->selectByPrimaryKey($continueMessageId);
-
-            if ($message == null || $message->getFromUserId() != $user->id()) {
-                throw new ProcessException(_("Error getting orginal message."), "no_reply_message");
+
+            $subject = $message->subject;
+            $subject = preg_replace('/^Re: /', '', $subject);
```
emmiegit: I think this would be cleaner rather than a substitute then concat:
```php
if (!str_starts_with($subject, 'Re: ')) {
    $subject = 'Re: ' . $subject;
}
```

web/php/Modules/Account/PM/PMDraftsModule.php
```diff
@@ -1,55 +1,52 @@
...
 {
 
+    /**
+     * Build drafts view.
+     * @param $runData
+     */
     public function build($runData)
     {
+        /** Not sure if this is worth breaking out into a setting. */
+        $per_page_limit = 30;
```
emmiegit, line 24: Presumably later we'll have more of a user settings system and can do it then?

I think the main issue with wikidot limits is they're wholly arbitrary. People wouldn't mind fixed limits much if they were standard, as in they were always 10 for "big" things, 50 for "short" things (like messages or drafts), or something like that.

web/php/Modules/Account/PM/PMDraftsModule.php
```diff
@@ -1,55 +1,52 @@
...
-        // get inbox messages for the user
-        $c = new Criteria();
-        $c->add("from_user_id", $userId);
-        $c->add("flag", 2); // for inbox
-
-        // also count them all!
-        $co = PrivateMessagePeer::instance()->selectCount($c);
-
-        $c->addOrderDescending("message_id");
+        $drafts_count = UserMessage::drafts($runData->user())->orderBy('id','DESC')->count();
```
emmiegit: If we're just doing `COUNT(*)` then `ORDER BY` shouldn't matter at all right?

web/php/Modules/Account/PM/PMDraftsModule.php
```diff
@@ -1,55 +1,52 @@
...
 
-        $perPage = 30;
-        // limits...
-        $totalPages = ceil($co/$perPage);
-        if ($pageNo>$totalPages) {
-            $pageNo = $totalPages;
+        $total_pages  = ceil($drafts_count/$per_page_limit);
```
emmiegit:
```suggestion
        $total_pages  = ceil($drafts_count / $per_page_limit);
```

web/php/Modules/Account/PM/PMDraftsModule.php
```diff
@@ -1,55 +1,52 @@
...
+        if ($this_page > $total_pages) {
+            $this_page = $total_pages;
```
emmiegit: Maybe `$this_page = min($this_page, $total_pages);`?

web/php/Modules/Account/PM/PMInboxModule.php
```diff
@@ -1,55 +1,52 @@
...
+        $drafts_count = UserMessage::inbox($runData->user())->orderBy('id','DESC')->count();
 
-        $perPage = 30;
-        // limits...
-        $totalPages = ceil($co/$perPage);
-        if ($pageNo>$totalPages) {
-            $pageNo = $totalPages;
+        $total_pages  = ceil($drafts_count/$per_page_limit);
+        if ($this_page > $total_pages) {
+            $this_page = $total_pages;
```
emmiegit: Ditto.

Actually there's a lot of repeating logic here, at some point we should step back and think about how we can standardize pagination logic somewhere. (Also relevant to my point above about standard page limits).

web/php/Modules/Account/PM/PMSentModule.php
```diff
@@ -1,55 +1,52 @@
...
-        // get inbox messages for the user
-        $c = new Criteria();
-        $c->add("from_user_id", $userId);
-        $c->add("flag", 1); // for sent
-
-        // also count them all!
-        $co = PrivateMessagePeer::instance()->selectCount($c);
-
-        $c->addOrderDescending("message_id");
+        $drafts_count = UserMessage::sent($runData->user())->orderBy('id','DESC')->count();
```
emmiegit: Ditto re count

web/php/Modules/Account/PM/PMSentModule.php
```diff
@@ -1,55 +1,52 @@
...
 
-        $userId = $runData->getUserId();
-
-        $pl = $runData->getParameterList();
-
-        $pageNo = $pl->getParameterValue("page");
-        if ($pageNo == null || $pageNo<0) {
-            $pageNo = 1;
+        $this_page = $runData->get('page');
+        if ($this_page == null || $this_page < 0) {
```
emmiegit:
```suggestion
        if ($this_page === null || $this_page < 0) {
```

web/php/Utils/NotificationMaker.php
```diff
@@ -19,36 +20,33 @@ public static function instance()
...
-        $not->setType("new_private_message");
+        $not->setUserId($message->recipient->id);
+        $not->setType('new_private_message');
 
-        $body = 'You have a new private message in your <a href="'.GlobalProperties::$HTTP_SCHEMA . "://" . GlobalProperties::$URL_HOST . '/account:you/start/messages">Inbox</a>!<br/>';
-        $body .= "From: ".WDRenderUtils::renderUser($fromUser)."<br/>";
-        $body .= 'Subject: <a href="'.GlobalProperties::$HTTP_SCHEMA . "://" . GlobalProperties::$URL_HOST . '/account:you/start/messages/inboxmessage/'.$message->getMessageId().'">'.htmlspecialchars($message->getSubject()).'</a><br/>';
+        $body = 'You have a new private message in your <a href="'.GlobalProperties::$HTTP_SCHEMA . '://' . GlobalProperties::$URL_HOST . '/account:you/start/messages">Inbox</a>!<br/>';
+        $body .= 'From: ' .WDRenderUtils::renderUser($fromUser). '<br/>';
+        $body .= 'Subject: <a href="'.GlobalProperties::$HTTP_SCHEMA . '://' . GlobalProperties::$URL_HOST . '/account:you/start/messages/inboxmessage/'.$message->id.'">'.htmlspecialchars($message->subject).'</a><br/>';
```
emmiegit, line 37: Wikidot's inline HTML building makes me sad every time I see it.

emmiegit (approved):
:envelope:

web/app/Models/UserMessage.php
danieltharp: `$this->save()` would still be working with the original instance of the object even if `save()` had already been called. The `$this->replicate()` method constructs a new (and unsaved) instance.

web/app/Models/UserMessage.php
danieltharp: Probably not, I wanted the ones for `isRead` and `isUnread` because they'll have roughly equal utility and I think adding the negation in that way isn't as clean. I may have gone overboard for things like Archived and Sent.

web/app/Models/UserMessage.php
danieltharp: Will do in the cleanup commit.

web/app/Policies/UserMessagePolicy.php
danieltharp: This will be added here and everywhere else that calls for some variation of `$user->id == 1`.

web/database/migrations/2021_08_06_003806_create_user_messages_table.php
danieltharp: Yes, I didn't bother with helper methods for trashed items because they already exist on the parent class, which oddly enough is called `trashed()`.

web/php/Actions/PMAction.php
danieltharp: I'm going to settle for an explanatory comment here, I don't want to spend too much more time on Action classes than is really needed.

web/app/Models/UserMessage.php
emmiegit: But `$this->setFlag()` would mutate it, right? So when you `$this->save()` it should save the updated flag bitset.

(I feel like there's something about laravel models here I'm not understanding)

web/php/Modules/Account/PM/PMComposeModule.php
danieltharp: See, that would make sense, but that's not what this code segment does. It's basically "ensuring" that there's only one `Re: ` prepended to a message: It removes the original in line 45 and then concatenates "Re: " to the subject when it's done in line 49.

You might say "That's dumb, why are we removing the string and then adding it back again?" and I would say "I don't know why we're doing that."

emmiegit: Right, it ensures there's only one `Re: `, but we can do the same by only prepending it if it doesn't already have `Re: ` at the start.

danieltharp: This is changed in the next commit.

web/php/Modules/Account/PM/PMDraftsModule.php
danieltharp: Right, earlier in this code I was doing multiple things with this one var, I forgot to remove that clause. Fixed.

web/php/Modules/Account/PM/PMDraftsModule.php
danieltharp: All of the related comments here down have been addressed.

PR #366, opened 2021-08-16
[WJ-704] Allow specifying wikis in triple-bracket links

emmiegit:
Wikidot allows you to `[[include]]` pages off-site using `:wiki-name:page-name` syntax (e.g. `:scp-wiki:theme:black-highlighter-theme`). This PR extends triple-bracket link syntax to be able to do the same. That is, you can already do `[[[Page Name Here]]]` to link to `/page-name-here`, and now you can do `[[[:scp-wiki:SCP-1000]]]` to link to a page on a different wiki.

This adds `LinkLocation` as a data type which either allows freeform string URLs like before, or pointing to a specific page via `PageRef`. This also changes backlinks to store a list of `PageRef`s rather than string slugs. These two necessitated changes to the AST tests and to the PHP FFI interface.

### DIFF
--- a/ftml/Cargo.lock
+++ b/ftml/Cargo.lock
@@ -309,7 +309,7 @@ dependencies = [
 
 [[package]]
 name = "ftml"
-version = "1.0.0"
+version = "1.1.0"
 dependencies = [
  "built",
  "cbindgen",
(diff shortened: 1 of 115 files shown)


danieltharp (approved):
I'm good with this, as we keep on with legacy class removal, adding a helper to get the site by unix_name is doable, or you can skip the criteria-based approach with something like `$site = Illuminate\Support\Facades\DB::table('site')->where('unix_name', $name)->first();`

emmiegit:
Makes sense. I'm going to leave this as-is, since I want a lot of these ad hoc Wikidot-interface helper methods to go away as we more cleanly integrate ftml as a wikitext provider. (See WJ-780)

emmiegit:
thanks @pxdnbluesoul

PR #367, opened 2021-08-16
[WJ-771] Login and Registration

danieltharp:
This PR retires all Wikidot code related to authentication and registration. It utilizes Laravel Jetstream (a package including Fortify for authn, Sanctum for API key management, and includes a basic frontend utilizing Blade and Livewire.

It adds a number of new features and enhancements of existing ones. Users can enable multi-factor authentication with a TOTP provider of their choosing. User passwords are checked against compromised password lists and rejected if that password has been seen too many times in data leaks. API keys can be generated for a user (personal access tokens with perpetual life). User avatars can be stored, which is uploaded to Amazon S3. Users without avatars will have one generated for them instead of a single default avatar.

Previous modules, templates, and screens related to Login and Account Creation have been removed, and the seeded pages that instantiate the modules have been removed.

Other changes:

- The new user flow breaks the legacy implementation of Memcache, so all calls to the `Ozone::$memcache` object have been replaced with calls to the `Cache` facade, allowing for greater flexibility in cache providers.
- Legacy seed.sql scripts have been removed.
- The Wikidot default avatar image has been removed.
- Exploratory work was performed in breaking down the `RunData` class. It's currently unused but visible in the `web/app/Helpers/PageBuilder.php` file.


docs/development.md
```diff
@@ -118,6 +118,12 @@ $ docker images        # List images
 $ docker rmi [ID]      # Remove the image with this ID
 ```
 
+Note: If you enable two-factor authentication on a local container you may find that
+the clock drift is too great for TOTP codes to work. In docker-compose (`wsl -d docker-compose`) 
+you can enter this command to sync your time up:
+
+`ntpd -d -q -n -p 0.pool.ntp.org`
```
emmiegit: Use this, and it'll show up as a full code block (appropriate for commands), and GitHub also adds a convenient copy-to-clipboard button.
`````suggestion
```
ntpd -d -q -n -p 0.pool.ntp.org
```
`````

File: web/app/Actions/Fortify/CreateNewUser.php [hunk 2]
```diff
@@ -21,22 +27,21 @@ class CreateNewUser implements CreatesNewUsers
...
-            'email' => [
-                'required',
-                'string',
-                'email',
-                'max:255',
-                Rule::unique(User::class),
-            ],
+            'username' => ['required', 'string', 'max:255', 'unique:users'],
+            'unix_name' => ['unique:users'],
+            'email' => ['required', 'string', 'email', 'max:255', 'unique:users'],
```
emmiegit, line 37: I'm assuming something about this requires these strings to be non-empty?

web/app/Actions/Fortify/CreateNewUser.php
```diff
@@ -21,22 +27,21 @@ class CreateNewUser implements CreatesNewUsers
...
         Validator::make($input, [
-            'name' => ['required', 'string', 'max:255'],
-            'email' => [
-                'required',
-                'string',
-                'email',
-                'max:255',
-                Rule::unique(User::class),
-            ],
+            'username' => ['required', 'string', 'max:255', 'unique:users'],
```
emmiegit: I believe we mentioned a minimum of 3 characters, to avoid preventable fights over unrecognizable single or dual-character names (e.g. `N`).

web/app/Actions/Fortify/LoginResponse.php
```diff
@@ -0,0 +1,34 @@
...
+     */
+    public function toResponse($request): Response
+    {
+        if($request->wantsJson())
+        {
+            return response()->json(['two_factor' => false]);
+        }
+        return session('backUrl') ?
+            redirect()->to(session('backUrl'))
+            : redirect()->route('dashboard');
```
emmiegit, line 34: Nitpick, I think this would be more readable as
```php
$previousUrl = session('backUrl');
return $previousUrl ? redirect()->to($previousUrl) : redirect()->route('dashboard');
```

web/app/Actions/Fortify/UpdateUserProfileInformation.php
```diff
@@ -8,37 +9,43 @@
...
-            'email' => [
-                'required',
-                'string',
-                'email',
-                'max:255',
-                Rule::unique('users')->ignore($user->id),
-            ],
+            'name' => ['string', 'max:255'],
+            'email' => ['required', 'email', 'max:255', Rule::unique('users')->ignore($user->id)],
+            'photo' => ['nullable', 'mimes:jpg,jpeg,png', 'max:1024'],
```
emmiegit: We should also permit `webp`. I think excluding `svg` is wise given potential security considerations.

web/app/Actions/Fortify/UpdateUserProfileInformation.php
```diff
@@ -8,37 +9,43 @@
...
+ * Class UpdateUserProfileInformation
+ * @package Wikijump\Actions\Fortify
+ */
 class UpdateUserProfileInformation implements UpdatesUserProfileInformation
 {
     /**
      * Validate and update the given user's profile information.
      *
-     * @param mixed $user
+     * This will likely be where we need to add some logic around tracking username changes.
```
emmiegit: Nitpick: I would add `TODO` so we can find this later when we grep for `TODO`s.

web/app/Helpers/PageBuilder.php
```diff
@@ -0,0 +1,344 @@
...
+        $this->initParameterList();
+
+        /**
+         * Then we take our extracted parameters and choose the template and screen to use.
+         */
+        $this->setTemplate();
+
+        /** Set the language and theme for the request. */
+        $this->language = $this->parameterArray['lang'] ?? GlobalProperties::$DEFAULT_LANGUAGE;
+        if ($this->parameterArray['skin']) { $this->skin = $this->parameterArray['skin']; }
```
emmiegit:
```suggestion
        if ($this->parameterArray['skin']) {
            $this->skin = $this->parameterArray['skin'];
        }
```

web/app/Helpers/PageBuilder.php
```diff
@@ -0,0 +1,344 @@
...
+             *
+             * The user makes a request for a page like:
+             * `wikijump.test/testpage`
+             *
+             * But nginx transforms it and provides it to the server as:
+             * `wikijump.test/index.php?Wiki__WikiScreen/wiki_page/testpage`
+             *
+             * And so the query string looks like:
+             *          0           1         2
+             * `Wiki__WikiScreen/wiki_page/testpage`
```
emmiegit, line 172: Curious, how does this work with URL parameters (e.g. `/scp-001/noredirect/true/offset/7`)?

web/app/Helpers/PageBuilder.php
```diff
@@ -0,0 +1,344 @@
...
+                /** We put the set in the same parameter store we're already building. */
+                $this->parameterArray[$key] = urldecode($value);
+                $this->parameterTypes[$key] = 'GET';
+                $this->allParameters['GET'][$key] = urldecode($value);
+            }
+
+            /** Lastly, POST values don't need as much manipulation. */
+            foreach ($_POST as $key => $value) {
+                /** Convert any CRLF values to LF */
+                $value = str_replace(["\r\n", "\r", "\n"], "\n", $value);
```
emmiegit: Shouldn't this just be this, like the others?
```suggestion
                $value = str_replace(["\r\n", "\r"], "\n", $value);
```

web/app/Helpers/PageBuilder.php
```diff
@@ -0,0 +1,344 @@
...
+            /**
+             * If we're not making an AJAX call, we're looking for a Screen instead of a template.
+             * Screens are in php/Screens and are the classes that use the
+             * templates in templates/screens.
+             */
+            $template = $this->parameterArray['template'];
+
+            /**
+             * If we didn't get a valid value back (not empty, contains only
+             *  letters, numbers, underscores, and slashes), use the Index screen. (nothing)
```
emmiegit, line 261: The `Index` screen is "nothing"?

web/app/Helpers/PageBuilder.php
```diff
@@ -0,0 +1,344 @@
...
+            if ($template == null || preg_match('/^[a-z0-9_\/]+$/i', $template) != 1) {
+                $template = 'Index';
+            }
+
+            /**
+             * We receive the call on the php file as a path using two
+             * underscores instead of a slash.  We need to pass the actual path to
+             * our next set of tasks.
+             */
+            $template = str_replace('__', '/', $template);
```
emmiegit, line 272: This is also done for actions. Does it happen anywhere else? (If so, 3+ usages suggests it would be better as a helper function).

web/app/Helpers/PageBuilder.php
```diff
@@ -0,0 +1,344 @@
...
+
+    /**
+     * Parse and return the corresponding path for a template or screen.
+     * @param $template
+     * @return void
+     */
+    public function findModuleClass($template)
+    {
+        /**  Retrieve the path to the module. */
+        if ($this->moduleTemplate != null) {
```
emmiegit: Nitpick: All these `!= null`s should be `!== null`.

web/app/Helpers/PageBuilder.php
```diff
@@ -0,0 +1,344 @@
...
+                     */
+                    $path .= $module_path[$k] . '/';
+                    if ($this->moduleTemplate != null) {
+                        $path .= 'DefaultModule.php';
+                    }
+                    if ($this->screenTemplate != null) {
+                        $path .= 'DefaultScreen.php';
+                    }
+                }
+                $classFiles[] = $path;
```
emmiegit: I personally prefer `array_push()` notation since I think it's more clear / explicit.

web/app/Helpers/PageBuilder.php
```diff
@@ -0,0 +1,344 @@
...
+                for ($k = 0; $k < $i; $k++) {
+                    /**
+                     * Build a string with the most specific path to the
+                     * fallback (DefaultModule.php) file.
+                     */
+                    $path .= $module_path[$k] . '/';
+                    if ($this->moduleTemplate != null) {
+                        $path .= 'DefaultModule.php';
+                    }
+                    if ($this->screenTemplate != null) {
```
emmiegit: Nitpick: `elseif`?

web/app/Helpers/helpers.php
```diff
@@ -12,3 +12,262 @@ function pg_is_error(Throwable $e, string $code) : bool
...
+        [ 'b' => 'b0bec5', 't' => '000000', ],
+        [ 'b' => '90a4ae', 't' => '000000', ],
+        [ 'b' => '78909c', 't' => '000000', ],
+        [ 'b' => '9e9e9e', 't' => '000000', ],
+        [ 'b' => 'fafafa', 't' => '000000', ],
+        [ 'b' => 'f5f5f5', 't' => '000000', ],
+        [ 'b' => 'eeeeee', 't' => '000000', ],
+        [ 'b' => 'e0e0e0', 't' => '000000', ],
+        [ 'b' => 'bdbdbd', 't' => '000000', ],
+    ];
```
emmiegit, line 272: Curious, how did you come up with these?

web/app/Models/UserMessage.php
```diff
@@ -256,7 +256,7 @@ public function scopeStarred(Builder $query, User $user) : Builder
     public function scopeSent(Builder $query, User $user) : Builder
     {
         return $query->where('from_user_id', $user->id)
-            ->where('flags', '&', self::MESSAGE_SENT);
+            ->whereRaw('flags::int::bit(16) & ' . self::MESSAGE_SENT . '::bit(16) != 0::bit(16)');
```
emmiegit: Given we have to repeat this ugly raw query fragment for each method that uses bitflags, and that we may use bitflags on other tables in the future, do you think it would make sense creating query helpers for these? Get, set, and clear. Like `->whereRaw(BitFlags::check('flags', self::MESSAGE_SENT))`?

web/database/factories/UserFactory.php
```diff
@@ -41,4 +45,18 @@ public function definition()
...
+     * Indicate that the model's email address should be unverified.
+     *
+     * @return Factory
+     */
+    public function unverified() : Factory
+    {
+        return $this->state(function (array $attributes) {
+                return [
+                        'email_verified_at' => null,
+                    ];
```
emmiegit: Weird indentation.

web/lib/ozoneframework/php/core/RunData.php
```diff
@@ -597,9 +596,8 @@ public function handleSessionEnd() {
 					$session->setSessionChanged(false);
 				}
 
-				$mc = OZONE::$memcache;
 				$key = 'session..'.$session->getSessionId();
-				$mc->set($key, $session, 0, 600);
+				Cache::put($key, $session,600);
```
emmiegit:
```suggestion
				Cache::put($key, $session, 600);
```

web/package.json
```diff
@@ -10,7 +10,14 @@
         "build": "webpack --config conf/webpack.dev.js",
         "build-prod": "webpack --config conf/webpack.prod.js",
         "test": "jest",
-        "test-coverage": "jest --coverage"
+        "test-coverage": "jest --coverage",
```
emmiegit, line 13: Add a space so we know which were added by tailwind.
```suggestion

        "test-coverage": "jest --coverage",
```

web/php/Screens/Feed/FrontForumFeed.php
```diff
@@ -126,38 +126,38 @@ public function build($runData)
         $categories = array();
 
         if (count($cats)>20) {
-            throw new ProcessException(_("Maximum number of categories exceeded."), "max_categories");
+            throw new ProcessException(_('Maximum number of categories exceeded.'), 'max_categories');
```
emmiegit: I just tested this, and Wikidot no longer as a maximum category limit (or that limit is well above 20). I created an issue to get rid of this: [WJ-784](https://scuttle.atlassian.net/browse/WJ-784)

web/php/Utils/WDPermissionManager.php
```diff
@@ -133,6 +134,7 @@ public function __construct()
 
     public function hasPermission($action, $user, $site = null)
     {
+
```
emmiegit: ?

emmiegit (approved):

web/app/Actions/Fortify/CreateNewUser.php
danieltharp: The `required` property implies not null, and unix_name is added automatically as part of the creation, but must be a unique value in the table. We'll have to figure out how we want that exception to be presented to the user better than the hysterically wishy-washy "It appears someone may have a name that's somewhat similar to yours!"

web/app/Actions/Fortify/UpdateUserProfileInformation.php
danieltharp: You know, I didn't even look at these. I'll add that, it's growing more ubiquitous.

web/app/Helpers/PageBuilder.php
danieltharp: After the 0th argument, it takes sets of two as key-value pairs which is how it knows what to do, and it's why `/norender` by itself doesn't do anything.

emmiegit: (Unrelated to this PR) I should see if we can be more flexible with this routing. A while back I had a system properly parse keys/values (including optional values) called [`wikidot-path`](https://github.com/scpwiki/wikidot-path).

danieltharp: Routing gets a tricky with optional arguments, and particularly when the order is variable. You can look at https://stackoverflow.com/a/22331171 for what's still really the state of things.

Now, we could probably emulate the existing method by doing something like:
`Route::get('/{page}/{argument1?}/{value1?}/{argument2?}/{value2?}/{argument3?}/{value3?}`

Something along those lines. Then we have parse the arguments and their respective values within the route closure and pass them to the Page Controller with the appropriate parameters.

emmiegit: Is it possible to have `Route::get('/{slug}/{params!}')` or something where `{params}` consumes the rest of the URL regardless of slashes? Then we could parse out the parameters and hand the context object to the PageController.

emmiegit: Looks like it's possible with `.*`? (Which is similar to how some other web framework I remember does it) https://stackoverflow.com/a/21555108

danieltharp: Actually yes, I think so. https://laravel.com/docs/8.x/routing#parameters-encoded-forward-slashes gives an example. You capture the first argument and then the second can be a `.*` regular expression, or we could scope it to `[a-zA-Z0-9\/]` or some such.

emmiegit: Very cool. I'll give this a spin sometime for WJ-51.

web/app/Actions/Fortify/CreateNewUser.php
danieltharp: Ah yeah, this was a validation step we added in the previous Action class, I'll update here.

web/app/Helpers/PageBuilder.php
danieltharp: Indeed. It's a pointer to a file that doesn't exist. It should be at `web/templates/screens/IndexScreen.tpl` but the file isn't there at all.

https://github.com/scpwiki/wikijump/blob/WJ-771-login-and-registration/web/lib/ozoneframework/php/core/RunData.php#L256

The equivalent module template is `Empty`:

https://github.com/scpwiki/wikijump/blob/WJ-771-login-and-registration/web/templates/modules/Empty.tpl

web/app/Helpers/helpers.php
danieltharp: https://github.com/LasseRafn/ui-avatars/blob/master/Utils/Input.php#L71

web/app/Models/UserMessage.php
danieltharp: This is a good idea and I've added `flagIsSet(int $flag)` and `flagIsNotSet(int $flag)` to the `UsesBitmasks` trait to make this happen.

web/lib/ozoneframework/php/core/RunData.php
danieltharp: Sorry, these are hard to tell sometimes in the IDE as it inserts the arg name and it looks like there's a space there.

PR #371, opened 2021-08-17
[WJ-508] Final removal of Profile code

danieltharp:
There was one remaining template and page (and an orphaned metadata entry) involving the Profile site, this PR removes them.


emmiegit (approved):
:wastebasket:

PR #372, opened 2021-08-18
[WJ-785] [WJ-786] Use a newline context flag instead of line-lookahead

emmiegit:
This PR modifies the `Parser` structure to contain a flag that tells it whether it was previously on a newline. This is useful to minimize the complexity of rules which require they start at the beginning of a line.

However we cannot fully remove lookahead-like behavior for newlines. On the line break rule, we check if the upcoming tokens look like they're going to be line-starters, and emit null if so. And we need to retain the block-lookahead rule for ignoring newlines if the upcoming block is of a type that is newline-separated.

It also adds a script to assist with comparing the results of failing AST tests, and removes `check_step_multiple()` as it is now unused.


danieltharp (approved):

emmiegit:
thanks @pxdnbluesoul

PR #375, opened 2021-08-19
[WJ-777] Add PageInfo to parsing as well

emmiegit:
Previously, only the render step required `PageInfo` information. Since parsing and rendering are performed together in the FFI anyways, and we need that information for table of contents generation, I've modified the parser to also require a `PageInfo` like the renderer.


danieltharp (approved):

emmiegit:
thanks @pxdnbluesoul

PR #376, opened 2021-08-19
[WJ-605] [WJ-778] Add simple table parsing, and attribute macro

emmiegit:
This PR adds support for Wikidot's "simple tables". [Wikidot documentation here](https://www.wikidot.com/doc-wiki-syntax:tables). These are the inline form you see most often:
```
||~ Name ||~ Location ||
|| Jim || USA ||
|| Hans || Germany ||
```

Additionally, it adds an `attr!()` macro to consolidate the various uses of attribute methods that have gotten all over the place. This resulted in some non-table AST tests needing to be updated, as well as existing HTML rendering code.

It also removes `RULE_TODO` as it is now unused.


ftml/src/render/html/element/table_of_contents.rs
```diff
@@ -1,77 +0,0 @@
-/*
```
emmiegit, line 1: This file is renamed to `toc.rs`, not deleted.

danieltharp (approved):

emmiegit:
thanks @pxdnbluesoul

PR #379, opened 2021-08-19
Update ForbiddenNames.php

bauerbach:
I think that the usernames of people permabanned from the current site should be included on here unless they appeal. If by any chance it becomes impossible to appeal by the usual means of appealing bans from the site for some reason, then an admin should have an alternate way of doing appeals.

Do you think forbidding new registrations of users permabanned from the current site is a good idea? Or is it only a good idea if they are unable to appeal?

To look at more info on why a user was banned, Google "05command disciplinary [username]", where [username] is replaced with the actual usernazme. Their 05 record (Disciplinary thread) should be the first result.

Also, you can search "+"perma" site:05command.wikidot.com" on Google to quickly see who has been permabanned.

### DIFF
--- a/web/php/Config/ForbiddenNames.php
+++ b/web/php/Config/ForbiddenNames.php
@@ -93,6 +93,13 @@ class ForbiddenNames {
         '/^robot$/',
         '/^O5-\w+$/',
         '/^SCP-\w+$/',
+'/^bauerbach\w+$/',
+      '/^brandonfave\w+$/',  
+'/^bradleya\w+$/',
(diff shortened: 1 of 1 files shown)


emmiegit:
This list is for names which are not appropriate for sites or users to have, due to the possibility of deception. For instance somebody shouldn't be able to make an account "wikijump-official" when they are, in fact, not an official Wikijump account.

For bans of regular users, those are stored in the bans table in the database. They should not be saved in a text configuration file.

PR #380, opened 2021-08-19
Looser acceptance for emails

emmiegit:
I'm not sure why I did a naive pattern for emails. It excludes plenty of common valid ones. This pattern is much more generous and simply assumes there is a `@` and then a `.` somewhere in the string.


danieltharp (approved):

emmiegit:
thanks @pxdnbluesoul

PR #382, opened 2021-08-27
[WJ-783] Datadog and S3 optimizations

danieltharp:
This pulls several necessary pieces of infrastructure changes into develop. There is some stuff still to go but they're standalone pieces to be done in a future PR.

### DIFF
--- a/install/aws/dev/docker/php-fpm/Dockerfile
+++ b/install/aws/dev/docker/php-fpm/Dockerfile
@@ -68,6 +68,7 @@ RUN apt install -y \
     libmemcached-dev \
     postgresql-common \
     libtidy-dev \
+    chrony \
     gettext && \
 
     # Memcached PHP lib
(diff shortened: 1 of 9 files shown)


install/aws/dev/docker/php-fpm/Dockerfile
```diff
@@ -95,6 +96,8 @@ COPY wikijump.ini conf/wikijump.ini
 RUN sed -i "s/BASEDOMAIN/${MAIN_DOMAIN}/g" conf/wikijump.ini && \
     sed -i "s/MAINWIKI/${WWW_DOMAIN}/g" conf/wikijump.ini && \
     sed -i "s/FILEDOMAIN/${FILES_DOMAIN}/g" conf/wikijump.ini && \
+    #  Set time servers to AWS NTP provider
+    sed -i '1s/^/server 169.254.169.123 prefer iburst minpoll 4 maxpoll 4\n/' /etc/chrony/chrony.conf
```
emmiegit, line 100: (Assuming chrony supports domains), is there a domain name version for AWS NTP or is it IP only? Is there an IPv6 version?

install/aws/dev/terraform/deploy/s3.tf
```diff
@@ -15,20 +15,56 @@ resource "aws_s3_bucket" "wikijump_assets" {
...
 
     expiration {
-      days = 1
+      days                         = 1
       expired_object_delete_marker = false
     }
   }
 }
+
+data "aws_caller_identity" "current" {}
```
emmiegit, line 36: I'm curious, if this is empty, what does this do?

emmiegit (approved):

install/aws/dev/docker/php-fpm/Dockerfile
danieltharp: No, reason being link-local addresses won't have DNS availability. But this is an address that they will never get rid of, too many resources including their own rely on it.

install/aws/dev/terraform/deploy/s3.tf
danieltharp: This is the prescribed way to retrieve the AWS account number for Terraform's IAM user that's running the workload.

PR #385, opened 2021-09-05
[WJ-604] [WJ-195] Add advanced tables

emmiegit:
Similar to #376, this PR adds more table syntax. Wikidot's "advanced tables" utilize its blocks for customized table structure, i.e. using `[[table]]`, `[[row]]`, `[[cell]]`, etc.

This also fixes WJ-195. Because matching Wikidot's behavior here would require invasive changes to the block parser, and is weird and undesirable behavior anyways, it was not implemented, and thus the bug is fixed.

As always the PR adds AST tests for the new syntax. It also adds a script to create new AST test files easier. Because new blocks are added, the block configuration and documentation are also updated.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #386, opened 2021-09-06
WJ-685 & WJ-66 (Partially) — Repurposing Site Tags as a Tag Restriction System

Yossipossi1:
This pull request is designed to repurpose the deprecated "Site Tags" feature for an ability to restrict tags to a specific list, as detailed in [WJ-66](https://scuttle.atlassian.net/browse/WJ-66) and [WJ-685](https://scuttle.atlassian.net/browse/WJ-685). This option is toggle-able in the general site settings. Most of the code is in Ozone, though some of the it is in Laravel. 

Obviously this is not the end of working on tags, but it's a good first step to get the framework in place for future development. I plan on revisiting this in the future to convert all code involving tags to Laravel.


web/php/Actions/ManageSiteAction.php
```diff
@@ -438,8 +438,9 @@ public function saveGeneralEvent($runData)
         $pl = $runData->getParameterList();
         $name = trim($pl->getParameterValue("name"));
         $subtitle = trim($pl->getParameterValue("subtitle"));
-
         $description = trim($pl->getParameterValue("description"));
+        $enableAllowedTags = trim($pl->getParameterValue("enable_allowed_tags"));
+        $enableAllowedTags = !empty($enableAllowedTags) ? 'true' : 'false';
```
emmiegit: Why are these strings instead of booleans?

web/php/DB/AllowedTags.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    public static function getAllowedTags($siteId): string {
+        $c = new Criteria();
+        $c->add("site_id", $siteId);
+        $dbTags = AllowedTagsPeer::instance()->selectByCriteria($c);
+        $allowedTags = '';
+        foreach ($dbTags as $dbTag) {
+            $allowedTags .= htmlspecialchars($dbTag->getTag()).' ';
+        }
+        return trim($allowedTags);
```
emmiegit: Why is this using space-separated tags in a string instead of `array<string>`? In the uses I can see you immediately `explode()`, so I don't think it makes sense to store it this way.

web/php/Actions/WikiPageAction.php
```diff
@@ -1407,8 +1414,17 @@ public function saveTagsEvent($runData)
...
-            throw new ProcessException(_('"Tags" field too long.'), "form_error");
+        $pageTagsArray = explode(' ', $tags);
+
+        $allowedTagsList = AllowedTags::getAllowedTags($siteId);
+        $allowedTagsList = explode(' ', $allowedTagsList);
+
+       if($enableAllowedTags == true) {
+            foreach ($pageTagsArray as $tag) {
+                if(!in_array($tag, $allowedTagsList)) {
+                    throw new ProcessException(_('Invalid tags.'), "form_error");
```
emmiegit: I would give it a more specific error message, like "Tag XYZ is not valid for this site."

web/php/Actions/ManageSiteAction.php
```diff
@@ -438,8 +438,9 @@ public function saveGeneralEvent($runData)
         $pl = $runData->getParameterList();
         $name = trim($pl->getParameterValue("name"));
         $subtitle = trim($pl->getParameterValue("subtitle"));
-
         $description = trim($pl->getParameterValue("description"));
+        $enableAllowedTags = $pl->getParameterValue("enable_allowed_tags");
+        $enableAllowedTags = !empty($enableAllowedTags) ? 'true' : 'false';
```
emmiegit: Since this is intentional (for now), I would leave a comment explaining why these are strings.

web/php/Actions/WikiPageAction.php
```diff
@@ -1407,8 +1414,17 @@ public function saveTagsEvent($runData)
 
         WDPermissionManager::instance()->hasPagePermission('edit', $user, $category, $page);
 
-        if (strlen8($tags)>256) {
-            throw new ProcessException(_('"Tags" field too long.'), "form_error");
+        $pageTagsArray = explode(' ', $tags);
+
+        $allowedTagsList = AllowedTags::getAllowedTags($siteId);
+
+       if($enableAllowedTags == true && !empty($tags)) {
```
emmiegit: Either
```suggestion
       if($enableAllowedTags && !empty($tags)) {
```
or
```suggestion
       if($enableAllowedTags === true && !empty($tags)) {
```

emmiegit (approved):

PR #388, opened 2021-09-07
[WJ-792] Update normalization to support international letters

emmiegit:
This PR adjusts `toUnixName()` to support non-ASCII characters. It performs NFKC normalization and lowercase folding, and then strips out any non-letter or non-digit characters (based on unicode properties). This way languages with non-latin scripts can make have native characters in their normalized URLs (e.g. CJK languages).

The `intl` extension for PHP is added to make use of its unicode library functions, and some documentation is added.

image

See also https://github.com/scpwiki/wikidot-normalize/pull/1


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #389, opened 2021-09-07
[WJ-792] Upgrade ftml to use new wikidot-normalize version

emmiegit:
This upgrades the `wikidot-normalize` dependency to use the version which has changes from https://github.com/scpwiki/wikidot-normalize/pull/1.

PR #390, opened 2021-09-07
Fix unicode underscore bug in path normalization

emmiegit:
The failing test in #389 was actually a result of a bug in [`wikidot-normalize`](https://github.com/scpwiki/wikidot-normalize), because it used the wrong method for getting character byte boundaries. This was fixed in 0.9.1 of that library, and a test was added to catch regressions.

This PR upgrades that dependency and bumps the patch version number to mark it.

Thanks to @leo60228 for catching this.

### DIFF
--- a/ftml/Cargo.lock
+++ b/ftml/Cargo.lock
@@ -303,7 +303,7 @@ dependencies = [
 
 [[package]]
 name = "ftml"
-version = "1.5.1"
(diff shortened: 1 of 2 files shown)


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #391, opened 2021-09-07
[WJ-797] Redirect non-normalized URLs

emmiegit:
If you were to visit a non-normal URL on mainline wikidot, such as `/SCP 001`, it will redirect you to `/scp-001`. This behavior was not mimicked by our version of wikidot, so this PR adds this redirection. To help achieve this, it also extracts the raw parameter list from the request URI (e.g. `/noredirect/true`) so the URL matches.

In Laravel this should be a cleaner redirect function, that makes use of a properly-parsed parameter list. But for now this is fine.

I tested to ensure this works with our new international normalization as well. Here I visited `/SCP--五`:
image

I also confirmed it preserves URL parameters as expected.

### DIFF
--- a/web/app/Helpers/LegacyTools.php
+++ b/web/app/Helpers/LegacyTools.php
@@ -192,25 +192,13 @@ class LegacyTools
         }
 
         // Gets parameters (e.g. /noredirect/true), if any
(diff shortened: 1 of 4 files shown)


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #392, opened 2021-09-08
[WJ-51] Allow empty values in URL parameters

emmiegit:
As described in [WJ-51](https://scuttle.atlassian.net/browse/WJ-51), it's always been awkward that you need to say `/some-page/noredirect/true` for the redirect to not take effect, even though `/some-page/noredirect` should seem like it stops redirection. It's worse because `/some-page/noredirect/false` is the same as if it were `true`.

This changes this, with smarter parsing of parameter keys and values based on [`wikidot-path`](https://github.com/scpwiki/wikidot-path). It permits empty values, and actually casts the type based on its value, such as producing a boolean `true`/`false` value if appropriate.

In the future we may want a nicer mapping of permitted keys and what types they want, but for now this is a nice improvement.

Test visiting `https://www.wikijump.test/redir/noredirect`:
image

### DIFF
--- a/web/lib/ozoneframework/php/core/ParameterList.php
+++ b/web/lib/ozoneframework/php/core/ParameterList.php
@@ -90,12 +90,25 @@ class ParameterList {
 		return !array_search($name, $this->parameterArray) === false;
 	}
 
-	public function getParameterValue($name, $type = null, $type2 = null) {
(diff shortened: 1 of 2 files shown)


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #393, opened 2021-09-08
Tweak redirect module message

emmiegit:
The message currently is `This is the Redirect module that redirects the browser directly to the "https://example.com" page.` which is kind of stilted and awkward.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #394, opened 2021-09-08
[WJ-793] Add footnotes

emmiegit:
This PR adds footnotes to ftml. `[[footnote]]` permits specifying a reference, and `[[footnoteblock]]` allows rendering the footnote block in a particular place of the user's choosing.

Multiple `[[footnoteblock]]`s are permitted. Nested footnotes are not, this yields an error.

Because an implicit `[[footnoteblock]]` is now added at the end of every syntax tree, this PR had to edit all AST test files. As always, it also adds new AST tests for the syntax.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #395, opened 2021-09-09
[WJ-488] Don't use <tt> for monospace container

emmiegit:
This replaces the use of `<tt>` for HTML rendering for `ContainerType::Monospace` with `<code class="wj-monospace">`. This was done because [`<tt>` is deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tt).


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #397, opened 2021-09-09
[WJ-807] Configure dependabot for automatic package updates

emmiegit:
We previously didn't have a `dependabot.yaml`, so WJ-807 asks that we make one.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -10,7 +10,7 @@ updates:
 
   # Frontend
   - package-ecosystem: npm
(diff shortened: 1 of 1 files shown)


Monkatraz (approved):
FYI: we may want to tone this down for the client monorepo if it gets too annoying lol

emmiegit:
Yeah there might be too many updates. We'll see, we can easily turn this to `weekly`.

emmiegit:
thanks @Monkatraz

PR #410, opened 2021-09-09
Make all vendored files binary

emmiegit:
This centralizes the `.gitattributes` files so that everything in `vendor/` directories are considered binary.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #413, opened 2021-09-09
[WJ-805] Handle [[iftags]] and [[ifcategory]] exclusion in the parser

emmiegit:
Because we have `PageInfo` data exposed during parsing now, we can just check then whether the conditions are valid or not. Then we include / exclude the elements as appropriate.

This also adds logic to ensure that styling inside an `[[iftags]]` or `[[ifcategory]]` aren't added, as well as tests to verify this.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #415, opened 2021-09-10
Switch client monorepo tests to being ran in a browser

Monkatraz:
This PR is a massive relief. This simplifies the monorepo and improves workflow massively. Tests are now ran in a headless browser, specifically Playwright, with the test runner being `@web/test-runner`. It uses Vite as a middleware, which means that Vite builds every test. This means that for every bit of module code in the monorepo, you can assume that Vite will be building it, and that it will be ran solely in a browser.

Involved with this change was switching away from `uvu` to Mocha, a very popular JS test framework. I've switched all the tests over to using it.

Finally, the last detail to note is that code coverage is a bit... interesting. I've got it to a satisfactory state, but it's going to report a much higher "baseline" code coverage for the entire monorepo now. The instrumentation system that `@web/test-runner` is Chrome's built in instrumenter, which means it's very accurate - but unfortunately, it counts literally every single line in a file during code coverage, and not just statements or whatever. That means blank lines and the like get counted as "covered".


.github/workflows/client.yaml
```diff
@@ -29,6 +29,9 @@ jobs:
         with:
           node-version: "16"
 
+      - name: Install Playwright Dependencies
+        run: npx playwright install-deps chromium
```
emmiegit, line 33: Where is Chromium stored? Is it kept in the PNPM cache (above)?

client/README.md
```diff
@@ -10,7 +10,7 @@
 ### Relevant Documentation
 
 * [PNPM](https://pnpm.io/)
-* [uvu](https://github.com/lukeed/uvu)
+* [Mocha](https://mochajs.org)
```
emmiegit, line 13: Curious, what was wrong with uvu? I don't understand the context for a lot of JS tool changes so it's not obvious to me.

client/modules/ftml-wasm/tests/ftml-original-tests.ts
```diff
@@ -1,12 +1,10 @@
...
-import * as uvu from "uvu"
-import * as assert from "uvu/assert"
+import { assert } from "@esm-bundle/chai"
 import * as lib from "../src/index"
-import type { IPageInfo, IParseWarning, ISyntaxTree } from "../vendor/ftml"
+import type { IPageInfo } from "../vendor/ftml"
+
+const jsons = import.meta.globEager("/../ftml/test/*.json")
+const htmls = import.meta.globEager("/../ftml/test/*.html?raw")
+const txts = import.meta.globEager("/../ftml/test/*.txt?raw")
```
emmiegit, line 7: What does it do ro these files without raw?

client/modules/ftml-wasm/tests/ftml.ts
```diff
@@ -1,191 +1,184 @@
...
+                  "element": "container",
+                  "data": {
+                    "type": "italics",
+                    "elements": [
+                      {
+                        "element": "text",
+                        "data": "1"
+                      }
+                    ],
+                    "attributes": {}
```
emmiegit: nitpick: attributes should be before elements

client/modules/ftml-wasm/tests/ftml.ts
```diff
@@ -1,191 +1,184 @@
...
+    //       tag_type: "http-equiv",
+    //       name: "Content-Type",
+    //       value: "text/html"
+    //     },
+    //     { tag_type: "name", name: "generator", value: "ftml 0.7.0" },
+    //     { tag_type: "name", name: "description", value: "" },
+    //     { tag_type: "name", name: "keywords", value: "" }
+    //   ],
+    //   style: ""
+    // })
```
emmiegit: Remove old commented code?

emmiegit (approved):

.github/workflows/client.yaml
Monkatraz: This installs the Chromium dependencies through the normal apt-get-whatever package manager, so it's unfortunately slow and is redone everytime. I don't know if there is a way to avoid this.

client/README.md
Monkatraz: Nothing, really. The reason Mocha is used is because the test runner uses it by default, and is well integrated with it. It also has some advantages when being run in a browser, I think.

.github/workflows/client.yaml
emmiegit: Whatever `/usr` directory it's installed to could be cached. Or actually probably you should cache the APT package cache dir.

client/modules/ftml-wasm/tests/ftml-original-tests.ts
Monkatraz: Tries to load them as JavaScript. Due to old web shenanigans, the file type of referenced file in a URL technically doesn't matter, so Vite follows this and doesn't assume the content type of a file, unless you tell it to.

client/modules/ftml-wasm/tests/ftml.ts
Monkatraz: iirc I left this here to show what testing for the meta field looked like, but I can probably just remove this.

client/modules/ftml-wasm/tests/ftml.ts
Monkatraz: fool. it passes tests. (I'll fix it)

although I'm curious if property order actually survives through the WASM boundary, or if property order is even a thing in Rust/WASM

client/modules/ftml-wasm/tests/ftml.ts
emmiegit: I think a more intentional comment would be better at that.

client/modules/ftml-wasm/tests/ftml.ts
emmiegit: Since `HashMap` doesn't have a notion of key ordering, I highly doubt serde is able to preserve that information.

.github/workflows/client.yaml
Monkatraz: Okay, I gave it a shot. I didn't try directly caching anything, because when I did some googling all I found was discouraging results. It seems caching installs like this is ill-advised. I tried using a container someone had made for Playwright, hoping that GitHub would maybe do some persistent Docker magic on it or something, but it was too messy and I don't think it worked anyways.

If you wanna give this a shot, go ahead, but for right now it's okay as it is.

emmiegit: Oh well, it was worth looking into at least.

Monkatraz:
thanks @ammongit

PR #416, opened 2021-09-11
[WJ-603] Messy includes

emmiegit:
Wikidot's include solution sucks, in several different respects. This renames the psuedoblock `[[include]]` to `[[include-messy]]` to designate clearly that it should not be preferred, and fixes up some aspects of its current searching / parsing solution.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #417, opened 2021-09-12
Add arguments to [[li]]

emmiegit:
Working on fixing slog-mock has revealed an issue with the `[[li]]` block. It does not accept arguments or end the head (even though it should), and was also incorrect in how list block items should be parsed out.

Because the structure of `ListItem::Elements` changed to accept attributes, the AST test files had to be updated.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #418, opened 2021-09-12
Replace slog-mock with full dummy macros

emmiegit:
The now-decommissioned [`slog-mock`](https://github.com/ammongit/slog-mock) crate would take in slog macros and then produce code which essentially took every argument and did `let _ = <item>;` to it. This would prevent unused value warnings when compiling without logging.

However, the compiler does not actually remove that code, it still runs it (since there could potentially be side effects), so it wasn't compiling out all formatting and debug code like expected. This PR replaces all the slog macros with empty ones which truly have no output. This means that all values used in logging were unused, which this PR addresses by marking them as such for both variants (i.e. making them `_variable`s).

With this merged, all logging code should truly be removed when the `log` feature is disabled. This should have performance benefits, especially for ftml.

There were also unexpected discoveries, such as a revealing a bug which was fixed in #417.

### DIFF
--- a/ftml/Cargo.lock
+++ b/ftml/Cargo.lock
@@ -324,7 +324,6 @@ dependencies = [
  "serde_repr",
  "slog",
  "slog-bunyan",
- "slog-mock",
  "sloggers",
  "str-macro",
  "strum",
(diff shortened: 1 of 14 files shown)


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #419, opened 2021-09-12
Monorepo mayhem (workers, dependencies, more)

Monkatraz:
This PR is a mess. It was preemptively named `worker-mayhem`, but it got out of hand quickly and became more than just workers. In general, this PR is about cleaning up certain sticky bits of the monorepo, specifically anything in relation to dependencies and building. There are a few other changes that are seemingly unrelated, but I was fixing bugs as they cropped up to prevent regressions.

This PR includes:
* Update all the dependencies, and fix CodeMirror's interleaved dependencies having a fit
* An upgrade to `ftml-wasm`s build/compile process. It doesn't use `wasm-pack` now, which makes the scripts a bit messy, but this comes with the benefit of fixing a bunch of annoying issues because `wasm-pack` doesn't let you configure `wasm-bindgen` very much.
* I only repacked `ftml-wasm` like, 900 times?
* Additional tweaks to increase `ftml-wasm` performance
* Vendored `indent.js` because WORKERS SUCK
* Removed a bunch of dependencies that are now no longer needed
* Simplified ESLint configuration
* Added a `languageList` facet so that there doesn't need to be a single package exporting every language
* Tweaked TypeScript, now every package has a `tsconfig.json`. This is all set up in such a way that the TypeScript language server doesn't bog the hell down on every package now, especially in Svelte.
* Tweaked the monorepo commands/scripts
* More stuff I probably don't remember


Monkatraz:
Oh, I forgot to mention in the description: you can't develop on Firefox anymore. Now now, I know how terrible that is, but blame Firefox for not implementing worker modules for 6 years now. I'm going to probably see if I can make a PR for Vite that allows bundling workers even in dev mode, but it's very tricky to handle bundles in development because Vite is literally designed not to be bundling in development.

emmiegit:
But the production version of the code still works on Firefox right?

client/.gitattributes
```diff
@@ -1,3 +1,3 @@
-pnpm-lock.yaml         binary
-vendor/                binary
+pnpm-lock.yaml         linguist-vendored linguist-generated
+**/vendor/**           linguist-vendored linguist-generated
```
emmiegit: You can add `binary` in addition to vendored/generated, since that impacts diffs.

Monkatraz:
Yes, thankfully. If I could break production code on a major browser, it would be Safari.

emmiegit (approved):

Monkatraz:
thanks @ammongit

PR #421, opened 2021-09-12
[WJ-602] Add include (elements)

emmiegit:
This adds a block `[[include-elements]]`, which unlike Wikidot `[[include]]` (called `[[include-messy]]` here), is processed like a regular block, and injects parsed elements from another page into the current page. A future set of changes will support variables for both kinds of includes.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #422, opened 2021-09-12
[WJ-803] Update logging levels

emmiegit:
Previously I had a strange paradigm where I wouldn't use `warn!` or `error!` for internal parser issues, because technically it was valid wikitext. However this makes logging kind of hard to read since everything is `debug!` with a few `info!`s. This changes logging to more readily use higher levels. Any "top-level" execution (including rules, blocks, and main functions) use `info!` to declare start instead of `debug!`. It also reverts `slog` to use the original level (i.e. excludes `trace!()`).

### DIFF
--- a/ftml/Cargo.lock
+++ b/ftml/Cargo.lock
@@ -303,7 +303,7 @@ dependencies = [
 
 [[package]]
 name = "ftml"
-version = "1.6.0"
+version = "1.6.1"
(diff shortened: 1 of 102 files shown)


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #424, opened 2021-09-13
[WJ-813] Add "is" attribute to ftml-generated HTML

emmiegit:
The [`is` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/is) is usable to help shape behavior on elements, and in the future we may use it for element hydration. This PR adds the attribute to currently-existing HTML element rendering.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #428, opened 2021-09-13
Get ftml-wasm-worker and deps to publish nicely, beautify logging

Monkatraz:
This PR gets `ftml-wasm-worker` and it's dependencies published on NPM without issues. It also adds some pretty colors to the logging, which I originally added so that errors could be seen easier. Additionally, the `publish` script now asks if you're sure you want to publish.

image
image

### DIFF
--- a/client/scripts/publish.js
+++ b/client/scripts/publish.js
@@ -1,9 +1,28 @@
-const { execSync } = require("child_process")
+const { section, linebreak, info, question, cmd } = require("./pretty-logs.js")
 
 const arg = process.argv[2]
 
-if (arg) {
-  execSync(`pnpm pack-module -- ${arg}`, { stdio: "inherit" })
-  process.chdir(`modules/${arg}/dist`)
-  execSync(`npm publish --access public`, { stdio: "inherit" })
-}
+;(async () => {
+  if (arg) {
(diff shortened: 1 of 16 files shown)


client/scripts/publish.js
```diff
@@ -1,9 +1,28 @@
...
-  execSync(`npm publish --access public`, { stdio: "inherit" })
-}
+;(async () => {
+  if (arg) {
+    cmd(`pnpm pack-module -- ${arg}`)
+
+    section("PUBLISH")
+    linebreak()
+
+    const answer = await question("Do you want to publish this package? [y/n] -> ")
```
emmiegit, line 12: Change to `y/N` to show no is default.

emmiegit (approved):

client/scripts/publish.js
Monkatraz: Will get fixed in the next PR (this file is getting deleted already lmao). Separate script will publish all packages at once because if the versioning/linking between packages gets messed up it will cause chaos

Monkatraz:
thanks @ammongit 

The CI is failing but I guess that's because I forgot to update the lockfile? It'll get resolved in the next PR.

PR #430, opened 2021-09-13
Update CodeMirror, FTML in-editor block & warning documentation

Monkatraz:
This PR generally gets everything up-to-date with the newer FTML stuff. Also updates CodeMirror.


emmiegit (commented):
No `pig.yaml` changes?

emmiegit:
You'll need to rebase after https://github.com/scpwiki/wikijump/pull/431 is merged, and update the example code to use it as well.

Monkatraz:
I added the `pig.yaml` changes. I'm not going to do it again, it's frustrating and tedious. I've been thinking about it and there is little that the pig-latin translation does for me. 

It does not help me find untranslated strings, because I have easier and simpler ways to do that. If I literally disconnect evry translation table from my client, I'll get default strings like `cmftml.blocks.toc.TITLE` which _very obviously_ indicates that it has a specifier and where that specifier is. If something is untranslated it'll just show up as English, which is wrong. I also have automated tools in my editor to go and find untranslated strings in markup.

It also doesn't help me test formatting, localization of dates/numbers, alternative fonts, weird characters, Unicode, nothing. It's literally just mangled English. I find it much more useful to switch to French, Ukrainian, etc. because that actually stresses the intl. library.

If we keep pig-latin, it's getting automated. Even then, I'm not going to be using it.

emmiegit:
Honestly yeah, having added to it twice I've thought it about that as well. And given a "proper" translation requires stuff like `[[div]]` not being translated but other words are, it's not even automatable easily.

Given the state of the localization system right now I think we can make a PR to cut it.

emmiegit (approved):

PR #431, opened 2021-09-14
Use single-nested list blocks in docs

emmiegit:
Minor change, this documentation was written when the only method of nesting lists was through `[[li]]`, which is incorrect here. This showcases proper sub-lists.


Monkatraz (approved):
Merging for you so I can rebase

PR #434, opened 2021-09-16
[WJ-810] Add include variables

emmiegit:
This stores variables gathered from instances of `[[include-elements]]` and substitutes them as it finds variable tokens. Additionally, it does replace for `[[include-messy]]` if there are any variables passed in.

Unfortunately it was not possible to convince the compiler that the lifetime of `'t` for variables doesn't cause issues with the rest of the render context object, so this pretty extensively clones values (even intermediate ones). Thankfully include variables aren't _that_ prevalent? :shrug: 

As always, it adds AST tests for the new syntax.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #439, opened 2021-09-20
[WJ-816] Refactor: Move partial elements to use separate logic

emmiegit:
This PR makes refactoring changes to move out "partial" elements from being regular `Element` variants and ensures that after each rule execution, that partial elements are not produced (save circumstances which permit them, that is the rules which are looking for those items).

These changes also split out the `tree/elements.rs` file into submodules, as it's gotten rather large, as well as adding a utility `validate_json` which can be used to ensure that the given JSON files represent valid syntax trees.

It also makes a change to the tree structure of list objects, which required updating AST test JSON files.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #444, opened 2021-09-21
cm-tarnation major refactoring, memory and performance upgrades

Monkatraz:
IT'S REALLY FAST NOW

Highlights:
- Everything is now based around the whole `Chunk` idea. Chunks store cached parse data, all emitted tokens, etc.
- It's very fast now. When feasible, token and parse data is now stored as `ArrayBuffer` instances. This lowers on memory significantly, and allows for a reduction on average chunk size. This means you can have more of them, which improves reusing them.
- Forward reuse works better now.
- Refactored both `Parser`, and `Tokenizer` into the `Host` class, which is now just `Parser`. This has removed a lot of fluff, junk, etc. from the whole system. This is why, despite all the additions, I've still deleted more than I've added.
- Nested language handling is done after the entire parse is done, which is accomplished by storing data in nodes. Dynamically nesting a language (e.g. `type` argument in a code block) has to be done with a special `nest` function in the language configuration. This improves performance.

Next PR will move `cm-tarnation` into its own repo, because it's generic and doesn't seriously depend on anything Wikidot/Wikijump related.


emmiegit (approved):
:rocket:

Monkatraz:
thanks @ammongit

PR #446, opened 2021-09-22
[WJ-671] Isolated preview

Monkatraz:
image

This PR prevents styling from breaking out of Sheaf's live preview. This prevents quite a few issues. This PR also fixes some weird tsconfig issues I've been having, particularly in my editor.

### DIFF
--- a/client/modules/components/src/Wikitext.svelte
+++ b/client/modules/components/src/Wikitext.svelte
@@ -47,10 +47,17 @@
   /** Shows render performance information if true. */
   export let debug = false
 
+  /**
+   * If true, the wikitext will be inserted into a shadow DOM so that it is
+   * entirely isolated from the outside DOM from which it is mounted.
+   */
+  export let isolated = false
+
   /** Prevents the rendering of elements which may cause a network request. */
   export let offline = false
 
   let element: HTMLElement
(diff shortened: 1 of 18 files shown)


client/modules/components/src/Wikitext.svelte
```diff
@@ -47,10 +47,17 @@
   /** Shows render performance information if true. */
   export let debug = false
 
+  /**
+   * If true, the wikitext will be inserted into a shadow DOM so that it is
+   * entirely isolated from the outside DOM from which it is mounted.
+   */
+  export let isolated = false
```
emmiegit, line 54: Why is this false by default?

emmiegit (approved):

client/modules/components/src/Wikitext.svelte
Monkatraz: It's a generic component, it's just only being used for the live preview right now. Currently, styling a page or something breaks out of that page's container, so this is just following default behavior. We _could_ actually change that, but that's probably another issue because that has a fair amount of ramifications (e.g. that always requires JS to do, breaks a bunch of articles, etc.).

Monkatraz:
thanks @ammongit

PR #447, opened 2021-09-22
Add some capability to at least test things in the dev-sandbox with Firefox

Monkatraz:
This PR adds the command `preview:dev-sandbox` so that I can at least _test_ if things work in Firefox, and use its profiler. I had to change some worker imports because I hate literally everything about bundlers and I wish everything wasn't terrible and awful


emmiegit (approved):

Monkatraz:
thanks @ammongit

PR #448, opened 2021-09-23
[WJ-841] FTML [[code]] block web component

Monkatraz:
This PR adds the new `ftml-components` module, and inside of it, a web component for handling syntax highlighting of the `[[code]]` block. Additionally, this PR adds updated CSS for code blocks.

image


client/modules/components/src/Wikitext.svelte
```diff
@@ -131,6 +133,12 @@
     element = document.createElement("div")
     element.classList.add("wikitext-body", "wikitext")
     root.appendChild(element)
+
+    // if we're isolated, we need to add the ftml stylesheet to the shadow dom
+    const style = document.createElement("style")
+    style.className = "ftml-base-stylesheet"
```
emmiegit, line 139: (nitpick) shouldn't this be an id not class?

client/modules/ftml-components/src/components/code.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  /**
+   * Extracts the language to highlight with from this elements classes.
+   * Specifically, the `wj-language-{name}` class.
+   */
+  private getLanguageFromClass() {
+    const classes = Array.from(this.classList)
+    for (const name of classes) {
+      if (name.startsWith("wj-language-")) return name.substr(12)
```
emmiegit, line 40: Maybe worth leaving a comment that this is always ASCII lowercase to avoid unnecessary changes in the future by people who don't remember.

ftml/misc/ftml-base.css
```diff
@@ -1,156 +1,248 @@
...
+  word-wrap: normal;
+  tab-size: 4;
+  white-space: pre;
+  cursor: text;
+  contain: content;
+}
+
+.wj-code code {
+  display: block;
+  font-family: "Courier New", Courier, monospace;
```
emmiegit, line 84: Should we use a different monospace font here?

emmiegit (approved):

client/modules/components/src/Wikitext.svelte
Monkatraz: IDs are _supposed_ to be unique, so if there was multiple `Wikitext` components they would conflict. However, this is _technically inside a shadow DOM_, so I don't actually know if that matters. Regardless, this method works fine.

ftml/misc/ftml-base.css
Monkatraz: I'm actually fine with Courier New, I forgot I was thinking of just plain Courier (which is what `monospace` is, as well) when I said the default font was ugly.

I could make a quick sorta-decent-native-font-stack but I think this is okay. This is one of those things that every site should try to customize, I think.

client/modules/components/src/Wikitext.svelte
emmiegit: Is it possible for there to be multiple of this component in one shadow DOM?

But regardless I guess this is fine as is now.

Monkatraz: lol in general I really try to avoid using ID unless it's super obviously a global-ish element, because using ID invites so many quirks.

for example, if you do:
```html
<body>
  <div id="foo">bar</div>
</body>
```

```js
// your editor will complain about this
console.log(foo.innerText) // but... this prints "bar"
```

yeah. blame IE

emmiegit: W-what?

Alright fair enough....

Monkatraz: They added it to [the god damn spec](https://html.spec.whatwg.org/multipage/window-object.html#named-access-on-the-window-object) because the IE devs decided to make "named elements" automatically properties of the global object. For some reason. I guess it was convenient? Everyone else added the behavior because IE was the pack leader and big companies literally tested their sites only on IE. Then, when the spec was getting more standardized, because _everyone had the behavior_, it was kind of undeniably part of the de-facto spec already in place.

Welcome to HTML and JS.

Monkatraz:
thanks @ammongit

PR #450, opened 2021-09-24
[WJ-854] Replace FTML base CSS with a different system in ftml-components

Monkatraz:
This PR replaces the base CSS found in `ftml/misc/ftml-base.css` with SCSS found in `ftml-components`. This is so that CSS can be organized more cleanly, and also use some of the infrastructure we have around SCSS, such as Stylelint. In the process of doing this, I've removed FTML's (the crate) handling of the base CSS file. You're going to need to import `ftml-components` to use FTML anyways, so I think this is fine.


Monkatraz:
I just realized I saved the `README.MD` with Prettier on, which I doubt hurts anything, but that's why there is formatting changes.

ftml/README.md
```diff
@@ -15,7 +15,7 @@
 ### Foundation Text Markup Language
 
 A Rust library to parse Wikidot text ("Wikitext") into an abstract syntax tree (AST).
-This aims to be a replacement for the aging [Text\_Wiki](https://github.com/gabrys/wikidot/tree/master/lib/Text_Wiki/Text) from Wikidot.
+This aims to be a replacement for the aging [Text_Wiki](https://github.com/gabrys/wikidot/tree/master/lib/Text_Wiki/Text) from Wikidot.
```
emmiegit, line 18: Really? This parses as a warning on my editor, which is why I added the escape. I didn't realize Prettier cared about this.

emmiegit (approved):

ftml/README.md
Monkatraz: I think Prettier tries to get rid of any escapes that aren't strictly needed? This was weird to me as well, but I do know some details about the CommonMark spec and I can't see any reason why that `_` would actually need to be escaped.

Monkatraz:
thanks @ammongit 

I'm probably doing footnotes next, because that'll help fill out some infrastructure needed for lightweight tooltips (Tippy.js is a fairly large package, so I might make my own).

PR #453, opened 2021-09-24
Refactor ItemRender use in inner()

emmiegit:
This uses a generic for `ItemRender` instead of a trait object, which eases developer ergonomics by not needing to put `&`s everywhere.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #454, opened 2021-09-24
[WJ-856] Add random ID generation to HtmlContext

emmiegit:
This adds a random number generator used for producing random IDs in generation of HTML attributes that need that. However it's modified so that the output is deterministic, which is necessary for HTML tests.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #455, opened 2021-09-24
[WJ-847] Remove pig latin localization

emmiegit:
As discussed, we are removing the Pig Latin localization. See [WJ-847](https://scuttle.atlassian.net/browse/WJ-847).


Monkatraz (approved):
rest IN PEACE PIG

emmiegit:
thanks @Monkatraz

PR #456, opened 2021-09-24
[WJ-808] Convert token indices to UTF-16

emmiegit:
In addition to the conversion we do to warning indices, this PR extends this to also affect tokens, so that those ranges are accurate.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #457, opened 2021-09-25
[WJ-827] Add underscore line breaks

emmiegit:
Wikidot has a syntactical construction where line breaks can be inserted where they're otherwise not allowed (e.g. in tables or lists) by using `" _\n"`, which is replaced with a regular hard line break.

Previously we had this, but only as a special rule inside tables. Since this is a top-level syntactical construct that is usable anywhere, it should just be a regular rule. This PR adds it, as well as tests for its functionality. This rule also supports a lone `_` on a line as a line break, to prevent the awkward requirement of a space when there's nothing else.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #458, opened 2021-09-25
[WJ-830] Add styling for note

emmiegit:
Because it's a not-commonly-used block that we really don't need, I'm not implementing `[[note]]` as a block. Instead, they will use the default class `wiki-note` to apply the same styling (i.e. `[[div class="wiki-note"]]`).

I'm adding a separate SCSS file for "default", that is, for our various `wiki-xxx` classes.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #459, opened 2021-09-25
[WJ-839] Components for footnote blocks, references, and hovers

Monkatraz:
This PR adds web-components for FTML's footnote markers, both in footnote blocks and everywhere else. This of course includes the styling for footnotes, as well.

image
image

@ammongit This will need the tests updated, pretty please.


client/modules/ftml-components/src/components/footnotes/_footnotes.scss
```diff
@@ -1,8 +1,78 @@
-a.wj-footnote-ref {
-  display: inline-block;
+.wj-footnote-list-item-marker,
+.wj-footnote-ref-marker {
+  @include reset;
+  display: inline;
+  color: #0000ee;
```
emmiegit, line 5: nitpick: can't this be abbreviated as `#00e`?

ftml/src/render/html/element/footnotes.rs
```diff
@@ -24,26 +24,53 @@ pub fn render_footnote(log: &Logger, ctx: &mut HtmlContext) {
     info!(log, "Rendering footnote reference");
 
     let index = ctx.next_footnote_index();
+    let id = &format!("{}", index);
+
+    let footnote_string = ctx.handle().get_message(log, ctx.language(), "footnote");
+    let label = &format!("{} {}.", footnote_string, index);
```
emmiegit, line 31: We can leave this for now, but probably this should be a templated locale string which informs where the number should go, since it's not necessarily after, and it's not necessarily `.`-terminated.

emmiegit (approved):

ftml/src/render/html/element/footnotes.rs
Monkatraz: Yeah, I thought about that. I didn't want to complicate things. This might be fixable on the JS side using the `Intl` standard library, but I'm not sure if it handles indexing like this (it probably does). So yeah, we'll have to wait on this until we actually know what we need to do for a particular language.

client/modules/ftml-components/src/components/footnotes/_footnotes.scss
Monkatraz: I copied the color straight from the [standardized HTML styling recommendations](https://html.spec.whatwg.org/multipage/rendering.html#phrasing-content-3). It could be shortened, but there is an actual meaning behind me picking this _exact_ color.

emmiegit: Huh, okay

Monkatraz:
thanks @ammongit 

I'm going to make a damn PR right now that adds my own functions for making timeouts, this is annoying

PR #462, opened 2021-09-25
Use latest wasm-pack in our ftml CI builds

emmiegit:
This was previously pinned due to an error. But now, it works.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #464, opened 2021-09-25
[WJ-858] wikijump/api module, Mockoon configuration

Monkatraz:
This PR adds a very simplistic API module for the client monorepo. It supports a configuration for a mock API, so that development of the frontend can continue in places which require API usage.

Mockoon is used for the API server, which Vite proxies to. The configuration for this server is derived from the OpenAPI schema. The mocked data generation is fairly terrible right now, but I expect it to be tweaked and adjusted as we go along and add things. I don't see much of a point in investing the effort if the API might need to change quite a bit.

The API code itself is mostly autogenerated from the OpenAPI schema. This is _incredibly handy_ as it takes 90% of the effort of writing the frontend API code away.


client/modules/api/README.md
```diff
@@ -0,0 +1 @@
+# @wikijump/api
```
emmiegit, line 1: Are all the modules README files blank like this?

client/modules/api/mockoon/mockoon-api.json
```diff
@@ -0,0 +1,12235 @@
+{
```
emmiegit, line 1: This file is huge, was it autogenerated?

client/modules/api/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "version": "0.0.0",
+  "keywords": [
+    "wikijump"
+  ],
+  "private": true,
+  "scripts": {
+    "generate-api": "swagger-typescript-api -p ../../../web/resources/api/api.oas3.yaml -o ./vendor --name api.ts --unwrap-response-data  --module-name-index -1 --clean-output && prettier --write \"vendor/**\"",
+    "mock-start": "mockoon-cli start --data ./mockoon/mockoon-api.json --name \"Wikijump\"",
+    "mock-stop": "mockoon-cli stop 0"
+
```
emmiegit: Extra line?

client/web/dev-sandbox/src/static/lib/index.ts
```diff
@@ -1,7 +1,20 @@
...
+  setTimeout(async () => {
+    try {
+      const api = new WikijumpAPI()
+      const response = await api.userGet("id", 1234)
+      console.log("mocked API response:")
+      console.log(response)
+    } catch (err) {
+      console.warn(err)
+      console.log("API probably isn't being mocked")
+    }
```
emmiegit, line 18: I don't think I understand what this snippet does.

emmiegit (approved):

client/modules/api/README.md
Monkatraz: Anything that's been published has a detailed README. Others have nothing like this or just a tiny description. This module is so barebones right now I didn't bother to write a blurb. This will change as I add stuff to the frontend that makes use of the API.

client/modules/api/mockoon/mockoon-api.json
Monkatraz: Sorta - you can treat it like a binary when added all at once like this. It's Mockoon's exported "environment" for Wikijump's API, which was derived from the OpenAPI schema. Smaller changes to the environment should actually be fairly readable.

client/web/dev-sandbox/src/static/lib/index.ts
Monkatraz: All it does is make an API client instance and try to make a request. If this simple request fails, it's probably because the mock server wasn't started. You have to start both Vite and Mockoon (either through the CLI or the application).

client/modules/api/package.json
Monkatraz: Not sure how that snuck in there

client/web/dev-sandbox/src/static/lib/index.ts
emmiegit: Oh I see, regardless of ID it gets a response, makes sense.

client/modules/api/README.md
emmiegit: Fair enough.

Monkatraz:
thanks @ammongit 

now we can do user-info blocks (can you tell I'm putting off doing collapsibles?)

PR #465, opened 2021-09-26
[WJ-824] Add math support via LaTeX

emmiegit:
[[WJ-824](https://scuttle.atlassian.net/browse/WJ-824)] [[Wikidot Documentation](https://www.wikidot.com/doc-wiki-syntax:math)]

This adds `[[$ ... $]]` inline math, `[[math]]`, and `[[eref]]`, which permits creation of mathematical or scientific notation via LaTeX. It uses [latex2mathml](https://github.com/osanshouo/latex2mathml) to generate MathML for use. However in the frontend we will us [KaTeX](https://katex.org/) to render it if the browser doesn't support MathML and Javascript is enabled.

It produces HTML output wrapped in a `wj-math-block` / `wj-math-inline` which has the equation index, name (as `data-name`), LaTeX source (hidden), and generated MathML.


emmiegit:
Note to @Monkatraz, after merging you'll need to ensure the feature `mathml` is selected for whatever wasm stuff you're doing. If it's excluded then `latex2mathml` won't be added to the build and the renderer will only emit the LaTeX source.

ftml/docs/Blocks.md
```diff
@@ -305,6 +308,20 @@ Some text __here!__
...
+Body: None
+
+Arguments:
+* None
+
+Example:
+```
+You can take the area of the circle[[eref Area-Circle]] and use it to find the object's volume.
+```
+
```
Monkatraz, line 324: I never realized this was a thing. This just shows a tooltip of the equation, right?

Monkatraz (approved):

ftml/docs/Blocks.md
emmiegit: Pretty much. It's like a footnote but not styled by major themes apparently lol

emmiegit:
thanks @Monkatraz

PR #466, opened 2021-09-26
[WJ-823] Add the date block

emmiegit:
This PR adds `[[date]]`.

The implementation here is fairly sophisticated compared to Wikidot's: it accepts multiple different date formats other than UNIX timestamps, it splits out the "hover" flag, adds `data-xxx` fields with various time-related information, adds a `tz` (timezone) argument, and makes full use of [`chrono`'s date formatting](https://docs.rs/chrono/0.4.19/chrono/format/strftime/index.html) when it comes to user-specified format strings.


ammongit:
thanks @Monkatraz

Monkatraz (approved):
this is the fanciest date block you'll ever see. we have all the dates. want your date to be utterly unreadable? we got it. you want your date to be the most readable thing you've ever seen? maybe. dates are hard.

emmiegit:
Lol yeah dates were pretty annoying. One thing that made me grimace a bit was knowing the whole time that Wikidot put like 1/10th the effort into their `[[date]]`.

emmiegit:
thanks @Monkatraz

PR #467, opened 2021-09-27
[WJ-861] [WJ-862] FTML math web components

Monkatraz:
This PR adds web components for `[[math]]` blocks (for conditionally polyfilling MathML) and `[[eqref]]` blocks. Polyfilling for MathML is done by the [hfmath library](https://github.com/LingDong-/hfmath). This library renders out _pure_ SVGs, and so doesn't require custom CSS and fonts. The render quality is pretty decent, although it's not perfect. Considering this is only a polyfill for Chrome, and Chrome will be eventually getting MathML anyways, I think this is fine.

Rendered normally:
image

Polyfilled:
image

With errors:
image

Hovering over an equation reference:
image


emmiegit:
So KaTeX.js isn't used here at all?

Monkatraz:
> So KaTeX.js isn't used here at all?

Yep. This avoids needing a lot of CSS and JS. However, KaTeX was always going to be my fallback if hfmath didn't work out.

client/modules/cm-lang-ftml/src/grammars/tex.yaml
```diff
@@ -31,12 +31,15 @@ repository:
...
 
   CommandGroup:
-    match: /(\\#?[a-zA-Z\d]+)(\{)([^]*?)(\})/
-    captures:
-      0: { is: Command }
-      1: { is: Bracket }
-      2: { type: CommandGroupString, tag: string }
-      3: { is: Bracket }
+    begin:
+      match: /(\\#?[a-zA-Z\d]+)(\{)/
```
emmiegit: Does this highlight brackets not associated with a `\item`? You can group stuff using `{..}` without another context.

client/modules/css/src/components/_code.scss
```diff
@@ -26,7 +26,8 @@ pre.code {
 
 // -- TOKEN COLORS
 
-.wj-code-token {
+// weird selector is so we get priority over the default FTML selectors
+.wj-code-token.wj-code-token {
```
emmiegit, line 30: Is this something we should change in the HTML generation?

emmiegit (approved):

client/modules/css/src/components/_code.scss
Monkatraz: Nah, this is just something you have to deal with when doing CSS. It's unfortunate, but you get used to it.

emmiegit: What does this mean for theme creators seeking to override this rule?

client/modules/cm-lang-ftml/src/grammars/tex.yaml
Monkatraz: Right... I knew that too. I'm not quite sure why I had the whole `CommandGroup` thing? I think it was about `\begin{align}` or whatever commands, but that requires actually discriminating between different commands to display properly. I'll remove this whole rule.

client/modules/css/src/components/_code.scss
Monkatraz: Well, if they wanted to override these selectors, they'll need yet more priority. However, the whole point of this particular bit of CSS is to assign CSS variables to all of these different tokens, so a theme creator just needs to change the variables.

Monkatraz:
thanks @ammongit

PR #471, opened 2021-09-27
[WJ-843] Styling and karma support for [[user]] blocks

Monkatraz:
Screenies:
image
image
image

This is all done through FTML's rendering, an SVG, and CSS. No JS needed for this basic widget.

I'm leaving the modal handling to another PR/issue. Why? Well, for one, adding modals is not easy, they have all kinds of issues. Really though, the user modal is not at all needed for FTML's functionality - you can just click through to the user's profile for now. I'm not even sure if I would want a modal, a tooltip might be better (look at Github's).


ftml/src/data/user_info.rs
```diff
@@ -53,10 +56,9 @@ impl UserInfo<'_> {
         UserInfo {
             user_id: 0,
             user_name: cow!("michal-frackowiak"),
+            user_karma: 5,
             user_avatar_data: cow!(AVATAR_BASE64_DATA),
-            user_profile_url: cow!(
-                "https://www.wikijump.com/user:info/michal-frackowiak",
-            ),
+            user_profile_url: cow!("/user:info/michal-frackowiak",),
```
emmiegit: Rustfmt is conservative and doesn't change punctuation in macros since it doesn't know if it matters or not.
```suggestion
            user_profile_url: cow!("/user:info/michal-frackowiak"),
```

ftml/src/data/user_info.rs
```diff
@@ -32,6 +32,9 @@ pub struct UserInfo<'a> {
     /// The user's display name.
     pub user_name: Cow<'a, str>,
 
+    /// The user's karma, from 0-5.
+    pub user_karma: u8,
```
emmiegit, line 36: Since there's only 5 values it should be an enum, but that's out of scope for this PR, I'll change this for you.

emmiegit (approved):

ftml/src/data/user_info.rs
Monkatraz: what is this, ENUMS? I don't even have INTEGERS in MY JAVASCRIPT

Monkatraz:
thanks @ammongit

PR #473, opened 2021-09-27
Remove wildcard-import for data

emmiegit:
It's kind of inelegant having all those data structures exposed at the root namespace, so this PR makes the root module `data` public, and adjusts internal imports.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #475, opened 2021-09-28
[WJ-815] Add [[tabview]]

emmiegit:
This PR adds support for tabs, with its HTML structure informed by [W3 Example Tabs](https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-2/tabs.html) and [MDM Tab Role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role). It adds tabs as a partial, new warnings for tab-related issues, and AST tests to ensure parsing functionality.


Monkatraz (approved):
Looks great

emmiegit:
thanks @Monkatraz

PR #476, opened 2021-09-28
Update client dependencies, minor tweaks due to new updates

Monkatraz:
Does what it says on the tin, basically. Main changes in this PR is me reformatting the Sheaf theme and repacking FTML.


Monkatraz:
I also removed running tests from the TS Validate workflow because they're already being ran in Coverage, so... don't see much of a point in running them twice. This speeds up that workflow by about 30 seconds.

emmiegit (approved):

Monkatraz:
thanks @ammongit

PR #479, opened 2021-09-30
[WJ-837] Build web/ with esbuild

Monkatraz:
This PR is a nice step forward. The `web/` folder is now much more accessible and its legacy JS is now built in less than a second with `esbuild`. You can now just run `pnpm dev` and `pnpm build` and it just works, as long as you have the prerequisites done, like installing Docker and Composer. There is a new file in `docs/` describing how to do this.

~~Additionally, this PR changes it so that the `client/` monorepo formatting infrastructure is now in a root level NPM module under `formatting/`. This module really helps to cleanup the `package.json` files and makes it so that all the rules are shared.~~

~~One thing I didn't do was add [`prettier-plugin-php`](https://github.com/prettier/plugin-php). This could be added to `formatting/`, and would let us automatically format `php` files, just like we do with CSS, TS, JS, etc. If there is interest in it, I could add it in this PR or in another one. If we wanted to go all out, I bet there would be a way of getting Rust's `rustfmt` to inherit from `formatting/` too, lol.~~

~~The `formatting/` module itself is a bit strange, due to the various config resolving quirks I had to contend with. However, it works fine once you get past those quirks.~~


emmiegit:
We're not autoformatting any PHP files until the amount of Wikidot code remaining is essentially nil. It's a very useful indicator for whether we've "touched" a file before, and it would also produce a lot of pollution in the git blame.

What's the benefit in this unified `formatting/`, like if we got `rustfmt` to use it?

Monkatraz:
>
> What's the benefit in this unified `formatting/`, like if we got `rustfmt` to use it?

It's just that all the formatting configuration is in one place, and reduces on the number of fluff dependencies per package. The `rustfmt` thing was kind of a joke, because `rustfmt` is so simple I don't think you would need to care about where it's configured, lol. The centralized formatting is a big benefit for `client/` and `web/` though, because I don't need to duplicate a bajillion configuration files.

emmiegit:
Okay, having a central formatting for frontend makes sense. I'm not sure I understand why it should be in the root directory? Well I guess it makes sense if I think of it like `.github`, a `.formatting` or something? idk

Monkatraz:
> 
> 
> Okay, having a central formatting for frontend makes sense. I'm not sure I understand why it should be in the root directory? Well I guess it makes sense if I think of it like `.github`, a `.formatting` or something? idk

Where else could it go? `web/` and `client/` are siblings, so the only place that makes sense is another sibling directory, or putting the configuration files themselves directly at root. Which would work, but that's really messy.

docs/frontend-development.md
```diff
@@ -0,0 +1,179 @@
...
+> [WSL2 download and installation page](https://docs.microsoft.com/en-us/windows/wsl/install-win10)
+
+In general, the process of running Wikijump on your machine to undertake development work is using what is called the "local" environment. The other environments are "development", which is a build running on a server for the sake of development work, and "production", which is the environment that consumers actually connect to.
+
+## Setup: Utilites and Programs
+
+You will need some utilites and programs to get started. You will to install the following:
+
+- [PHP 7.4](https://www.php.net/downloads)
+- [Composer (PHP dependency manager)](https://getcomposer.org/)
```
emmiegit, line 24: You don't actually need either, you can fully run PHP locally since it's set up in the Docker containers. The same is not true of pnpm etc though.

emmiegit:
I think my concern is more about the name `formatting`. We are probably going to accumulate other miscellaneous configuration-like files, maybe a directory more suited for that name at the root makes sense? `config/` or some such

docs/frontend-development.md
```diff
@@ -0,0 +1,179 @@
...
+>
+> Once that's all done, just leave Docker Desktop running in the background whenever you want to do development work.
+
+Node, NPM, and PNPM are well-behaved on Windows and Linux, and the difference in usage between operating systems is negligable.
+
+## Setup: Configuration
+
+Wikijump has a lot of configuration options that can set. The primary configuration file can be found in `web/conf/`, named `wikijump.ini`. Freshly cloned, your Wikijump repo won't have this file. However, it will have an example configuration, `wikijump.ini.example`, which you can just copy and rename to `wikijump.ini` for the time being.
+
+Second, there is a Docker configuration file that configures the various containers that host Wikijump in the local environment. You can find this file in `install/local/dev/`, named `docker-compose.yaml`. Like before, you won't have this file with a fresh Wikijump clone. However, again like before, there is a `docker-compose.yaml.example`, which you can just copy and rename to `docker-compose.yaml`.
```
emmiegit, line 61: It's worth noting `docker-compose.dev.yaml` I think.

Monkatraz:
> 
> 
> I think my concern is more about the name `formatting`. We are probably going to accumulate other miscellaneous configuration-like files, maybe a directory more suited for that name at the root makes sense? `config/` or some such

I'm fine with `config/`, but that maybe implies Wikijump instance configuration. Would it be `config/formatting/` or just renaming it to `config/`?

emmiegit:
`config/formatting/` makes a lot of sense to me. If you think it's unclear and implies Wikijump instance configuration, would `.config` be better?

Monkatraz:
> 
> 
> `config/formatting/` makes a lot of sense to me. If you think it's unclear and implies Wikijump instance configuration, would `.config` be better?

Hm, yeah, I think `.config` is good. Let me make those changes real quick.

EDIT: donezo

.config/formatting/prettier.config.js
```diff
@@ -0,0 +1,33 @@
...
+
+  "plugins": [
+    "./node_modules/prettier-plugin-organize-imports",
+    "./node_modules/prettier-plugin-jsdoc",
+    "./node_modules/prettier-plugin-svelte"
+  ],
+
+  // plugin-jsdoc
+  "tsdoc": true,
+  //Set shorter than 90 so if line wrapping is on comments don't go past it
```
emmiegit:
```suggestion
  // Set shorter than 90 so if line wrapping is on comments don't go past it
```

web/nabs.yml
```diff
@@ -0,0 +1,28 @@
+# -- BUILD
```
emmiegit, line 1: nabs?

emmiegit (approved):

web/nabs.yml
Monkatraz: Handy NPM utility for writing `package.json` scripts without the usual pain that comes with doing that.

docs/frontend-development.md
Monkatraz: I think you'll need it for `laravel-vite` later on, so while what you're pointing out is technically true now, I don't think it will be later.

Monkatraz:
Turns out, the whole unified formatting thing just doesn't work due to weird stupid limitations! Sigh.

Monkatraz:
god finally

thanks @ammongit

just touching the `web/` folder causes utter chaos

PR #481, opened 2021-09-30
[WJ-836] Use Cascadia Mono for the code font

Monkatraz:
Cascadia Mono is a variable monospace font with support for bold and italics. It also has wide unicode support, so Roboto Mono isn't needed now. I think it looks pretty good.

image
image


emmiegit (approved):

Monkatraz:
thanks @ammongit

PR #483, opened 2021-10-01
[WJ-821] Add [[embed]]

emmiegit:
This is a bit different from Wikidot's `[[embed]]`, but it's hopefully more useful. Instead of regex-validated direct insertion of HTML, this builds the HTML properly from premade templates. This is more robust, secure, but suffers from the downside of needing to be manually implemented for each source instead of just adding a regex rule. It will also cause problems with existing direct iframes, but those can just be transposed into regular `[[iframe]]` blocks.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #485, opened 2021-10-01
[WJ-873] Have ftml emit autonomous custom elements instead of customized built-in elements

emmiegit:
This removes the `is` attribute and changes the element types to be in the convention of `wj-xxx`. This also updates the AST tests.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #486, opened 2021-10-01
Hotfix: Code blocks are not paragraph safe

emmiegit:
This changes `Element::Code` to no longer be considered phrasing, since it's div-like. This was previously an issue, but only surfaced after we changed to custom elements.

### DIFF
--- a/ftml/src/tree/element/object.rs
+++ b/ftml/src/tree/element/object.rs
@@ -356,7 +356,7 @@ impl Element<'_> {
             Element::User { .. } => true,
             Element::Date { .. } => true,
             Element::Color { .. } => true,
-            Element::Code { .. } => true,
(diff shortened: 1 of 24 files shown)


ftml/src/tree/element/object.rs
```diff
@@ -356,7 +356,7 @@ impl Element<'_> {
             Element::User { .. } => true,
             Element::Date { .. } => true,
             Element::Color { .. } => true,
-            Element::Code { .. } => true,
+            Element::Code { .. } => false,
```
emmiegit, line 359: This is the core change, everything else is metadata or updating tests.

Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #488, opened 2021-10-01
[WJ-447] [WJ-840] FTML collapsible DOM changes, styling, functionality

Monkatraz:
This PR makes `[[collapsible]]` blocks work now, with great accessibility to boot. They use `<details><summary></summary></details>`, which is the "disclosure widget". This means that the main collapsible functionality requires literally no JS. The only JS involved here is to make the bottom hide-button location work.

image

https://user-images.githubusercontent.com/34875062/135695356-d4a569dc-b58b-4827-94b8-c711fa76b1dc.mp4


ftml/src/render/html/element/collapsible.rs
```diff
@@ -87,78 +87,56 @@ pub fn render_collapsible(log: &Logger, ctx: &mut HtmlContext, collapsible: Coll
...
-                                    "class" =>
-                                        "wj-collapsible-block-link "
-                                        collapsible_class(show_bottom),
-                                    "href" => "javascript:;",
-                                ))
-                                .inner(log, hide_text);
-                        });
-                });
+            // Bottom open/close button
+            if show_bottom {
```
emmiegit, line 127: `show_top` has different logic than `show_bottom`?

emmiegit (approved):

ftml/src/render/html/element/collapsible.rs
Monkatraz: The top button has to always be in the DOM because that button is how you control the collapsible when it's folded. CSS controls hiding the top button if it's not supposed to be visible.

Monkatraz:
thanks @ammongit

PR #489, opened 2021-10-02
Remove install/legacy/

emmiegit:
I have created a branch [`legacy-install`](https://github.com/scpwiki/wikijump/tree/legacy-install) to maintain this info if it's later needed. By removing it here we don't have to worry about it affecting grep results or accidentally being changed as part of larger efforts.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #490, opened 2021-10-02
[WJ-879] Refactor multi-step Docker build

emmiegit:
This PR adjusts our build process so that the various components which are included in the final image happen intentionally as separate steps, which are then copied from. Importantly, does not use `git clone`, but copies source files directly from the local filesystem, improving speed and making local iteration easier.

Because we take sources from the repo directly, we now no longer need to duplicate several files used in the deployment process. These have been moved to `install/files/`, with environment-dependent ones in one of `local`, `dev`, or `prod` respectively.

This also adds workflow files for ensuring container builds work before pushing to `develop` or `prod`.


Monkatraz (approved):
yikes... great work, you poor soul

emmiegit:
It wasn't that bad honestly

emmiegit:
thanks @Monkatraz

PR #491, opened 2021-10-03
Reorganize client/, add Asset class

Monkatraz:
This PR is a little messy, but it's because it's the first part of two PRs (maybe three). This PR reorganizes client so that it has very few folders inside of it. It's now just `.changeset`, `modules`, and `scripts`. The next PR will move `client/` (its contents, not the folder itself) into `web/`. This, while unfortunately making `web/` yet messier, will simplify a lot of tooling issues. Ultimately, `web/` _will_ be cleaned up, so I think this is okay.

Main changes:
* Reorganized the `misc` folder out
* Removed the `state` module (its code was given to the other modules, such as `api`)
* Added the `Asset` class, which will soon be used for the new frontend
* Removed the `web/` folder in `client/` (the `docs` website was unused, and `dev-sandbox` is now in `modules` but probably won't exist for much longer)

Once this PR, the followup PR, and #490 are done, work on the new frontend can commence.


client/modules/api/src/asset.ts
```diff
@@ -0,0 +1,149 @@
+/** The location of static assets, from host root. Can't end with a `/`. */
+export const STATIC_FOLDER = "/files-common"
```
emmiegit:
```suggestion
export const STATIC_FOLDER = "/files--common"
```

emmiegit (approved):

Monkatraz:
thanks @ammongit

PR #492, opened 2021-10-03
[WJ-877] Refactor tags table schema

Yossipossi1:
[WJ-877](https://scuttle.atlassian.net/browse/WJ-877) ("Refactor tags table schema") has been mostly resolved with this PR. This PR modifies the `page` table to have an column that stores JSONB arrays for the tags instead of a separate table. Additionally, it also refactors the php in numerous areas to account for this change, and drops the now-deprecated `page_tags` table.

### DIFF
--- a/web/database/migrations/2021_09_30_065641_modify_tags_table.php
+++ b/web/database/migrations/2021_09_30_065641_modify_tags_table.php
@@ -1,0 +1,43 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class ModifyTagsTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+      // Creating new column in Page table.
+      Schema::table('page', function (Blueprint $table) {
+        $table->jsonb('tags')->default('[]');
+      });
+
+      // Moving all tags from 'page_tag' table to new column in 'page' table.
+      $pages = DB::table('page_tag')->pluck('page_id')->toArray();
+      foreach ($pages as $page) {
+        $tags = DB::table('page_tag')->where('page_id', $page)->pluck('tag')->toArray();
+        DB::table('page')
+          ->where('page_id', $page)
+          ->update(['tags' => $tags]);
+      }
+
+      // Drop 'page_tag' table.
+      Schema::drop('page_tag');
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        //
+    }
+}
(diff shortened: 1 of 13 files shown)


web/database/migrations/2021_09_30_065641_modify_tags_table.php
```diff
@@ -0,0 +1,43 @@
...
+    {
+      // Creating new column in Page table.
+      Schema::table('page', function (Blueprint $table) {
+        $table->jsonb('tags')->default('[]');
+      });
+
+      // Moving all tags from 'page_tag' table to new column in 'page' table.
+      $pages = DB::table('page_tag')->pluck('page_id')->toArray();
+      foreach ($pages as $page) {
+        $tags = DB::table('page_tag')->where('page_id', $page)->pluck('tag')->toArray();
```
emmiegit, line 24: nitpick:
```suggestion
        $tags = DB::table('page_tag')->where('page_id', $page)->pluck('tag')->toArray();

```

web/database/migrations/2021_09_30_065641_modify_tags_table.php
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        //
```
emmiegit: You need to implement a `down`

web/php/DB/Page.php
```diff
@@ -161,20 +161,7 @@ public function getSite()
 
     public function getTags()
     {
-        $c = new Criteria();
-        $c->add('page_id', $this->getPageId());
-        $tags = PageTagPeer::instance()->select($c);
-        return $tags;
-    }
-
-    public function getTagsAsArray()
```
emmiegit, line 170: Can you confirm `getTagsAsArray()` does not appear at all in the codebase now?

Monkatraz (approved):

web/php/Screens/Wiki/WikiScreen.php
```diff
@@ -147,16 +147,12 @@ public function render($runData)
...
-            $c = new Criteria();
-            $c->add("page_id", $page->getPageId());
-            $c->addOrderAscending("tag");
-            $tags = PageTagPeer::instance()->select($c);
-            $t2 = array();
-            foreach ($tags as $t) {
-                $t2[] = $t->getTag();
+            // Get the tags and convert them to string.
+            $tags = PagePeer::getTags($pageId);
+            if($tags != null) {
```
emmiegit:
```suggestion
            if($tags !== null) {
```

emmiegit: Why would `$tags` be null?

emmiegit (approved):

web/php/DB/Page.php
Yossipossi1: Fixed one that I missed, thank you. Should be entirely gone now.

PR #503, opened 2021-10-06
Remove remaining uses of the old local domain

emmiegit:
This PR fixes up some remaining occurrences of `wikijump.test`. This also changes the variables in the containers, which caused issues with local deployment.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -4,7 +4,7 @@ This document will explain how to set up Wikijump on your machine for local deve
 
 ## Environment
 
-### Operating system
+### Operating System
 
 To run Wikijump, you will need to be running Linux.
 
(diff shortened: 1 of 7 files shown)


emmiegit:
@Monkatraz Given that `development.md` wasn't updated in your PR, and they contained those large bits of duplicated instructions between them, to what extent should we merge or delegate those documents?

Monkatraz:
I'm not sure. I think that `development.md` should probably be very backend focused. I think it's okay if info is duplicated as long as if comes from different contexts.

Monkatraz (approved):

emmiegit:
Then I feel like it would make sense to rename it to `development.md`?

But honestly, you need to have the frontend working even if you're only changing the PHP, so wouldn't I need to follow the same instructions there?

Or perhaps:
* `development.md` -- minimal doc that describes basic local deployment, frontend and backend
* `frontend.md` -- describes things specific to frontend not needed if you're not touching it
* `backend.md` -- ditto for backend

Monkatraz:
Yeah, something like that sounds fine. I do think that have one cohesive guide rather than it split into multiple files is better, but it doesn't matter that much.

emmiegit:
Okay, in that case I prefer the one cohesive `development.md` that covers everything. Would you be able to do that sometime, since you wrote the frontend doc?

emmiegit:
thanks @Monkatraz

Monkatraz:
> Okay, in that case I prefer the one cohesive `development.md` that covers everything. Would you be able to do that sometime, since you wrote the frontend doc?

Yeah, I'll get to it sooner or later, probably whenever I need to touch the docs folder again (probably won't be long lol)

PR #504, opened 2021-10-06
[WJ-883] Flexible, consistent base Blade layout

Monkatraz:
This PR adds a detailed layout that the new frontend views will inherit from. It supports pretty much every meta tag you should care about, but doesn't require that you specify any of them, as every "argument" is completely optional.

The formatting is a bit peculiar - I used [blade-formatter](https://github.com/shufo/blade-formatter), with `aligned-multiple` arguments and a line wrap size of 90. At some point, I'll need to standardize formatting of Blade templates, but I'm not quite sure how yet. There isn't a decent Prettier plugin for these, I don't think.


emmiegit:
`blade-formatter` looks nice, though I wonder if the smaller line width will cause issues eventually.

web/resources/views/next/base.blade.php
```diff
@@ -0,0 +1,172 @@
+{{-- -
```
emmiegit, line 1: Does blade-formatter add this `-`?

web/resources/views/next/base.blade.php
```diff
@@ -0,0 +1,172 @@
...
+    <meta property="og:locale" content="{{ app()->getLocale() }}">
+    @isset($social_url)
+        <meta property="og:url" content="{{ $social_url }}">
+    @endisset
+    @isset($social_title)
+        <meta property="og:title" content="{{ $social_title }}">
+    @endisset
+    @isset($social_description)
+        <meta name="description" content="{{ $social_description }}">
+        <meta property="og:description" content="{{ $social_description }}">
```
emmiegit, line 81: Are these two always the same? From what I understand they can be different in some circumstances. (Nitpick)

web/resources/views/next/base.blade.php
```diff
@@ -0,0 +1,172 @@
...
+    @endisset
+
+    {{-- Social Other --}}
+    @isset($rating)
+        <meta name="rating" content="{{ $rating }}">
+    @endisset
+
+    {{-- Browser Metadata --}}
+    @isset($title)
+        <title>{{ $title }}</title>
```
emmiegit, line 108: Isn't a `<title>` required?

web/resources/views/next/base.blade.php
```diff
@@ -0,0 +1,172 @@
...
+
+    {{-- Styles --}}
+    @vite('resources/css/base.scss')
+    @stack('styles')
+
+    {{-- Vite --}}
+    @client
+
+    {{-- Misc --}}
+    <link rel="author" href="humans.txt"> {{-- TODO --}}
```
emmiegit, line 155: I'm unfamiliar with this pattern, what is `humans.txt`?

web/resources/views/next/base.blade.php
```diff
@@ -0,0 +1,172 @@
...
+
+    @stack('head')
+</head>
+
+<body class="light codetheme-dark">
+    @yield('content')
+
+    <div id="toasts"></div>
+    <div id="modals"></div>
```
emmiegit, line 164: Do we want these to have `wj-`? (I guess this also applies to other classes used here too)

emmiegit (approved):

web/resources/views/next/base.blade.php
Monkatraz: No, but it's related to the formatter. It tries to start and end comments with no newlines, and would mess up my formatting quite a bit. I add the `-` to satisfy it and let me keep my newline, effectively.

web/resources/views/next/base.blade.php
Monkatraz: I'm not sure _why_ they would be different. The only reason to make them different is if you want Facebook/Twitter/Discord/etc. descriptions to be different from like, Google and browser bookmark descriptions.

I mean, maybe you do, but the fact that different services pull from either `description` or `og:description` (and you won't know which), means that the only sane thing to do (IMO) is to set them both to the same value.

web/resources/views/next/base.blade.php
Monkatraz: Nope, at least technically.
image

For all practical purposes, yeah, you should _really_ give one. However, this layout isn't trying to validate what you give it, so, best to stick with the pattern and keep it optional.

web/resources/views/next/base.blade.php
Monkatraz: It's a cool little file you can include with websites to give credit to those who contributed to it. [See here](https://humanstxt.org/). It doesn't have to be machine readable, and it's just a text file.

web/resources/views/next/base.blade.php
Monkatraz: Well, I suppose there could be conflicts from a user doing `id="toasts"` or whatever, but if it was namespaced they obviously could still mangle IDs. Really, we should be sanitizing user input for IDs (it's done by DOMPurify, for example), as it can be a sorta security issue in weird scenarios.

For classes, this is more debatable, but I haven't gotten around at all to site styling beyond the basics I made for the editor. We'll be able to go into way more detail when I get around to making the first actual view.

web/resources/views/next/base.blade.php
emmiegit: Oh I see, hmm.

web/resources/views/next/base.blade.php
emmiegit: I mean like the HTML validator doesn't like if it's missing, I don't doubt browsers can handle it like they do with malformed HTML.

Monkatraz: It isn't malformed if it doesn't include `<title>`. It's just bad practice. I don't see anything in MDN docs that make supplying a `<title>` element required.

Monkatraz:
thanks @ammongit 

it's really rewarding to actually be working on the actual site lol

PR #505, opened 2021-10-06
Restructure docker build to better cache steps

emmiegit:
The separated `setup.sh` had the benefit of reducing intermediate image size since there were fewer steps. However some commands are very expensive, and needing to continually rerun them is unfortunate. This PR splits that script into several components, to better cache progress when only a small portion has been changed.

Additionally, it moves the stage where all of `web/` is copied to as late as possible, to benefit from caching. This is because, if anything in `web/` is changed (a near-certainty when developing anything on the frontend or backend), this will cause a full rebuild, including all the long PHP extension compilation steps.


install/files/prod/env
```diff
@@ -0,0 +1,60 @@
...
+PUSHER_APP_ID=
+PUSHER_APP_KEY=
+PUSHER_APP_SECRET=
+PUSHER_APP_CLUSTER=mt1
+
+MIX_PUSHER_APP_KEY="${PUSHER_APP_KEY}"
+MIX_PUSHER_APP_CLUSTER="${PUSHER_APP_CLUSTER}"
+
+# Initial password for the Admin user
+ADMIN_INITIAL_PASSWORD=jump1vtezJAx91ONtzk4tCLJf1BT9x0r6jiMHkGm
```
Monkatraz: Is this supposed to be in the repo?

Monkatraz (approved):

install/files/prod/env
emmiegit: I thought it was weird having `admin1` as our provisional prod password so I generated something randomly for fun I guess lol. Maybe I should change it back to `admin1` so it's clearly provisional.

emmiegit:
thanks @Monkatraz

PR #506, opened 2021-10-08
[WJ-759] [WJ-221] Remove "site super settings"

emmiegit:
This useless table had only one field: `can_custom_domain`, which was in its own table instead of the site settings table. The PR removes the table, and also removes the concept of custom domains being enable-able in the first place, since it doesn't make sense for that to be a wiki-specific option.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #509, opened 2021-10-08
[WJ-765] Remove ucookie

emmiegit:
As far as I can determine, this is wholly unused. What exactly it's supposed to do is not clear, there's code to generate (insecurely, of course) some key, but it's not stored anywhere. If anybody has hypotheses about what this is, they are appreciated.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #510, opened 2021-10-08
[WJ-762] [WJ-345] Remove MailForm, DatabaseStorage, and StorageItem.

emmiegit:
Wikidot has a strange table `storage_item` which is useful for storing random, unstructured nonsense in the database, for use by miscellaneous modules. This is obviously a bad pattern, and it was only used by the `MailForm` module (which permits users to send random emails as the platform), which we wanted to remove for various reasons anyways. This PR removes both, as well as `DatabaseStorage`, which is utility class for handling `StorageItem` database rows.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #511, opened 2021-10-10
[WJ-893] "Frame" layout which wiki views extend

Monkatraz:
This is PR 1 of 2. The second PR will handle styling separately, as this PR is already getting kind of big. This new layout handles stuff like the header, navbar, footer, etc. These elements will probably be present on practically every page a user visits.

This PR has a lot of renamed/moved files, as I removed the `@wikijump/css` package by moving all of its contents to `web/resources/`. I want to focus on keeping the `web/modules/` folder on being JS oriented and for each module to be fairly self-contained.


web/resources/views/next/frame.blade.php
```diff
@@ -0,0 +1,126 @@
...
+            @includeWhen(isset($navbar_items), 'next.components.nav-dropdowns', [
+                'items' => $navbar_items,
+            ])
+        </nav>
+
+
+        {{-- Sidebar --}}
+        @isset($sidebar_content)
+            <aside id="sidebar" aria-label="{{ __('frame.aria_sidebar') }}">
+                {!! $sidebar_content !!}
```
emmiegit, line 79: Unrelated to your PR, but I love this syntax so much more than Smarty's (which is just "any field you haven't added `|escape` to). Very clearly indicates HTML is being added directly.

emmiegit (approved):

Monkatraz:
thanks @ammongit

PR #513, opened 2021-10-10
[WJ-729] Remove the licenses table, make it config

emmiegit:
This PR removes the `licenses` table, which did not scale horizontally and was essentially just constant data, into `config/licenses.php`, which is then processed by `app/Services/License` and exposed for the bottom of pages and for the part of the admin panel which sets licenses.

It also renames the `license_default` column in the `category` table to `license_inherits` since that is more clear in terms of what it does, and the concept of `license_other`, which allows admins to supply custom license HTML, but it's janky and dealing with user-provided HTML is always a minefield.

Screenshots:

image
image
image


web/database/migrations/2021_10_09_055220_remove_license_table.php
```diff
@@ -0,0 +1,61 @@
...
+        // Drop old columns
+        Schema::table('category', function (Blueprint $table) {
+            $table->dropColumn('license_id');
+            $table->dropColumn('license_other');
+            $table->dropColumn('license_default');
+        });
+
+        // Add new columns
+        Schema::table('category', function (Blueprint $table) {
+            $table->string('license_id')->default('cc_by_sa_4');
```
Yossipossi1, line 25: I'm not sure having CC-BY-SA 4.0 be the default license is the best option here. I know there's some weirdness involved with 4.0 across countries; are we sure we don't want to have the default be 3.0? Might be worth asking the Licensing Team for their input.

emmiegit: It's worth noting that this is just what's set on site creation, it's trivial to override it, it's not prescribing all sites on Wikijump use it as their license.

Yossipossi1: No yeah, I know. I'm just saying that for Wikis that aren't concerned with licensing when they're created, wouldn't it be a better option to have it be default CC-BY-SA 3.0 instead, since the legality in other countries for 4.0 has been kind of funky?

web/php/DB/Page.php
```diff
@@ -78,27 +78,18 @@ public function getCategory()
...
 
-        $category = CategoryPeer::instance()->selectById($categoryId, $siteId);
-        return $category;
+        return CategoryPeer::instance()->selectById($categoryId, $siteId);
     }
 
     public function getTitleOrUnixName()
     {
         $title = $this->getTitle();
-        if ($title == null || $title === '') {
```
Yossipossi1, line 88: Isn't it also possible for a title to be blank? Would there be problems if this check was removed?

emmiegit: This isn't a removal, this is a change, see below.
PhpStorm recommended it as a change since if you do `$title == null` the `$title === ''` becomes redundant.

emmiegit: Well I think the real confusion here is `==` bs. I'm going to just change it to `===` so it has the meaning I think you're implying here.

Yossipossi1 (approved):

web/database/migrations/2021_10_09_055220_remove_license_table.php
emmiegit: I've asked Licensing about this.

emmiegit:
thanks @Yossipossi1

PR #516, opened 2021-10-10
[WJ-757] [WJ-896] Remove Wikidot's site backups

emmiegit:
Wikidot has a site backup feature, but its outputs are not complete, organized, and cannot be used to restore a site. We will eventually add a real backup feature, but for now we should remove this legacy code since it's not useful.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #517, opened 2021-10-10
[WJ-888] [WJ-898] Remove "site clone" feature and FriendlyCaptcha

emmiegit:
This PR removes the ability to clone sites, as well as our FriendlyCaptcha integration. This does _not_ remove the `Duplicator`, because that is needed to create new sites, and until we replace that, we should retain this here. (That said, the ability to create new sites is currently nonfunctional for a number of reasons outside the scope of this PR).


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #518, opened 2021-10-10
[WJ-899] Remove star dependency from ftml

emmiegit:
This will permit us to publish the latest version of `ftml` as a crate on [crates.io](https://crates.io).


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #520, opened 2021-10-10
Redo nginx config, add files--static and files--built pathing, move index.php to public

Monkatraz:
This monster of a PR (even though it's fairly short) does a lot. Primarily, it makes it so that the `web/public/` folder is actually where Laravel's `index.php` is, and adds the paths `files--static` and `files--built` to the `public/` folder.

This means that `web/web/` is entirely legacy, and once we are rid of Wikidot's old code, can just be deleted. Laravel actually expects `index.php` to be inside of `public/`, so this fixes a bunch of issues `laravel-vite` was having with our setup.

Additionally, issues with static assets sometimes having to be proxied has been resolved. This is due to the new `files--static` folder, which you can think of as our new `files--common`, except with no built assets. This folder is bound to nginx via your filesystem in dev, so there isn't any need for Vite to be handling it now. Built files go into `files--built`.

Since `web/web/` is 100% legacy now, it may be interesting to see if we can move `web/bin/`, `web/templates/`, `web/lib/`, and `web/php/` into `web/web/`. This would significantly cleanup `web/` and make all top level directories and files except `web/web/` solely related to our Laravel implementation.


install/files/dev/env
```diff
@@ -3,7 +3,7 @@ APP_ENV=dev
 APP_DEBUG=false
 APP_KEY=
 APP_URL=https://www.wikijump.dev/wikijump--next
-ASSET_URL=https://www.wikijump.dev/wikijump--next/assets
+ASSET_URL=/files--built
```
emmiegit, line 6: Honestly I'm surprised this works.

install/files/nginx.conf
```diff
@@ -2,156 +2,178 @@
...
+        }
+
+        # Static files directory (legacy)
+        location /files--common/ {
+            root /var/www/wikijump/web/web;
+            try_files $uri =404;
+        }
+
+        # Make sure some particular files are always available
+        location = /robots.txt {}
```
emmiegit, line 102: And add `humans.txt`?

install/files/nginx.conf
```diff
@@ -2,156 +2,178 @@
...
+
+            # If for some reason you absolutely MUST boot an Ozone instance directly, you must prepend the page
+            # with `/wikijump--legacy/`
+            rewrite ^\/(?:wikijump--legacy\/)(.*)$ /../web/ozone.php?Wiki__WikiScreen/wiki_page/$1? last;
+
+            # Rewrite URL so we can pass to PHP
+            rewrite ^\/(.*)$ /index.php?Wiki__WikiScreen/wiki_page/$1? last;
+
+            # Final fallback (not sure how this would happen)
+            try_files $uri $uri/ /index.php?$args;
```
emmiegit: Since this shouldn't happen (pages should definitely be final) can we have this HTTP 500 instead? I'd rather catch this as an error rather than debug some weird, forgotten behavior.

install/files/nginx.conf
```diff
@@ -2,156 +2,178 @@
...
+        }
+
+        # Point PHP files to the PHP-FPM service
+        location ~ \.php$ {
+            # regex to split $uri to $fastcgi_script_name and $fastcgi_path
+            fastcgi_split_path_info ^(.+?\.php)(/.*)$;
+
+            # Check that the PHP script exists before passing it
+            try_files $fastcgi_script_name =404;
```
emmiegit, line 147: Does this work? I don't think nginx has a copy of the PHP files in its container.

install/files/nginx.conf
```diff
@@ -2,156 +2,178 @@
...
+
+            # Set buffer size large enough for Laravel debug stack traces to come through
+            # LOCAL AND AWS DEV ONLY - There shouldn't be any reason to set this for prod.
```
emmiegit, line 150: rip this is installed for all three, hmm

web/modules/api/src/asset.ts
```diff
@@ -1,5 +1,5 @@
 /** The location of static assets, from host root. Can't end with a `/`. */
-export let STATIC_FOLDER = "/wikijump--next/assets" as const
```
emmiegit, line 2: What is `export let as const`?

emmiegit (approved):

install/files/nginx.conf
Monkatraz: I've been thinking about it and I think this gets hit when you have a bare URL, like `www.wikijump.localhost` with nothing after it. So, I think I need to do what you suggested (add something like `return 500 "unresolvable path"`) and then add a `location = / { ... }` block.

install/files/nginx.conf
Monkatraz: nginx copies the entire `web/` folder, so it has all the PHP files. I didn't touch anything in the PHP-FPM stuff.

web/modules/api/src/asset.ts
Monkatraz: `export let` is normal, `STATIC_FOLDER` is just a live binding (or it was). `as const` is a TypeScript thing, it means to infer the type _exactly_ as it was given, and not generically. Like:
```ts
let foo = "asdf" // type: string
let bar = "asdf" as const // type: "asdf"
const foobar = "asdf" // type: "asdf"
```

install/files/dev/env
Monkatraz: As far as I know, all `ASSET_URL` is for is the `asset()` function (and anything else that acts like that function). It just prepends any asset URL with the `ASSET_URL` string, which might just be empty.

install/files/nginx.conf
Monkatraz: Okay, I just rewrote this fallback to be a bit more clear about what's happening. It also throws a 500 if (SOMEHOW) `index.php` is missing.

install/files/nginx.conf
emmiegit: Really? Wack.

install/files/nginx.conf
Monkatraz: We can fix this in another PR. There might be a way to setup nginx to figure this out automatically? Although when I looked around it seems that trying to use env vars in nginx is a bit odd and I don't think very encouraged.

Monkatraz:
thanks @ammongit 

let's see if dev works now, lol. god I hope so

PR #524, opened 2021-10-10
Fix localization build

emmiegit:
The `/locales` build [was failing](https://github.com/scpwiki/wikijump/runs/3850612773?check_suite_focus=true) because pylint thought that `messages/messages.py` referred to the head module `messages`, and could not resolve names properly. To avoid this disconnect between the runtime and the linter, I have renamed the file.

### DIFF
--- a/locales/messages/__init__.py
+++ b/locales/messages/__init__.py
@@ -6,7 +6,6 @@
 Parent file for the messages module.
 """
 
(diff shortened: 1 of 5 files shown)


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #526, opened 2021-10-11
[WJ-884] Initial work for wikitext settings in ftml

emmiegit:
This PR adds the concept of `WikitextSettings` to ftml, which has a number of flags that control parsing or rendering behavior, and adds logic to handle the settings we currently have. The FFI and WASM are changed to allows settings to be passed in, and some tests checking the difference in behavior by `WikitextMode` were added.


ftml/src/ffi/exports.rs
```diff
@@ -42,13 +44,19 @@ fn render<R: Render>(
             .expect("Passed PageInfo structure from C was null")
             .to_page_info()
     };
+    let settings = unsafe {
```
Monkatraz, line 47: that's it, I'm not using your crate now /s

ftml/src/render/handle.rs
```diff
@@ -74,23 +75,36 @@ impl Handle {
...
             ImageSource::File1 { file } => (&info.site, &info.page, file),
             ImageSource::File2 { page, file } => (&info.site, page, file),
             ImageSource::File3 { site, page, file } => (site, page, file),
         };
 
-        // TODO
-        Cow::Owned(format!(
+        // TODO: emit url
+        Some(Cow::Owned(format!(
             "https://{}.wjfiles.com/local--files/{}/{}",
```
Monkatraz, line 105: unrelated to PR: ftml probably needs a utility for handling URLs without hardcoding them

ftml/src/settings.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub allow_local_paths: bool,
+}
+
+impl WikitextSettings {
+    pub fn from_mode(mode: WikitextMode) -> Self {
+        match mode {
+            WikitextMode::Page | WikitextMode::Draft => WikitextSettings {
+                mode,
+                enable_page_syntax: true,
+                use_true_ids: true,
```
Monkatraz, line 58: I think drafts need to have `use_true_ids` set to false, because Sheaf's MorphDOM optimization gets very confused by duplicated IDs (which you can do in ftml). Also some feature may exist that renders more than one draft at once (perhaps comparisons) so there is that as well.

Monkatraz (approved):

ftml/src/render/handle.rs
emmiegit: That's why all of it is in `Handle`, which will (eventually) be replaced with calls to PHP.

ftml/src/settings.rs
emmiegit: This is an oversight, I intended to do this but forgot.

emmiegit:
thanks @Monkatraz

PR #538, opened 2021-10-12
[WJ-730] Follow-up: fully clean out EventLogger

emmiegit:
The previous WJ-730 PR missed `php/Utils/EventLogger.php`, which was actually being used in a few places. Tested locally by editing a page.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #539, opened 2021-10-12
Remove xdiff, improve page diffs

emmiegit:
This PR removes `xdiff` entirely, as it is not used. Wikidot just uses the system's `diff` command-line utility instead.

While working on this PR, I stumbled into Wikijump's current diff situation, and greatly improved it:
image

This adds both contextual diffs (only shows parts that changed, with context around them), and changes the inline / word-based diff (the one that was impossible to read) with regular line diffs produced by `diff`.

I think you can agree this version is better than the original:
image


Monkatraz (approved):

emmiegit:
This doesn't actually show line numbers in the diff, but Wikidot doesn't show those in the page source anyways, and in the interest of a short PR I'm just going to merge this here. This is already way better than Wikidot's diff.

emmiegit:
thanks @Monkatraz

PR #541, opened 2021-10-13
[WJ-868] Page view, font adjustments

Monkatraz:
This PR adds:
* A new, shiny Blade page view. It doesn't have any of its interactive elements yet, e.g. page actions, but the rest is there.
* CJK font adjustments, so your browser doesn't download five 10mb fonts trying to find a UI symbol
* A new route, `/page--test`, which generates random pages in varying configurations (very useful for testing)

This PR also made it so that `License` instances are used directly by the view rather than using `licenseHtml`.

image


install/files/dev/wikijump.ini
```diff
@@ -336,6 +336,11 @@ file = wikijump.log
 ; Default: "text_wiki"
 wikitext_backend = "text_wiki"
 
+; [frontend]
+; Purpose: Which views to use for the frontend. Choose either "legacy" or "next".
+; GlobalProperties Reference: #FEATURE_FRONTEND
+; Default: "legacy"
+frontend = "legacy"
```
emmiegit, line 343: Good idea :+1:

web/app/Http/Controllers/PageController.php
```diff
@@ -0,0 +1,102 @@
+<?php
+
+declare(strict_types=1);
```
emmiegit, line 3: Just curious, since I know you have Prettier and lots of plugins, are these new pages formatted using PSR-12?

web/app/Http/Controllers/PageController.php
```diff
@@ -0,0 +1,102 @@
...
+
+      if ($values['site']) {
+        $title = $values['site']->getName();
+      }
+
+      if ($values['wikiPage']) {
+          $page = $values['wikiPage'];
+
+          $sidebar_content = $values['sideBar1Content'] ?? null;
+          $page_content = $values['pageContent'] ?? null;
```
emmiegit, line 40: PHP variables should be `$camelCase`.

web/app/Http/Controllers/PageController.php
```diff
@@ -0,0 +1,102 @@
...
+          $tags = $values['tags'] ?? null;
+          $revision = $page->getRevisionNumber();
+          $timestamp = $page->getDateLastEdited()->getTimestamp();
+          $page_title = $page->getTitleOrUnixName();
+          $title = $page_title . ' | ' . $title;
```
emmiegit: Since this is only composed of static strings and variables, you can use interpolation:
```suggestion
          $title = "$page_title | $title";
```

web/app/Http/Controllers/PageController.php
```diff
@@ -0,0 +1,102 @@
...
+              $lic = $values['category']->getLicense();
+              if ($lic->url()) {
+                  $license = $lic;
+              }
+          }
+
+          if ($values['breadcrumbs']) {
+              $breadcrumbs = [];
+              foreach ($values['breadcrumbs'] as $breadcrumb) {
+                  $breadcrumbs[] = [
```
emmiegit, line 62: Prefer `array_push()` over `$array[] = `.

web/app/View/Composers/PageMockedComposer.php
```diff
@@ -0,0 +1,139 @@
+<?php
```
emmiegit, line 1: Maybe we should move this (and future classes) to `web/app/View/Composers/Mock/`?

web/app/View/Composers/PageMockedComposer.php
```diff
@@ -0,0 +1,139 @@
...
+        $sidebar_content = $this->generateSidebarContent();
+        $navbar_items = $this->generateNavbarItems();
+        $license = $this->getRandomLicense();
+        $breadcrumbs = $this->generateBreadcrumbs();
+
+        $page_title = $f->streetName;
+        $revision = $f->randomNumber(3);
+        $timestamp = $f->dateTimeThisYear->getTimestamp();
+        $tags = $f->unique()->words(10);
+        $title = $page_title . ' | Wikijump';
```
emmiegit: We shouldn't hardcode "Wikijump" anywhere:
```suggestion
        $title = $page_title . ' | ' . GlobalProperties::$SERVICE_NAME;
```

web/resources/views/next/frame.blade.php
```diff
@@ -105,9 +109,15 @@
...
 
-            @isset($license_content)
+            @isset($license)
                 <div id="footer_license" aria-label="{{ __('frame.aria_license') }}">
-                    {!! $license_content !!}
+                    <a href="{{ $license->url() }}">
+                        @if ($license->unless())
+                            {{ __('frame.license_unless', ['license' => $license->name()]) }}
+                        @else
+                            {{  __('frame.license', ['license' => $license->name()]) }}
```
emmiegit, line 118: Bless :pray:

emmiegit (approved):
This is very nice, great work!

web/app/Http/Controllers/PageController.php
Monkatraz: They're formatted by hand. I haven't set automatic formatting up, because I don't write enough of the PHP yet to make decisions like that on a whim, lol

emmiegit: Is it easy enough for you to have the new files formatted with PSR-12? If not it's fine.

web/app/Http/Controllers/PageController.php
Monkatraz: oh boy

Okay, even _our new code_ is not consistent regarding this. `app/Models/Users` is snake case, for example. I guess I'll change it, but we need to setup codesniffer.

web/app/Http/Controllers/PageController.php
Monkatraz: From the [PHP manual](https://www.php.net/manual/en/function.array-push.php):

>  Note: If you use array_push() to add one element to the array, it's better to use $array[] = because in that way there is no overhead of calling a function.

web/app/View/Composers/PageMockedComposer.php
Monkatraz: This code is also incorrect because it should be using the site name. I suppose I could make it random. Also, a bunch of namespaces start with `Wikijump`, so like... we're deep in hardcoding that name.

web/app/Http/Controllers/PageController.php
emmiegit: What? Really? Ugh I hate that syntax I think it's not easy/obvious to read.

web/app/View/Composers/PageMockedComposer.php
emmiegit: Well let me be more precise, in the code, the service name should be easily customizable. The code itself is obviously going to call itself "wikijump".

web/app/View/Composers/PageMockedComposer.php
Monkatraz: I'm not sure how long this class is going to last, actually. The proper way to do this is to setup a `Page` factory that makes genuine pages. This is instead kind of a hack, and as soon as I start to add interactivity it'll behave Very Weirdly.

web/app/Http/Controllers/PageController.php
Monkatraz: ...is that why every variable in PHP starts with `$`? lmao

I'll change the code here and in other places where its relevant

web/app/View/Composers/PageMockedComposer.php
emmiegit: Fair enough, I assumed it was a permanent addition.

web/app/Http/Controllers/PageController.php
Monkatraz: image

Ah. That explains it. The choice of variable capitalization is not in the standard...

emmiegit: Oh huh. Well I've been using camelCase, I'm not opposed to snake_case, but I kind of want things to be consistent so...

Monkatraz: PHP internally uses snake_case, but a bunch of frameworks use camelCase. I think camelCase is more common in Laravel stuff, although I actually prefer snake_case. I think what I might do is with this alone, and set up a linter/formatter to correct this across all of our new code, whichever way we go.

web/app/Http/Controllers/PageController.php
Monkatraz: I will setup formatting with the next PR.

Monkatraz:
thanks @ammongit

PR #542, opened 2021-10-14
Add prettier/plugin-php, run Prettier on all of the non-legacy PHP code

Monkatraz:
This PR adds `@prettier/plugin-php` to our Prettier config. Some of the formatting is a little strange, but in total Prettier cleaned up a lot of little things really well. It not caring for extensive method chaining on a single line is the main thing I've noticed.

You can run Prettier with `pnpm lint`, or have it autofix with `pnpm lint:fix`. The `web/` GitHub workflow will check PHP formatting as well.


web/.prettierrc.yaml
```diff
@@ -11,13 +11,24 @@ plugins:
...
 
 # plugin-jsdoc
 tsdoc: true
 # Set shorter than 90 so if line wrapping is on comments don't go past it
 jsdocPrintWidth: 75
 
+# plugin-php
+phpVersion: "7.4"
+trailingCommaPHP: true
+braceStyle: "psr-2"
```
emmiegit, line 24: PSR-2? The formatting standard we're targeting is PSR-12 right?

emmiegit (approved):

emmiegit:
Looks good, thanks for excluding legacy stuff. It'll be nice having the new code be not only sensible, but also well-formatted!

web/.prettierrc.yaml
Monkatraz: It's just the name of the option, you couldn't set it to `psr-12` if you wanted to. PSR-12 inherits the PSR-2 curly brace style, so I suppose that's why it is named that.

Monkatraz:
thanks @ammongit

PR #544, opened 2021-10-15
Add api--v0 handling, API compliant AuthController, and API mocking fallback

Monkatraz:
Pretty big but fairly easy to understand PR. The most frustrating problem I had was with just figuring out the nightmare that is CSRF, but I eventually got it working. The `WikijumpAPI` class (in JS) has been expanded quite a bit, e.g. it can now handle subdomains.

The next PR will be a shiny new login panel and user account widget. These will use the new authentication routes as well as some mocked routes, like getting your user details.


web/app/Http/Controllers/AuthController.php
```diff
@@ -0,0 +1,105 @@
...
+            'password' => 'required|string',
+            'remember' => 'sometimes|boolean',
+        ]);
+
+        $login = $credentials['login'];
+        $password = $credentials['password'];
+        $remember = $credentials['remember'] ?? false;
+        $isEmail = filter_var($login, FILTER_VALIDATE_EMAIL);
+
+        $attempt = $isEmail
```
emmiegit: Maybe `$success` is better?

web/app/Http/Controllers/AuthController.php
```diff
@@ -0,0 +1,105 @@
...
+    public function logout(Request $request): Response
+    {
+        if (Auth::check()) {
+            Auth::logout();
+            $request->session()->invalidate();
+            $request->session()->regenerateToken();
+            return new Response('', 200);
+        } else {
+            // user isn't logged in, so we can't log them out
+            return new Response('', 400);
```
emmiegit: Shouldn't this be 401?

web/app/Http/Controllers/AuthController.php
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * Gets the authentication state of the client.
+     * Endpoint: `POST:/auth/check` | `authCheck`
+     * @param Request $request The current request.
+     */
+    public function check(Request $request): Response
+    {
+        $session_valid = $request->session()->isStarted();
+        $authed = Auth::check();
+        return new Response(['sessionValid' => $session_valid, 'authed' => $authed], 200);
```
emmiegit, line 81: I feel a little weird returning 200 for false here, but I suppose it makes sense.

emmiegit (approved):

Monkatraz:
thanks @ammongit

PR #545, opened 2021-10-16
Replace is_null() with $x === null.

emmiegit:
This replaces the PHP function `is_null()` with the direct equality check `$x === null` (or `$x !== null`, as appropriate).


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #546, opened 2021-10-16
[WJ-901] Remove diff-based revisions entirely

emmiegit:
Previously in #539 I removed diff-based revision storage since it was commented out / unused and added an assertion on the column to ensure it wasn't being set anywhere. In this PR, I remove those columns entirely, as we are only going to be storing full page sources in our revisions table.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #547, opened 2021-10-16
[WJ-910] Toast!

Monkatraz:
This PR adds toast notification support.

```js
toast("danger", "dangerous toast!", 0)
toast("info", "Toast?", 0)
toast("success", "Toast!", 0)
toast("warning", "oh no... toast", 0)
setTimeout(() => toast("info", "see ya later"), 1000)
```

image

https://user-images.githubusercontent.com/34875062/137595191-7cefc060-2d12-4ff7-b97d-3a5262ce419c.mp4

### DIFF
--- a/web/modules/components/src/Toasts.svelte
+++ b/web/modules/components/src/Toasts.svelte
@@ -1,9 +1,10 @@
 <script lang="ts">
-  import { cubicOut } from "svelte/easing"
   import { toasts } from "./lib/toasts"
+  import { anim } from "./lib/animation"
   import Button from "./Button.svelte"
   import Icon from "./Icon.svelte"
   import { t } from "@wikijump/api"
+  import { tick } from "svelte"
 
   $: listToasts = Array.from($toasts)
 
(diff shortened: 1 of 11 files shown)


web/modules/components/src/Toasts.svelte
```diff
@@ -0,0 +1,164 @@
...
+  }
+</script>
+
+<ul class="toasts" aria-live="polite" aria-relevant="additions">
+  {#each listToasts as toast (toast)}
+    <li
+      class="toast-block"
+      on:introstart={intro}
+      on:outrostart={outro}
+      transition:anim={{
```
emmiegit, line 49: Can you gate fhis behind a prefers-reduced-motion?

web/modules/components/src/lib/toasts.ts
```diff
@@ -0,0 +1,47 @@
...
+  time = 5000
+) {
+  const remove = () => {
+    toasts.update(cur => {
+      cur.delete(toastData)
+      return new Set(cur)
+    })
+  }
+  const toastData = { type, message, remove }
+  toasts.update(cur => new Set(cur.add(toastData)))
```
emmiegit, line 38: I don't understand why it's doing set operations like this to remove a toast.

emmiegit (approved):

web/modules/components/src/lib/toasts.ts
Monkatraz: Svelte's most efficient form of reactivity is based off of strict identity, literally`===`. If I modify the set during a store update, Svelte has no way of knowing that the set actually changed and just assumes it didn't (aka, strict equality comparison).

By replacing the set, Svelte sees that the stores value is different from last time and thus refreshes the UI.

That might be a little difficult to follow, but it's one of those things you get used to.

web/modules/components/src/Toasts.svelte
Monkatraz: I could, and thought about it, but I think it might do more harm than good. Prefers reduced motion is usually related to constant, rhythmic motion that can be aggravating.

In this case, these animations are important in preventing the toasts from suddenly jumping when one disappears. This can easily get confusing and make you lose track of which toast just disappeared and which ones moved. By having an animation, you'll be glancing at the toasts rather quickly, well before the animation completes.

Monkatraz:
thanks @ammongit

PR #551, opened 2021-10-17
[WJ-914] Add localization function to PHP

emmiegit:
This is part 1 for an issue to add ICU localization capability to our PHP codebase.

The PR adds a `__` function, similar to Laravel's, but using our translation stack. It needs to be initialized before dependencies are so Laravel won't add its own helper function of the same name.

It also moves `colors()` to `config/` since it's not really a global helper function as it as constant data to receive.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #556, opened 2021-10-18
[WJ-909] Login/register modal/panel

Monkatraz:
Since this PR is getting kinda big, I'm going to do the login view template and route in another PR. I need to update the localization messages first, anyways.

image

image

https://user-images.githubusercontent.com/34875062/137738674-5a7a702a-644e-40bd-b506-1caf2e838090.mp4


locales/en.yaml
```diff
@@ -46,13 +46,47 @@ components:
...
+  LOGIN: Login
+  REGISTER: Register
+  SPECIFIER: Email or Username
+  SPECIFIER_PLACEHOLDER: Enter email or username...
+  EMAIL: Email
+  EMAIL_PLACEHOLDER: Enter email address...
+  EMAIL_INFO: Your email address is private.
+  USERNAME: Username
+  USERNAME_PLACEHOLDER: Enter username...
+  USERNAME_INFO: You'll be able to change this later.
```
emmiegit, line 71: Though one note about this is we don't allow unlimited changes like say Twitter because we have past aliases for renames.

web/modules/components/src/Sprite.svelte
```diff
@@ -0,0 +1,30 @@
...
+
+  let viewBox = ""
+  // prettier-ignore
+  switch (i) {
+    case "wj-karma":             viewBox = "0 0 64 114" ; break;
+    case "wj-toggle":            viewBox = "0 0 128 64" ; break;
+    case "wj-checkbox":          viewBox = "0 0 64 64"  ; break;
+    case "wj-clipboard-copy":    viewBox = "0 0 24 24"  ; break;
+    case "wj-clipboard-success": viewBox = "0 0 24 24"  ; break;
+    case "wj-downarrow":         viewBox = "0 0 26 18"  ; break;
```
emmiegit, line 24: What does this do?

web/modules/components/src/TextInput.svelte
```diff
@@ -0,0 +1,198 @@
...
+    <!-- special case: input is a password type -->
+    <!-- when this happens, we'll turn the icon into a "show password" button -->
+    {#if input?.type === "password"}
+      <!-- prettier-ignore -->
+      <span
+        class="textinput-icon is-password"
+        use:tip={$t("components.textinput.SHOW_PASSWORD")}
+        aria-hidden="true"
+        use:whileHeld={{
+          pressed: () => { if (input) input.type = "text" },
```
emmiegit, line 73: Not a toggle? Might this be an accessibility issue?

emmiegit (approved):

web/modules/components/src/TextInput.svelte
Monkatraz: You can't even see this button with a screen reader, but the reason I think that's fine is because a screen reader or other a11y utilities will just read out the password field, they don't care about the browser's visual obscuring thing

web/modules/components/src/Sprite.svelte
Monkatraz: In addition to the Iconify icons, Wikijump has a spritesheet with a bunch of configurable icons/sprites. The new toggle component uses it for example. Karma is also part of the spritesheet.

web/modules/components/src/TextInput.svelte
emmiegit: Is there a comment somewhere explaining that?

Monkatraz: I don't think it really needs one, considering the element has `aria-hidden` and the element isn't focusable.

web/modules/components/src/Sprite.svelte
Monkatraz: Oh but if you meant the viewBox thing in particular, it sets the "intrinsic size" and aspect ratio of the SVG, other wise it would be some horrible huge shape

Monkatraz:
thanks @ammongit

PR #557, opened 2021-10-18
Update locale messages

Monkatraz:
This migrates/updates all of the new frontend's locale messages to the `/locales` system we have. I left the old Jetstream `web/resourceslang/en/...` files alone, but eventually they'll get deleted once we remove Jetstream.


web/app/Services/License/License.php
```diff
@@ -37,16 +37,9 @@ private function buildHtml(): string
...
-            $prefix = __(
-                'Unless stated otherwise, the content of this page is licensed under',
-            );
-        } else {
-            $prefix = __('The content of this page is licensed under');
-        }
+        $prefix = $this->unlessClause ? 'Unless stated otherwise, the' : 'The';
 
-        return "$prefix $link.";
+        return "$prefix content of this page is licensed under $link.";
```
emmiegit, line 42: Wait, why isn't this using the localization function anymore?

emmiegit (approved):

web/app/Services/License/License.php
Monkatraz: The new frontend doesn't use this at all, and I didn't want to add a new localization key for legacy content

emmiegit: I see

Monkatraz:
thanks @ammongit

PR #558, opened 2021-10-19
[WJ-914] Unify localization in PHP

emmiegit:
This PR is part 2 of [WJ-914](https://scuttle.atlassian.net/browse/WJ-914), see part 1 here: #551.

The PR removes system `gettext` from the container, instead using the PHP dependency [`gettext/gettext`](https://github.com/php-gettext/Gettext) to manually load the appropriate language files via `MoLoader`. The localization files are copied into `web/public/files--built/locales/`. The legacy Smarty helper is adjusted, and an alias `_()` is added until we can remove all the old localization code.

----

For local testing, I added `TEXT: {{ __('components.theme_picker.TOOLTIP') }}` to a Blade template. Results below:

Default (`en`):
image

Switched to `en_GB`:
image


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #559, opened 2021-10-19
[WJ-914] Remove old language files

emmiegit:
Now that we have our localization system connected to PHP (#558), we can remove Laravel's language files since we are not using their localization system. (Note: the `__` function now uses our localization method, Laravel doesn't add their helper since it sees the function is already defined).


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #564, opened 2021-10-20
[WJ-182] [WJ-312] [WJ-905] Gesture controlled sidebar

Monkatraz:
This PR got kinda large, sorry. This PR adds:
* A gesture controlled sidebar. Gesture controls enable and disable automatically depending on screen size.
* Overhauled the gesture library stuff to use classes rather than weird closures
* Fair few misc. CSS tweaks in order for mobile page rendering to be serviceable (more will come later)

Because of how messy the DOM/user-input related utility functions/classes are currently, with them being spread between `/util` and `/components`, I'm going to be making another PR very shortly that makes a new module called `/dom`, which will hold all of these various bits and pieces.


web/modules/components/src/lib/gesture.ts
```diff
@@ -1,118 +1,261 @@
...
+      : "horizontal"
+  }
+
+  /**
+   * The "sign" of the gesture, as in multiplying a pixel offset by this
+   * value will move an element in the direction of this gesture (assuming
+   * the axis is correct).
+   */
+  get sign(): -1 | 1 {
+    return this.direction === "up" || this.direction === "left" ? -1 : 1
```
emmiegit, line 101: Do we need to add parentheses around the two or conditions?

web/resources/css/abstracts/_variables.scss
```diff
@@ -0,0 +1,14 @@
+/// Above content z-index.
+$z-above: 100;
+
+/// Popover z-index, e.g. dropdowns.
+$z-popover: 200;
+
+/// Panel z-index, e.g. sidebar.
+$z-panel: 300;
```
emmiegit, line 8: Dropdowns are below sidebars?

emmiegit (approved):

emmiegit:
Wow this is really big. Shame more gesture stuff isn't native to browsers.

web/modules/components/src/lib/gesture.ts
Monkatraz: Good catch, but apparently not. Prettier removes any unnecessary parenthesis, and if I add any there it removes them.

web/resources/css/abstracts/_variables.scss
Monkatraz: "Panels" are floating UI elements disconnected from the main page, but your usual dropdown is some sort of interactive element _on_ the main page. Not sure if that made any sense, but you would want the floating sidebar to appear above the navbar dropdowns.

Monkatraz:
> 
> 
> Wow this is really big. Shame more gesture stuff isn't native to browsers.

Yeah, it's really frustrating. This is why so many mobile sites have basically no gesture support.

Monkatraz:
thanks @ammongit

PR #567, opened 2021-10-21
Mark legacy directories as linguist-vendored

Monkatraz:
I think it's a little unfair to include the ginormous pile of legacy Wikidot code in our project's linguist data. Vendored files will still get included in diffs, it's just that we won't see this anymore:
image

Smarty is 6.8%. That's painful.

### DIFF
--- a/web/.gitattributes
+++ b/web/.gitattributes
@@ -1,6 +1,11 @@
 *                      text=auto
-web/**                 linguist-vendored
 pnpm-lock.yaml         binary linguist-vendored linguist-generated
 **/vendor/**           binary linguist-vendored linguist-generated
 *.json                 linguist-language=JSON-with-Comments
 CHANGELOG.md           export-ignore
+
+bin/**                 linguist-vendored
+lib/**                 linguist-vendored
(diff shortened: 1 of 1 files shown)


emmiegit (changes requested):
I disagree with this change, the language bar should get cut as we go and actually delete legacy files.

In this way it's actually valuable for seeing where our repo is at in terms of remaining legacy nonsense to cut, and a sense of accomplishment when we can submit a PR to delete old garbage. The smarty stuff for instance will disappear when we can make the final switchover to the new frontend, and the % of PHP will continue to shrink as we delete legacy code and rewrite stuff in other services.

Monkatraz:
Interesting, considering `web/**` was already ignored. But alright.

emmiegit:
Web was being ignored? I wasn't aware of that.

Monkatraz:
Yeah, it is. Well, actually, it was originally `*.js` and `*.css`, but when we moved `client/` into `web/`, I changed it to `web/**` (as that was the intention of ignoring JS and CSS files).

PR #568, opened 2021-10-22
[WJ-923] Login/register views

Monkatraz:
This PR implements the login and register views. I also added a "plain" mode for the `next.frame` view, but I didn't actually use it in this PR. When I implement some sort of settings view it'll get used.

image
image


web/resources/scripts/auth-login.ts
```diff
@@ -0,0 +1,13 @@
...
+import "./auth.scss"
+
+window.addEventListener("DOMContentLoaded", () => {
+  const container = document.getElementById("auth_form_container")
+  if (!container) throw new Error("No container found")
+  const form = new LoginForm({ target: container })
+
+  form.$on("login", () => {
+    // redirect to home
+    location.href = "/"
```
emmiegit, line 12: Is it difficult to redirect back to where the user was previously?

emmiegit (approved):

web/resources/scripts/auth-login.ts
Monkatraz: Short answer: yes.

Long answer: I could do it via Laravel, but that means giving the JS the previous page somehow. I can technically do it with _just_ JS, but that means literally just making the user go back a page in their history. That means they could go forward again back into the login page, which is a bit weird. I've added TODOs for redirecting back to where the user came from.

Monkatraz:
thanks @ammongit

PR #570, opened 2021-10-22
Update to edition 2021

leo60228:
The new resolver led to some harmless changes in selected features, but otherwise ftml is completely unaffected by the breaking changes.


emmiegit:
> The new resolver led to some harmless changes in selected features,

Can you clarify?

leo60228:
One of the changes in the new edition means that, for example, features selected by build dependencies don't affect the final binary. This seems to have affected ftml, as several dependencies were rebuilt after I changed the edition, but didn't cause any issues.

emmiegit:
Okay apparently I can't push to your branch so I'm just going to merge for now and make a separate PR with my changes.

emmiegit (approved):

PR #573, opened 2021-10-23
Update PHP FFI for ftml

emmiegit:
After wikitext settings were added to ftml, the PHP was not updated to pass in that information. This PR adds a PHP version of `WikitextSettings`, converts from `ParserRenderMode` to it, and passes that data to libftml. It also changes some variable names to `$snake_case`.

Side note: This is one of the things I don't like about PHP FFI. There's no way for the build to automatically fail if the C part of the FFI changes at all.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #575, opened 2021-10-23
Fix debug segfault in PHP-FFI for ftml

emmiegit:
Previously, xdebug would try to access invalid memory when ftml calls were made (even though the FFI calls themselves don't do anything unsafe). This was determined to be an artifact of how Rust allocates zero-length vectors. Since a `NULL` pointer isn't used, xdebug felt it was okay to read through the pointer, unaware that it was meant to have a length of zero (i.e. has no valid elements). By changing the FFI handling to always emit `(NULL, 0)` for zero-length vectors, this crash no longer occurs.

This PR also changes some other things, such as only installing xdebug on local, and adding some helpful debug utilities in the local containers.

### DIFF
--- a/ftml/src/ffi/vec.rs
+++ b/ftml/src/ffi/vec.rs
@@ -27,6 +27,13 @@ pub unsafe fn cptr_to_slice<'a, T>(ptr: *const T, len: usize) -> &'a [T] {
 }
 
 pub fn vec_to_cptr<T>(vec: Vec<T>) -> (*mut T, usize) {
+    // Special handling for empty lists:
+    // Always report pointer as NULL, discard vector like normal.
+    if vec.is_empty() {
+        return (ptr::null_mut(), 0);
+    }
+
+    // Otherwise, get raw aspects of Vec for export
     let mut slice = vec.into_boxed_slice(); // shrinks capacity to len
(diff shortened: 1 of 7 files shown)


ftml/src/ffi/vec.rs
```diff
@@ -42,6 +49,12 @@ pub unsafe fn drop_cptr<T, F>(ptr: *mut T, len: usize, drop: F)
 where
     F: FnMut(T),
 {
+    // Special handling for empty lists,
+    // since it's always (NULL, 0).
+    if ptr.is_null() {
+        return;
+    }
+
     Vec::from_raw_parts(ptr, len, len).drain(..).for_each(drop);
```
lambda-fairy, line 58: Not related to this PR, but: `drop` is a function in the Rust standard library. I'd suggest renaming the parameter to clarify that it's a custom function not the standard one.

Monkatraz (approved):

ftml/src/ffi/vec.rs
emmiegit: I guess. I need it drop since that's what Rust destructors are called. I'll think about it.

emmiegit:
thanks @Monkatraz @lambda-fairy

PR #576, opened 2021-10-23
[WJ-925] Build 'default' locale

emmiegit:
The "default" locale is one where each translation message is just the message key. It is similar to the template in this way, but is usable as an actual localization in various debugging situations. This PR adds automatic generation of the default locale to the `locales` build.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #577, opened 2021-10-23
[WJ-928] Authentication overhauling, password confirmation view

Monkatraz:
This PR was _meant_ to be about adding all the views Fortify needed, e.g. password reset, but I realized that ultimately we need to remove Fortify so that we can roll our own controllers and views.

The next PR will probably handle that + removing Jetstream. That PR won't _replace_ Fortify, it will just remove it. That means you won't be able to register, reset the password, etc, but we weren't doing that anyways.

image


locales/en.yaml
```diff
@@ -131,6 +133,7 @@ account_panel:
     REGISTER_FAILED: Failed to register account. Check your email, username, and password.
     EMAIL_TAKEN: Email already in use.
     INVALID_INPUT: You need to fill out all required fields.
+    CONFIRM_FAILED: Wrong password. Check what password you have entered.
```
emmiegit:
```suggestion
    CONFIRM_FAILED: Wrong password. Check the password you have entered.
```
?

web/app/Helpers/helpers.php
```diff
@@ -30,3 +30,27 @@ function _(string $key): string
         return __($key);
     }
 }
+
+if (!function_exists('previousUrl')) {
```
emmiegit: imo this isn't super necessary. We have the function guard for `_` since it's both temporary and possibly provided by gettext, but this helper is something of our own, and also we're a client not a framework, so I'm not sure who would override this.

web/app/Helpers/helpers.php
```diff
@@ -30,3 +30,27 @@ function _(string $key): string
...
+    /**
+     * Returns the previous URL for the session.
+     * Will return an empty string if the previous URL is the same as
+     * the current one.
+     */
+    function previousUrl(): string
+    {
+        $url = url()->previous();
```
emmiegit: What is `url()`?

web/app/Http/Controllers/AuthController.php
```diff
@@ -4,48 +4,95 @@
...
 /**
  * Controller for authenticating users.
  * API: `/auth`
  */
 class AuthController extends Controller
 {
+    /** Guard used to handle authentication. */
+    private StatefulGuard $guard;
+
+    /** @param StatefulGuard $guard */
```
emmiegit:
```suggestion
    /**
     * @param StatefulGuard $guard
     */
```

web/app/Http/Controllers/AuthController.php
```diff
@@ -4,48 +4,95 @@
...
-            'remember' => 'sometimes|boolean',
-        ]);
+        // atempt to get the user for the given credentials
+        $result = Authentication::authenticate($request);
 
-        $login = $credentials['login'];
-        $password = $credentials['password'];
-        $remember = $credentials['remember'] ?? false;
-        $is_email = filter_var($login, FILTER_VALIDATE_EMAIL);
+        if (!$result->ok()) {
```
emmiegit: I would prefer a positive check here. Like Rust `Result` has `is_ok()` and `is_err()` as methods that can be invoked.

web/app/Services/Authentication/AuthenticationResult.php
```diff
@@ -0,0 +1,80 @@
...
+use Wikijump\Models\User;
+use Wikijump\Common\Enum;
+
+/**
+ * Not an actual exception.
+ * Enumeration of different potential errors during an authentication attempt.
+ */
+final class AuthenticationError extends Enum
+{
+    const FAILED_TO_VALIDATE = 'failed_to_validate';
```
emmiegit: I think I prefer kebab-case for string enum values.

web/resources/lib/components/AccountPanel.svelte
```diff
@@ -3,12 +3,22 @@
...
-  import { Button, Tabview, Tab } from "@wikijump/components"
+  import { Button, Tabview, Tab, toast } from "@wikijump/components"
   import LoginForm from "./LoginForm.svelte"
   import RegisterForm from "./RegisterForm.svelte"
 
   /** Gets passed to this panel if it's been mounted inside of a dialog. */
   export let closeDialog: (() => void) | undefined
+
+  function login() {
+    closeDialog?.()
```
emmiegit, line 14: What does this syntax mean?

emmiegit (approved):

web/resources/lib/components/AccountPanel.svelte
Monkatraz: It's the optional chaining operator, which short circuits object accessing when a property is nullish. Interestingly, they allowed you to use the operator to optionally call functions. So, if `closeDialog` isn't nullish, it'll be called, but otherwise it does nothing.

Optional chaining is so useful that it used to be a major selling point of TypeScript, as it had optional chaining before anything else did.

web/app/Services/Authentication/AuthenticationResult.php
Monkatraz: Ah, yeah, I agree, oops

web/app/Helpers/helpers.php
Monkatraz: It's Laravel's global for well, URLs. I should use the `URL` namespace instead.

web/app/Helpers/helpers.php
Monkatraz: I actually added this because in other Laravel packages, and in Laravel itself, they guard every global this way. I don't know why, I agree that it _seems_ unnecessary, but I'm hesitant to be different to everyone else, lol. Should I remove it?

web/app/Http/Controllers/AuthController.php
Monkatraz: Okay, AuthenticationResult now has the methods `isOk()` and `isErr()` to match Rust, and so now this check uses `isErr()`.

web/app/Helpers/helpers.php
emmiegit: They do that because Laravel is a framework, and the end user might want to use that function name for something else (like we do with `__`).

Monkatraz: Alright, I removed it.

Monkatraz:
thanks @ammongit

PR #578, opened 2021-10-23
Handle falling back to a default set of favicons

Monkatraz:
Sike, I'm not removing Fortify yet! This PR is simple and it adds falling back to default Wikijump favicons if you don't provide any to the view template. There is also some fancier handling of the resulting favicon `<link>` elements depending on what combination of favicons you give it.


emmiegit (approved):

Monkatraz:
thanks @ammongit

PR #579, opened 2021-10-24
[WJ-915] Begone Jetstream

Monkatraz:
This PR removes Jetstream. Other than the deletion of a bajillion files, this PR transfers avatar handling from Jetstream over to our own code. It does basically the same thing as Jetstream.


web/app/Models/User.php
```diff
@@ -145,6 +193,7 @@ protected function defaultProfilePhotoUrl(): string
             $initials .= $initial[0];
         }
 
+        // TODO: replace this with a local solution
```
emmiegit, line 196: We could easily self-host the ui-avatars code, but it seems easier to just delegate to them.

emmiegit (approved):

web/app/Models/User.php
Monkatraz: Don't agree - this causes crossorigin issues and it doesn't even work in local right now because of it. Plus it just "feels wrong", lol

emmiegit: Okay, so I guess we can copy/adapt their source for a route locally? We could also just have these parameters baked into a config instead of needing to pass them each time.

Monkatraz: There are a lot of ways to go about this, imo. Generation, default image, using HTML, etc. Let's leave that to another PR.

Monkatraz:
thanks @ammongit

PR #580, opened 2021-10-24
Add pure bash implementation of migration script

emmiegit:
This adds a script to create new migrations for PHP, but without requiring `artisan` or any of the other aspects of the PHP ecosystem on your local machine. (Since all it's really doing is creating a templated file). It can be run in any directory, and will create new migration file(s) based on the name you give it, similar to `php artisan make:migration <name...>`.

### DIFF
--- a/web/scripts/create-new-migration.sh
+++ b/web/scripts/create-new-migration.sh
@@ -37,7 +37,7 @@ function check_name() {
 	fi
 }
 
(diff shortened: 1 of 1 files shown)


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #585, opened 2021-10-25
Use lazy loading in LocalizationService.

emmiegit:
When running artisan commands locally this error is common:
```
> Illuminate\Foundation\ComposerScripts::postAutoloadDump
> @php artisan package:discover --ansi

   ErrorException 

  filesize(): stat failed for ~/git/wikijump/web/public/files--built/locales/en.mo

  at vendor/gettext/gettext/src/Loader/Loader.php:48
     44▕      * Reads and returns the content of a file.
     45▕      */
     46▕     protected static function readFile(string $file): string
     47▕     {
  ➜  48▕         $length = filesize($file);
     49▕ 
     50▕         if (!($fd = fopen($file, 'rb'))) {
     51▕             throw new Exception("Cannot read the file '$file', probably permissions");
     52▕         }

      +2 vendor frames 
  3   app/Services/Localization/LocalizationService.php:27
      Gettext\Loader\Loader::loadFile()

  4   app/Services/Localization/LocalizationService.php:20
      Wikijump\Services\Localization\LocalizationService::loadTranslations()

```
This is because the expected localization files aren't present, as it's part of the build process. The PR fixes this by lazily loading localizations when the first translation message is requested rather than unconditionally at program load.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #586, opened 2021-10-25
[WJ-928] Begone Fortify

Monkatraz:
This PR removes Laravel Fortify. Fortify is scaffolding for authentication controllers and other API endpoints. We are doing our own API endpoint thing, so we shouldn't be using Fortify.

I added a class, `UserValidation`, which is presently unused. However, it contains methods for validating user data that _will_ be used later. I made this class now because I was deleting a bunch of Fortify actions and I wanted to keep the validation code from those actions.


web/app/Models/User.php
```diff
@@ -20,13 +20,18 @@
...
 use Wikijump\Helpers\InteractionType;
 use Wikijump\Traits\HasInteractions;
 use Wikijump\Traits\HasSettings;
 use Wikijump\Traits\LegacyCompatibility;
 
+// TODO: add two factor authentication
+// see how Fortify does it - it's not too complicated
+// relevant fields in model:
+//   two_factor_secret
+//   two_factor_recovery_codes
```
emmiegit: Since we're not using Fortify we can probably go ahead and name these `multi_factor_xxx` since that's technically more accurate and we couldn't do that before because it has specific names.

emmiegit (approved):

web/app/Models/User.php
Monkatraz: It Has Been Done (thanks for the migration creator script)

Monkatraz:
thanks @ammongit

PR #562, opened 2021-10-19
[WJ-471] Add PHP ImageMagick extension

emmiegit:
This adds the PHP extension `imagick`, which is needed for MFA barcode generation. I also sort the list of `apt` dependencies for consistency.

PR #588, opened 2021-10-27
[WJ-471] Remove PHP ImageMagick extension

emmiegit:
Reverts scpwiki/wikijump#562

It turns out we don't actually need this extension to generate SVG MFA barcodes, and since we don't need imagemagick for other parts of the codebase, and every PHP extension we add massively slows down the build process, we should remove this.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #589, opened 2021-10-27
[WJ-920] Replace Ozone page link tables

emmiegit:
This is the first part of [WJ-920](https://scuttle.atlassian.net/browse/WJ-920), which starts replacing page tables. Here we replace the page link tables. Now we have the following:

* `page_link` -- represents links to external URLs (strings)
* `page_connection` -- represents generic relations between two pages, with an enum for link, include, etc
* `page_connection_missing` -- same as above, but for wanted/missing pages. instead of a page ID it takes the slug instead

These tables also store nice things like the `count` of how many links. The PR also has changes like refactoring the outdating code to be less duplicated and more generic. Along the way I found silly typo-ish bugs in ftml regarding backlinks data.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #592, opened 2021-10-29
Container madness (improved handling of Vite and Docker)

Monkatraz:
The diff count is due to me updating dependencies.

Main takeaways:
* The `php-fpm` container no longer needs to build client assets.
* Builds are quite a bit faster, as `node:alpine` is used instead of `circleci/node`.
* The Traefik container is no longer used for local.
* Vite's development assets are now proxied by nginx, meaning that, to your browser, everything is on the same domain. This has a bunch of other benefits.
* A build failure will stop the `pnpm dev` script.


web/app/Providers/AppServiceProvider.php
```diff
@@ -14,6 +16,28 @@ class AppServiceProvider extends ServiceProvider
...
+        $vite_manifest_path = config('vite.build_path') . '/manifest.json';
+
+        if (!file_exists(public_path($vite_manifest_path))) {
+            // get manifest from nginx container
+            // we have to catch a potential exception here, because
+            // nginx may or may not be running.
+            // also, if the manifest isn't there laravel-vite will just
+            // throw its own error, so we'll just let it handle that
+            try {
+                $file = file_get_contents("http://nginx:80/$vite_manifest_path");
```
emmiegit, line 32: Does the `php-fpm` container depend on `nginx`?

web/app/Providers/AppServiceProvider.php
```diff
@@ -14,6 +16,28 @@ class AppServiceProvider extends ServiceProvider
      */
     public function register()
     {
+        // we don't build the client assets in the php-fpm container,
+        // so we'll need to retrieve the manifest for them from the
+        // nginx container
+
+        $vite_manifest_path = config('vite.build_path') . '/manifest.json';
+
+        if (!file_exists(public_path($vite_manifest_path))) {
```
emmiegit, line 25: Should this be in boot not register?

web/app/Providers/AppServiceProvider.php
```diff
@@ -14,6 +16,28 @@ class AppServiceProvider extends ServiceProvider
...
+            // throw its own error, so we'll just let it handle that
+            try {
+                $file = file_get_contents("http://nginx:80/$vite_manifest_path");
+
+                if ($file) {
+                    file_put_contents(public_path($vite_manifest_path), $file);
+                    Log::debug('Wrote Vite manifest file (retrieved from nginx)');
+                }
+            } catch (Exception $err) {
+            }
```
emmiegit, line 39: Why is there an empty catch here? Either the error should surface, or we should log if we're fine this not succeeding.

web/scripts/dev.js
```diff
@@ -84,10 +84,6 @@ const { createServer } = require("vite")
     server.close()
     legacyShell.kill()
     composeShell.kill()
-
-    // deals with an annoying "Terminate batch script? (Y/N)"
-    // prompt on Windows. fun fact: this only works _sometimes_
-    if (process.platform === "win32") process.exit(0)
```
emmiegit, line 90: What changed here?

emmiegit (approved):

web/app/Providers/AppServiceProvider.php
Monkatraz: No, but it doesn't need to. This will just get retried on every request until `nginx` is up.

web/app/Providers/AppServiceProvider.php
Monkatraz: That would make sense, yeah, but I wanted to make sure this got ran before anything else did (aka, before `laravel-vite` could possibly need the manifest). I'm not sure if I actually need to do this, but it's a "just in case" thing.

web/app/Providers/AppServiceProvider.php
Monkatraz: Keep in mind that this gets ran on every request - so if it fails, it'll just get tried again. Additionally, if the manifest really just isn't there, or can't be written, `laravel-vite` will throw its own errors. So, basically, I _am_ surfacing the error.

web/scripts/dev.js
Monkatraz: I just don't think it works, lol. I've seen no difference. Plus, it probably does bad things to non-`cmd.exe` command prompts (e.g. Powershell).

web/app/Providers/AppServiceProvider.php
emmiegit: In that case please leave a comment. I'm not comfortable with empty catch blocks since it's hard to tell what's going on there.

Monkatraz: But... I did? There is a big comment above the `try` block.

emmiegit: Oh I see what you mean

Monkatraz:
thanks @ammongit

PR #600, opened 2021-10-31
Rename "unix_name" to "slug" in user table.

emmiegit:
"Unix name" is not a standard term and is a bit confusing, since unix paths aren't restricted to lowercase alphanumerics and dashes like Wikidot slugs are.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #601, opened 2021-10-31
[WJ-929] Use alpine container for Postgres

emmiegit:
The PostgreSQL Alpine base image is lighter than the default (I believe Ubuntu) one, which will be faster and save money in deployment.


install/local/dev/postgres/Dockerfile
```diff
@@ -7,22 +7,22 @@ ARG ENVIRONMENT="local"
...
 WORKDIR /src
 
 # Create system user
-RUN useradd wikijump --create-home --shell /bin/bash
+RUN adduser -S wikijump
 
 # Setup database
 USER postgres
-RUN pg_createcluster --user postgres --locale en_US.UTF-8 "${POSTGRES_VERSION}" "${POSTGRES_CLUSTER}"
+RUN initdb -D ${PGDATA} --locale=en_US.UTF-8
```
Monkatraz: Out of curiosity, why does a database have a locale? lol

Monkatraz (approved):

install/local/dev/postgres/Dockerfile
emmiegit: Mostly for string encoding

emmiegit:
Because of very bizarre build issues, I'm setting all our CI builds to use `DOCKER_BUILDKIT`.

Essentially, when running a build without it, the step after any filesystem changes were made in the database image were immediately reverted. You could `&& ls ${PGDATA}` to see that the directories had been created, but the step right after it with `RUN ls ${PGDATA}` showed no changes.

emmiegit:
thanks @Monkatraz

PR #607, opened 2021-11-02
[WJ-920] Add page contents table to replace sources and compiled

emmiegit:
This adds a new `page_contents` table which stores source wikitext and compiled HTML, replacing Ozone tables `page_source` and `page_compiled`. Unfortunately because Wikidot has the `Outdater` system, all insertions into this table require the compiled data to be blank and updated later. But there are also some other artifacts of the separate source and compiled tables system that are left over, after all the page tables are migrated I want to clean up this whole process because it's very messy and there's a lot to improve on.

Noteworthy: Because Laravel and Ozone have separate database transactions, all insertions to the new table have to be _after_ the `$db->commit()` call, or the foreign key constraints won't match. There isn't much I can do right now, when the aforementioned restructure happens we can move those back to the correct places.

I've checked page edits, page creation, page deletion, tag edits, and they all work as expected. I also fix the tag cloud (or at least get it not to error anymore).

I also removed `DependencyFixer` since we don't really want its functionality and it's a mess, as well as previews, which are not used because they're hacky and simplistic, and in the code they're a mess. The Ozone framework's DB template files are deleted, since we're trying to get rid of all those model files anyways. I also fix a longstanding Ozone typo (`$fieldName` misspelled as `$filedName`). I removed `OzoneTest.php` because my changes would've required it to be rewritten, and also since we don't actually run it anyways.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #609, opened 2021-11-04
Set locales builder encoding to UTF-8 explicitly

emmiegit:
Previously this service had a disabled lint for setting encoding. Given that this handles international characters, and an assumption that UTF-8 will be used is not portable, this PR changes it to explicitly specify the encoding for all file I/O.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #612, opened 2021-11-06
[WJ-920] [WJ-745] [WJ-766] [WJ-710] Remove Ozone abuse reporting

emmiegit:
This removes Wikidot's old method of abuse reporting. I originally considered a new table for this, but since this is a lower priority item, and this is largely Ozone nonsense, I think for now the best course is just to remove it. We can implement proper abuse reporting later, probably in `1.1` or `1.2`.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #619, opened 2021-11-11
Fix: Require DOCKER_BUILDKIT in pnpm build commands

emmiegit:
Because `DOCKER_BUILDKIT=1` is not set by default, builds fail due to the previously-mentioned annoying postgres container persistence issue. This change modifies `nabs.yaml` to set the environment variable each time to ensure that builds succeed.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #624, opened 2021-11-15
[WJ-936] Remove section editing

emmiegit:
image

Wikidot allows you to edit in sections, like MediaWiki. This is intended to be helpful feature, by editing only a part of a page, as separated by headings.

This has a few problems:
* Wikidot headings are not used quite in the same way as they are in MediaWiki.
* Most users do not use section editing when engaging with pages.
* The feature is buggy and inconsistent, and does not work for some pages at all.
* It complicates the page edit code.

As such, it has been removed.


Monkatraz (approved):
lol I didn't even know this was a feature

I also don't understand why it would ever be a feature considering Wikidot doesn't have a wikitext AST...

emmiegit:
There's also an append mode. It's at least simpler than this, but I also want to remove it for similar reasons.

The section editing on MediaWiki is genuinely very nice! But yeah it completely doesn't make sense on Wikidot.

emmiegit:
thanks @Monkatraz

PR #635, opened 2021-11-29
[WJ-938] Remove append mode

emmiegit:
This removes the Wikidot feature of editing pages in "append mode", where contents are simply added to the end. Similar to the remove of editing by sections, this reduces page edit complexity and a feature which is not well-used.

See also #624


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #641, opened 2021-12-02
[WJ-939] Add experimental separate API service

emmiegit:
This PR adds the experimental API service DEEPWELL (named for the reality alteration-resistant database used by the Foundation). This service will hold joint custody over the database with PHP, until tables can be transferred to it to be held as sole custody. It serves a REST API on port 2747 which will enable interfacing with the database.

Presently this is limited to simple CRUD, but in the future will be extended to lower-level logical operations such as page operations.

The goal is for complex logic which currently lives solely in PHP (e.g. `Outdater.php`) to be able to migrate here as we perform the rest of the database table refactoring tasks. This is also an opportunity to transfer those tables to sole custody here, if appropriate development-wise. If not, dual custody allows Laravel to also access the data itself via Eloquent.

This could also simplify ftml interactions, since presently the need for a "remote handle" is complicated because access to the database requires going through PHP, and FFI has proven to be difficult. If ftml invocation is moved here, then a native handle implementation can be written, the FFI code (in Rust and PHP) can be removed, and the associated latency will also be removed.


deepwell/src/locales.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /// Returns the translation for the given message key.
+    /// If no translation exists, then the message key itself is returned.
+    pub fn translate<'a>(&'a self, locale: &'_ str, key: &'a str) -> &'a str {
+        match self.catalogs.get(locale) {
+            Some(catalog) => catalog.gettext(key),
+            None => key,
+        }
+    }
```
Monkatraz, line 74: Translation strings have placeholders (e.g. like `$t('foo.BAR', { foo: 'foobar' })`), so this is technically an incomplete implementation. I don't know if there is an ICU message library for Rust, but I'd be surprised if there isn't one.

deepwell/src/methods/user.rs
```diff
@@ -0,0 +1,254 @@
...
+        pronouns: Set(None),
+        dob: Set(None),
+        bio: Set(None),
+        about_page: Set(None),
+        avatar_path: Set(None),
+        created_at: Set(Some(now())),
+        updated_at: Set(None),
+        deleted_at: Set(None),
+        ..Default::default()
+    };
```
Monkatraz, line 86: Just for the sake of my understanding: why is every field a `Set`?

Monkatraz (approved):

deepwell/src/methods/user.rs
emmiegit: It is `Set` as opposed to `Unset`, it's a way of telling the active model that I'm only updating these fields, the rest should be left as-is.

deepwell/src/locales.rs
emmiegit: Oh yeah I forgot about that. I need to find an equivalent to the `MessageFormatter` in PHP.

If you don't mind, I'll do this in a follow-up PR.

emmiegit:
thanks @Monkatraz

PR #648, opened 2021-12-09
Update NPM dependencies, recompile FTML binary

Monkatraz:
Does what it says on the tin. Chores, basically.

There are some fancy new stuff that comes with the various dependency updates, mostly notably TypeScript 4.5. TypeScript 4.5 has this, which I'll be using from now on:
```ts
import { type foo, bar } from "foopkg"
```
I'll look later to see if there is an ESLint rule that we can use to upgrade all of the old import syntax later.


emmiegit (approved):

Monkatraz:
thanks @ammongit

PR #652, opened 2021-12-10
[WJ-941] [WJ-933] MJML email templates and rendering

Monkatraz:
This PR adds support for using MJML based email templates, along with the base email templates all other email templates will inherit from. MJML is rendered via the shell in PHP, using the MRML Rust utility.

Email templates can be written in MJML, Markdown, basic Blade, and plain text. Only MJML and plain text have base email templates. MJML templates are first piped through Blade, so that variables may be used in MJML templates. This also lets emails be localized.

Handling of unsubscribe hasn't been done yet in the PHP, but the base email template has markup already setup for it.

Screenie:
image


locales/en.yaml
```diff
@@ -77,6 +77,17 @@ wiki:
...
+email:
+  GOTO_SITE: Go to website
+  SUBSCRIBED: You received this email because you are subscribed.
+  UNSUBSCRIBE: Unsubscribe
+  UNSUBSCRIBE_TEXT: "If you wish to unsubscribe, go here:"
+
+  social:
+    BLOG: Wikijump Blog
+    TWITTER: "@GetWikijump"
+    GITHUB: scpwiki/wikijump
```
emmiegit: Should this be a translation thing? It seems like you'd want this to be customized per-install, so I guess in `web/config` instead.

Like, looking at how these are used, it seems better to have this be the names of the platforms instead ("Twitter", etc).

web/app/Mail/MJMLMailable.php
```diff
@@ -0,0 +1,88 @@
...
+    protected string $mjml;
+
+    /**
+     * Sets the MJML template to use. This template has the highest priority.
+     * Templates will be rendered by Blade first, and then by MJML.
+     *
+     * @param string $mjml Path to the MJML template, in Blade template syntax..
+     * @param array $data Data to pass to the template.
+     */
+    public function mjml(string $mjml, array $data = []): self
```
emmiegit: `: self` works?

emmiegit (approved):

locales/en.yaml
Monkatraz: Yeah, I've been conflicted on how this should be handled... If this is configured (by not editing the templates, that is), then I guess I would do something like:
```php
$social = [];
$social[] = ['type' => 'web', 'name' => 'Wikijump Blog'];
$social[] = ['type' => 'twitter', 'name' => '@GetWikijump'];
// and so on
```
and put that in a `web/config` file, like you said. I wish I could type arrays...

web/app/Mail/MJMLMailable.php
Monkatraz: Apparently? I was trying to figure out how to type `$this`, but that can't be in that "returns" slot in the function signature, but `self` can go in there. Although, now that I think about it, that breaks polymorphism... ugh, I'll change that.

locales/en.yaml
emmiegit: You don't need to push individually like that:
```php
$social = [
    ['type' => 'web', 'name' => 'Wikijump Blog'],
    ['type' => 'twitter', 'name' => '@GetWikijump'],
    // ...
];
```

emmiegit: But yeah it's not a clear cut case, I agree

web/app/Mail/MJMLMailable.php
emmiegit: Honestly I think putting the class name there again is more readable anyways.

locales/en.yaml
Monkatraz: I only used the syntax because I was lazy when writing the reply lol.

What do you think of this in `web/config/wikijump.php`?
```php
/**
 * Social media links that appear in the footer of emails.
 *
 * Each array element is an array with the following keys:
 * - name: the name of the social media service (MJML syntax)
 * - url: the URL to link to
 * - text: the text to display
 * - src: icon image source
 *
 * None of the keys are required.
 */
'mail_social_links' => [
    [
        'name' => 'web',
        'url' => 'https://www.wikijump.org',
        'text' => 'Wikijump Blog',
    ],
    [
        'name' => 'twitter',
        'url' => 'https://twitter.com/getwikijump',
        'text' => 'Twitter',
    ],
    [
        'name' => 'github',
        'url' => 'https://github.com/scpwiki/wikijump',
        'text' => 'GitHub',
    ],
],
```

web/app/Mail/MJMLMailable.php
Monkatraz: Yeah, but that also breaks polymorphism. You have to use a doc comment with `@return $this` for it to work correctly. It feels kind of weird that `self` is valid but `this` isn't... PHP things I guess

locales/en.yaml
Monkatraz: Okay, I've changed it so that the social links are configurable via `web/config/wikijump.php`.

web/app/Mail/MJMLMailable.php
emmiegit: Lol alright makes sense

locales/en.yaml
emmiegit: Looks fine to me.

Monkatraz:
thanks @ammongit

PR #653, opened 2021-12-11
[WJ-931] [WJ-932] Generic MJML message template, email and password email messages

Monkatraz:
This PR adds quite a few things, but all of them are related to the new `MJMLMessage` class. This is an extension of the `MailMessage` class, which is used by Laravel to construct generic emails without needing an all new template. `MJMLMessage` does the same thing, except with our own customized template.

This was surprisingly complicated to do, mainly because I think Laravel's mailing code is weirdly hard-coded and complex.

Screenies:
image
image

### DIFF
--- a/locales/en.yaml
+++ b/locales/en.yaml
@@ -82,6 +82,8 @@ email:
   SUBSCRIBED: You received this email because you are subscribed.
   UNSUBSCRIBE: Unsubscribe
   UNSUBSCRIBE_TEXT: "If you wish to unsubscribe, go here:"
+  SUBCOPY: >
+    If you are having trouble clicking the ''{action}'' button, copy paste the URL below into your web browser:
 
 # Generic UI components
 components:
--- a/locales/en.yaml
+++ b/locales/en.yaml
@@ -85,6 +85,27 @@ email:
   SUBCOPY: >
     If you are having trouble clicking the ''{action}'' button, copy paste the URL below into your web browser:
 
+  verify_email:
+    SUBJECT: Verify Email Address
+    GREETING: Verify your email address
+    INTRO: Please verify your email address by clicking the button below.
+    ACTION: Verify Email Address
+    OUTRO: If you did not create an account, no further action is required.
+
+  reset_password:
+    SUBJECT: Password Reset Requested
+    GREETING: Resetting your password
+    INTRO: >
+      You are receiving this email because we received a password reset request for your account. Click the button below to reset your password.
+    ACTION: Reset Password
+    EXPIRES: >
+      This password reset link will expire in {count, number} {count, plural,
+        =1 {minute.}
+        other {minutes.}
+      }
+    OUTRO: If you did not request a password reset, no further action is required.
+
+
 # Generic UI components
 components:
   spinny:
--- a/web/app/Services/MJML/MJML.php
+++ b/web/app/Services/MJML/MJML.php
@@ -48,8 +48,9 @@ final class MJML
         // command has a 5 second timeout
         // if rendering takes longer than that, something is very wrong
         // average render time should be in milliseconds
-        $proc = Process::fromShellCommandline('mrml render', null, null, $mjml, 5);
-
+        $proc = new Process(['mrml', 'render']);
+        $proc->setInput($mjml);
+        $proc->setTimeout(5);
         $proc->run();
 
         if (!$proc->isSuccessful()) {
--- a/web/app/Services/MJML/MJML.php
+++ b/web/app/Services/MJML/MJML.php
@@ -22,14 +22,22 @@ final class MJML
      * @param string $template_path Blade template path.
      * @param array $data Data to be passed to the template.
      */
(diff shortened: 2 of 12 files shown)


emmiegit:
> mainly because I think Laravel's mailing code is weirdly hard-coded and complex.

What does Laravel expect emails to look like?

locales/en.yaml
```diff
@@ -82,6 +82,29 @@ email:
   SUBSCRIBED: You received this email because you are subscribed.
   UNSUBSCRIBE: Unsubscribe
   UNSUBSCRIBE_TEXT: "If you wish to unsubscribe, go here:"
+  SUBCOPY: >
```
emmiegit, line 85: "SUBCOPY"? Maybe "UNSUBSCRIBE_MANUAL" or something would be better?

locales/en.yaml
```diff
@@ -82,6 +82,29 @@ email:
...
+    If you are having trouble clicking the ''{action}'' button, copy paste the URL below into your web browser:
+
+  verify_email:
+    SUBJECT: Verify Email Address
+    GREETING: Verify your email address
+    INTRO: Please verify your email address by clicking the button below.
+    ACTION: Verify Email Address
+    OUTRO: If you did not create an account, no further action is required.
```
emmiegit, line 93: I like this format for emails :+1:

Monkatraz:
> 
> What does Laravel expect emails to look like?

The most annoying thing I was encountering was how I was supposed to be providing templates and views to certain classes. There are at least 3, maybe 4 ways of providing template identifiers (e.g. `emails.message.mjml`) or HTML (returning something that implements `Htmlable`). I'm really struggling to get specific here because it's just such a tangled web.

For example, to provide a "view" to use in a MailMessage, I think this would be the TypeScript type:
```ts
type TemplateIdentifier = string

type MessageView = string | ViewList | ViewTable

// no idea what `raw` is for, or what it does. it's not mentioned in documentation
type ViewList = [view: TemplateIdentifer | Htmlable, plain?: TemplateIdentifier | Htmlable, raw?: TemplateIdentifier | Htmlable]

interface ViewTable {
  "html"?: TemplateIdentifier | Htmlable,
  "text"?: TemplateIdentifier | Htmlable,
  "raw"?: TemplateIdentifier | Htmlable
}
```
I still have no idea if my understanding of this one specific thing is correct.

Monkatraz:
Actually, my type of that is definitely not the full picture. `MessageView` is actually the type to provided to the _mailer renderer_, but `MailMessage` has an alternative form of that type that I don't get at all. The type provided to `MailMessage` also differs from what you provide to `Mailable`, which is closer to what `Mailer` wants, but isn't exactly the same either.

web/app/Services/MJML/MJML.php
```diff
@@ -40,7 +48,7 @@ public static function render(string $template_path, array $data = []): HtmlStri
         // command has a 5 second timeout
         // if rendering takes longer than that, something is very wrong
         // average render time should be in milliseconds
-        $proc = Process::fromShellCommandline('mrml render', null, null, $raw_mjml, 5);
+        $proc = Process::fromShellCommandline('mrml render', null, null, $mjml, 5);
```
emmiegit: As a security note, I don't like shell invocation if direct invocation is possible. Looking at the [Symfony `Process` docs](https://symfony.com/doc/current/components/process.html), it looks like what we want is something like:
```php
$proc = new Process(['mrml', 'render']);
$proc->setInput($mjml);
$proc->setTimeout(5);
$proc->run();
```

emmiegit (approved):
:envelope:

emmiegit:
Hmm wow, not sure why the dynamic typing on that is so wild.

locales/en.yaml
Monkatraz: I actually didn't come up with that name. That's what Laravel calls it. I figured it was some standard jargon with emails, but it appears to be Laravel specific. I'll change it.

web/app/Services/MJML/MJML.php
Monkatraz: To be clear, the way this is being compiled is via `stdin` and not pasting anything into the command line. I agree not using the shell would be preferable, but how do you pass `stdin` to the `Process` instance?

locales/en.yaml
emmiegit: Oh huh, maybe it is standard, I've just never heard of it.

Looking quickly online, it just seems to be Laravel. Idk, your choice whether to change it, this is total bikeshedding lol.

web/app/Services/MJML/MJML.php
emmiegit: Whoops sorry, forgot to include that. See the edited code snippet above.

Monkatraz: Oh that's simple, okay, I'll do that.

Monkatraz:
thanks @ammongit

PR #655, opened 2021-12-13
[WJ-932] Account registration, email verification

Monkatraz:
This PR gets the pipeline from registering -> verifying working. This involved a lot of headache. In the process of adding that pipeline, I've also added an `Asset` class for handling clientside assets with less headache. This allowed for adding the `@inline` directive, which the new `email-verify-link` view uses so that it loads very quickly.


install/files/local/env
```diff
@@ -23,7 +23,7 @@ SESSION_LIFETIME=120
 
 MEMCACHED_HOST=cache
 
-MAIL_MAILER=smtp
+MAIL_MAILER=log
 MAIL_HOST=smtp.mailtrap.io
 MAIL_PORT=2525
 MAIL_USERNAME=null
```
emmiegit, line 29: Since `local` has stubbed emails, you could set the rest of these `MAIL_*` variables to be empty.

web/app/Common/Asset.php
```diff
@@ -0,0 +1,192 @@
...
+    public function __construct(string $name)
+    {
+        $this->name = $name;
+
+        // if the path doesn't start with a `/`, it's a Vite path
+        if (!Str::startsWith($name, '/')) {
+            $vite = App::make(Vite::class);
+
+            // check if path has any slashes in it
+            $is_shorthand = strpos($name, '/') === false;
```
emmiegit, line 45: > `int|false`

:frowning_face:

emmiegit (approved):

emmiegit:
Thanks for all your work on this!

install/files/local/env
Monkatraz: I think I'll just leave them alone for now, just so that the various .env files are only minimally modified. Makes it easier to diff them.

web/app/Common/Asset.php
Monkatraz: ?
`strpos` returns false if the substring can't be found. Otherwise, the `int` returned is the index. Unless this is a joke I'm not getting lmao

Monkatraz:
thanks @ammongit

web/app/Common/Asset.php
emmiegit: No just displeasure at weird PHP typing lol

PR #658, opened 2021-12-13
[WJ-931] Password recovery flow

Monkatraz:
This PR adds the flow for handling password recovery (forgot -> email -> reset -> login). This wasn't as bad as I thought it was going to be, honestly.

This set of authentication PRs reveals to me that handling all of these forms has turned a bit messy. I'm probably going to be looking for a Svelte form library that will clean this up a bit. Everything is contained in its own component at least, so it's not too bad.

### DIFF
--- a/web/app/Http/Controllers/AccountController.php
+++ b/web/app/Http/Controllers/AccountController.php
@@ -154,7 +154,7 @@ class AccountController extends Controller
      * Handles updating a password during password recovery.
      * Not part of the "proper" API.
      */
-    public function handlePasswordUpdate(Request $request): Response
+    public function handlePasswordRecoveryUpdate(Request $request): Response
     {
         $request->validate([
             'token' => 'required',
--- a/web/app/Http/Controllers/AccountController.php
+++ b/web/app/Http/Controllers/AccountController.php
@@ -159,8 +159,13 @@ class AccountController extends Controller
         $request->validate([
             'token' => 'required',
             'email' => 'required|email',
-            'password' => 'required|min:8|confirmed',
+            'password' => 'required|string',
         ]);
+
+        // do our more strict check on the password
+        if (!UserValidation::isValidPassword($request->input('password'))) {
+            return new Response('', 400);
+        }
 
         $credentials = $request->only(['email', 'password', 'token']);
 
--- a/web/app/Http/Controllers/AccountController.php
+++ b/web/app/Http/Controllers/AccountController.php
@@ -4,10 +4,13 @@ declare(strict_types=1);
 
 namespace Wikijump\Http\Controllers;
 
+use Illuminate\Auth\Events\PasswordReset;
 use Illuminate\Contracts\Auth\StatefulGuard;
 use Illuminate\Http\Request;
 use Illuminate\Http\Response;
 use Illuminate\Support\Facades\Hash;
+use Illuminate\Support\Facades\Password;
+use Illuminate\Support\Str;
 use Wikidot\Utils\WDStringUtils;
 use Wikijump\Models\User;
 use Wikijump\Services\UserValidation\UserValidation;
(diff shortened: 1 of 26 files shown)


web/app/Http/Controllers/AccountController.php
```diff
@@ -112,4 +115,76 @@ public function sendVerificationEmail(): Response
...
+                return new Response('', 403);
+
+            default:
+                return new Response('', 500);
+        }
+    }
+
+    /**
+     * Handles updating a password during password recovery.
+     * Not part of the "proper" API.
```
emmiegit, line 155: What do you mean by this?

web/app/Http/Controllers/AccountController.php
```diff
@@ -112,4 +115,76 @@ public function sendVerificationEmail(): Response
...
+                ->setRememberToken(Str::random(60));
+
+            $user->save();
+
+            event(new PasswordReset($user));
+        });
+
+        switch ($status) {
+            case Password::PASSWORD_RESET:
+                return new Response('', 200);
```
emmiegit: Nitpick: not 202?

web/modules/api/src/api.ts
```diff
@@ -104,27 +100,57 @@ class WikijumpAPIInstance extends Api<void> {
...
-   * Returns a base URL but for a different subdomain.
+   * Helper for sending a `GET` request via the API.
    *
-   * @param subdomain - The subdomain to use.
+   * @param to - The path to send the request to.
+   * @param query - The query parameters to send, if any.
    */
-  subdomainURL(subdomain: string) {
-    return `${window.location.protocol}//${subdomain}.${window.location.host}/${API_PATH}`
+  get<T = void>(to: string, query: Record<string, string> = {}) {
```
emmiegit, line 108: Since this is the return type, shouldn't it be:
```suggestion
  get<T = void>(to: string, query: Record<string, string> = {}): T {
```
?

web/routes/user-services.php
```diff
@@ -0,0 +1,68 @@
...
+});
+
+// Auth (unverified) routes
+Route::middleware('auth')->group(function () {
+    // Email verification routes
+
+    Route::get('/verify-email', function (User $user) {
+        // don't allow showing the verification notice if the user is already verified
+        if ($user->hasVerifiedEmail()) {
+            return redirect('/');
```
emmiegit, line 40: Nitpick: In the future we may want this to redirect to the user's profile/settings page?

web/routes/user-services.php
```diff
@@ -0,0 +1,68 @@
...
+        EmailVerificationRequest $request
+    ) {
+        $request->fulfill();
+        return response('', 200);
+    })
+        ->middleware('signed')
+        ->name('verification.verify');
+
+    // TODO: remove this temporary helper route
+    Route::get('/logout', [AuthController::class, 'logout'])->name('logout');
```
emmiegit, line 59: RIP `GET`.

Why is this temporarily needed?

emmiegit (approved):

web/app/Http/Controllers/AccountController.php
Monkatraz: It's not accessed via `/api--v0` or whatever. I suppose I could make it part of the actual API, but a token has to be provided and it's kind of unlike all the other API endpoints.

web/modules/api/src/api.ts
Monkatraz: TypeScript infers automatically.

web/routes/user-services.php
Monkatraz: Sure, when we have one lmao

web/routes/user-services.php
Monkatraz: It's hard to logout right now. You have to go to somewhere in the old frontend to find a button for it.

I'll be adding a more proper logout flow later when I add some account management UI.

Monkatraz:
thanks @ammongit

PR #659, opened 2021-12-14
Add Laravel Telescope

Monkatraz:
This PR adds Laravel Telescope, which is a _very_ useful monitoring tool for Laravel. The way it's been added is such that it only runs locally. This required some `nginx` changes, but now the `nginx` config supports including extra `.conf` files found in the `/etc/nginx/conf.d/wikijump` folder. This means that some of the `nginx` directives that were only useful for local are now limited to local.

Laravel Telescope does unfortunately add some vendored files to the `public/vendor` folder, but it's only a few files.

You can access Telescope via the `/telescope` route. This is _technically_ in conflict with page names, but since it's only local I'm not sure it really matters much.


danieltharp (approved):

danieltharp:
I could probably spend this year and next trying to do something as good as Telescope and not come close. Since you've got it under that `require-dev` directive, go for it.

emmiegit (approved):
:telescope:

Monkatraz:
thanks @ammongit, @danieltharp

PR #662, opened 2021-12-16
Instantiate DEEPWELL services

emmiegit:
This moves database logic out of the method implementations and into the appropriate services (currently just user) as needed. It instantiates the service pattern, with its initialization, error type, and other details.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #663, opened 2021-12-17
`wj-component-loader` element for async component loading

Monkatraz:
This PR adds a new HTML web-component that can be used to asynchronously load components. The purpose of this element is to allow for using clientside components (like Svelte) from Blade templates.

The tag, `wj-component-loader`, passes all attributes (except `load`) to the element. It also passes all slots, even to Svelte components. Doing the latter for Svelte is pretty hacky, but what's being done in the code to enable this was taken from `svelte-tag`, which has been around a while for doing this kind of thing.


web/resources/lib/component-loader/component-loader.ts
```diff
@@ -0,0 +1,263 @@
...
+      m: function mount(target: Node, anchor?: Node) {
+        insert(target, fragment.cloneNode(true), anchor)
+      },
+      d: function destroy(detaching: boolean) {
+        // @ts-ignore
+        if (detaching && fragment.innerHTML) {
+          detach(fragment)
+        }
+      },
+      l: noop
```
emmiegit, line 260: What are these single letters about? Svelte aliases?

emmiegit (approved):

web/resources/lib/component-loader/component-loader.ts
Monkatraz: Yeah, I think they're just single letter names to make the component smaller when generated. This is internal Svelte witchcraft so I dunno exactly what everything does, lmao

Monkatraz:
thanks @ammongit

PR #665, opened 2021-12-18
[WJ-906] Client status widget

Monkatraz:
This PR is kind of chunky. Ignore the huge line count, that's due to me updating the Mockoon JSON (I guess their format changed).

The main thing added is the `ClientStatus` component, which looks like this:
image
image

This required a lot of stuff be added, along with some reorganization. I still need to do a _bunch more stuff_, mainly:
1. Hook up [Ziggy](https://github.com/tighten/ziggy) so that we get our Laravel routes in JavaScript
2. Entirely reorganize the locale file, with a focus of on the meaning of a string rather than what it's for (e.g. it doesn't really matter where `auth.LOGIN` is, because all it needs to be is a translation string for the word `Login`)
3. Actually make some of the `user` API routes
4. A bunch of misc. cleanups, polish passes, etc. on this whole arrangement

I didn't do any of the above because this PR is already getting too big, lol


web/modules/api/src/api.ts
```diff
@@ -71,34 +71,60 @@ class WikijumpAPIInstance extends Api<void> {
...
+      this.updateIdentity()
       return res
     }
   }
 
+  /**
+   * Updates the current client `UserIdentity`. Usually called when
+   * authentication state changes.
+   */
+  private async updateIdentity() {
```
emmiegit, line 111: Not sure I fully understand what this does.

web/modules/components/src/Sprite.svelte
```diff
@@ -23,6 +29,13 @@
...
 </script>
 
-<svg class="wj-ui-sprite sprite-{i}" {viewBox} {...$$restProps} aria-hidden="true">
+<svg
+  class="wj-ui-sprite sprite-{i}"
+  width={size}
+  style="margin: {margin};"
+  {viewBox}
+  {...$$restProps}
+  aria-hidden="true"
```
emmiegit, line 38: Why is this after `restProps`?

web/resources/lib/components/UserInfo.svelte
```diff
@@ -0,0 +1,77 @@
+<script lang="ts">
+  import WikijumpAPI, { identity as currentIdentity } from "@wikijump/api"
+  import { Sprite } from "@wikijump/components"
+  import type { UserIdentity } from "@wikijump/api"
+
+  export let user: null | string | number = null
+
+  export let noavatar = false
+
```
emmiegit, line 9: Personally I prefer positive variables, like `hasAvatar`, which avoids double-negative constructions like `!noAvatar`.

emmiegit (approved):

emmiegit:
Very nice widget, thanks for your work on this.

web/modules/api/src/api.ts
Monkatraz: It's just getting the client's current `UserIdentity` object (the one in the API schema) and setting a store's value to the retrieved identity. It looks kind of complicated, but it's just to make sure the clientside API can't get itself stuck in a weird state.

web/modules/components/src/Sprite.svelte
Monkatraz: lol idk man this stuff happens

It doesn't actually matter, but at some point I'll probably go through a cleaning/documentation pass.

emmiegit:
The ftml failures are updated tests and new lints, you can ignore them for now.

web/resources/lib/components/UserInfo.svelte
Monkatraz: Fully agree, this is just a thing with Svelte. If this was instead `avatar` and it defaulted to `true`, you'd to do this:
```svelte
<UserInfo avatar={false} />
```
versus:
```svelte
<UserInfo noavatar />
```

When you chain multiple of these boolean attributes together it makes things more readable. I think there is a proposal for adding `!avatar` as a valid attribute, but I have no idea if that got anywhere.

Monkatraz:
thanks @ammongit

PR #666, opened 2021-12-18
[WJ-948] Add support for Fluent localization in frontend

Monkatraz:
This PR is part 1 of 2 for WJ-948. In this PR, the `@wikijump/fluent` module has been created. This is a low-level handler for Fluent resources. A couple of components have been converted to using this localization system in order to test and demonstrate it.

Despite the "low-level" description, the amount of code needed to get a working and robust system was not a lot. I am pretty damn happy with how it works, and I think in terms of file size it's going to be way smaller than our previous system.

In the next PR, I'll be converting the entire frontend to using Fluent.


emmiegit:
Question: When we remove the current localization system, will we rename `/locales/fluent` to just `/locales`?

Monkatraz:
> Question: When we remove the current localization system, will we rename `/locales/fluent` to just `/locales`?

Depends if we have anything other than localization files in `/locales`, I guess. If it turns out we don't, yeah we can move everything up a level.

locales/fluent/base/en.ftl
```diff
@@ -0,0 +1,15 @@
+### Base localization / generic strings
```
emmiegit, line 1: Do triple `#` mean anything special to fluent?

web/modules/fluent/src/default.ts
```diff
@@ -0,0 +1,73 @@
...
+
+// avoid our initial locale also being in our fallbacks
+if (fallbackLocales[0] === initialLocale) fallbackLocales.shift()
+
+// push fallback locale at the end if we don't have it anywhere
+if (initialLocale !== FALLBACK_LOCALE && !fallbackLocales.includes(FALLBACK_LOCALE)) {
+  fallbackLocales.push(FALLBACK_LOCALE)
+}
+
+// TODO: can this be made reactive like svelte-i18n?
```
emmiegit, line 23: For my sake, could you tell me a bit about what this means?

web/modules/fluent/src/locales.ts
```diff
@@ -0,0 +1,41 @@
+import { FluentComponent } from "./component"
+
+export const FALLBACK_LOCALE = "en"
+
+const filenameRegex = /[^\/]+\.ftl$/
+const componentRegex = /[^\/]+(?=\/[^\/]+\.ftl$)/
+const localeRegex = /.+(?=\.ftl$)/
+
+// only import folders one level deep
+// TODO: do we need nested directories?
```
emmiegit, line 10: Hopefully not. I think we should have this be one directory deep and make clear in the `/locales` README that you can't nest.

emmiegit:
Not this PR, but is it possible we can add a build step for verifying all the Fluent files are valid and don't have misspelled message keys etc.?

Monkatraz:
> Not this PR, but is it possible we can add a build step for verifying all the Fluent files are valid and don't have misspelled message keys etc.?

I don't see why not. If you want to do such a thing in Python (like we're already doing), there is an official Python package for Fluent that could be used.

emmiegit:
Not this PR, but is it possible we can add a build step for verifying all the Fluent files are valid and don't have misspelled message keys etc.?

> > Not this PR, but is it possible we can add a build step for verifying all the Fluent files are valid and don't have misspelled message keys etc.?
> 
> I don't see why not. If you want to do such a thing in Python (like we're already doing), there is an official Python package for Fluent that could be used.

Makes sense. I'll take a look at that when I can.

emmiegit (approved):

emmiegit:
Nice work on this. Glad we're getting closer to that full proper localization system lol

locales/fluent/base/en.ftl
Monkatraz: Yes, actually.

`###` - Comments that describe a file
`##` - Comments that break things into sections
`#` - Plain comments

I'm sure the idea of this is that it allows for translation UIs to break up these files automatically.

web/modules/fluent/src/default.ts
Monkatraz: `svelte-i18n` has this feature where if you change your language it'll update every string on the fly. That's because it's formatting functions are all observables.

Although, now that I think about it, since we have serverside views, we'd have to refresh the page anyways, so I don't think the reactivity is worth it.

web/modules/fluent/src/locales.ts
Monkatraz: Okay, agreed.

locales/fluent/base/en.ftl
emmiegit: Very interesting! Glad they have this convention.

We should make note of this in the `/locales` README I think.

web/modules/fluent/src/default.ts
emmiegit: Yeah reactivity is nice but it doesn't seem particularly worthwhile here.

locales/fluent/base/en.ftl
Monkatraz: Yeah. I haven't edited the README because I wanted the conventions to settle a bit. I'll edit the README in the next PR.

Monkatraz:
thanks @ammongit 

I'll remove the TODOs you talked about in the next PR.

I'm also going to see about making the Vite importer for the `ftl` files "compress" them by removing comments and whitespace. That way we don't need to care about lots of comments making it to JS.

PR #669, opened 2021-12-20
[WJ-948] Migrate (clientside) frontend to Project Fluent

Monkatraz:
This PR completes the migration of the frontend to Fluent. In the process, I've made sure to convert every localization string, not just the frontend-related ones.

Also, this PR has some random CSS in it because I realized the CSS for Sheaf's autocomplete tooltips wasn't working right, lol. Didn't want to bother with making a whole PR for it.

Finally, since this PR is pretty big, I again avoided making the README. I'll do that as its own PR.


locales/fluent/base/en.ftl
```diff
@@ -1,8 +1,44 @@
 ### Base localization / generic strings
+### Messages should only go into this file if they are widely used,
+### or are particularly important to localize.
 
-# Special message that shows whenever a message is still being loaded
+## Terms
+
+-service-name = Wikijump
```
emmiegit, line 7: What do leading dashes mean?

locales/fluent/base/en.ftl
```diff
@@ -12,4 +48,57 @@ logout = Logout
 register = Register
   .toast = You have been registered.
 
-create-account = Create an account
+specifier = Email or Username
+  .placeholder = Enter email or username...
+
+username = Username
+  .placeholder = Enter username...
+  .info = You will be able to change this later.
```
emmiegit, line 56: Unrelated to this PR, but I wonder if we should show what the user slug will be during registration. Especially since if the slug conflicts the user can't be created.

locales/fluent/cmftml/en.ftl
```diff
@@ -0,0 +1,112 @@
+### CodeMirror FTML Extension
+
+cmftml-undocumented-block = This block is valid, but hasn't yet been documented.
+
+## Linting
+
+cmftml-lint =
+  .warning-source = ftml({ $rule } = { $kind } at { $token }) [{ $from }, { $to }]
```
emmiegit, line 8: This is so comprehensive, thanks for doing all this.

web/modules/cm-lang-ftml/src/data/block.ts
```diff
@@ -0,0 +1,227 @@
...
+import { htmlArgumentCompletions } from "./html-attributes"
+import { type Argument, type BlockConfiguration, type ModuleConfiguration } from "./types"
+
+const BlockTipFactory = new EditorSvelteComponent(BlockTip)
+const ModuleTipFactory = new EditorSvelteComponent(ModuleTip)
+
+export class ArgumentData {
+  declare name: string
+
+  declare type: `${"string" | "int" | "float" | "bool"}${"[]" | ""}`
```
emmiegit, line 20: I'm amazed this works.

web/modules/cm-lang-ftml/src/data/data.ts
```diff
@@ -0,0 +1,67 @@
+import blocksTOML from "@root/ftml/conf/blocks.toml"
```
emmiegit, line 1: Interesting you can do this

emmiegit (approved):

emmiegit:
Lot of great work, thanks for doing all this

locales/fluent/base/en.ftl
Monkatraz: These are "terms", which I would think of as being like environment variables. They have a bunch of uses, especially because they support stuff like attributes and placeholders that can be given by other messages.

locales/fluent/base/en.ftl
Monkatraz: Good idea, actually. I'll do that.

web/modules/cm-lang-ftml/src/data/data.ts
Monkatraz: It's because I whipped up a Vite plugin a while back for it. It's very nice to be able to just import random assets and it just works lol

locales/fluent/cmftml/en.ftl
Monkatraz: lol no problem

I'm gonna be so happy when I can hook up Sheaf to the frontend proper. It's gonna be so god damn fancy.

locales/fluent/base/en.ftl
emmiegit: Makes sense!

web/modules/cm-lang-ftml/src/data/block.ts
Monkatraz: Template string types are one of those things in TypeScript that shouldn't work but does.

In the `defaults.ts` file in the Fluent module, there is a type for Unit strings (like "millimeters" for "20mm") which is like 1800+ individual strings if you were to type it out as a union. That's because it supports doing `unit-per-unit`. There are some hilarious types like "feet-per-kelvin" because of that.

Monkatraz:
thanks @ammongit 

_now it is your turn......._

actually should be reasonably easy to add to Deepwell, since you don't need to bother with asynchronous loading or whatever.

PR #671, opened 2021-12-21
Cleanup TypeScript imports

Monkatraz:
This PR cleans up TypeScript imports through the new `type` import modifier. It also enables a more strict check on imports in `tsconfig.json`.

Note: If you see `import type { foo } from "foo"`, that's actually _different_ than `import { type foo } from "foo"`. Why? Because the latter compiles to `import {} from "foo"`, which _may_ have side effects. That's why type-only imports still use `import type`.

I also fixed some ftml tests that I screwed up lol


Monkatraz:
Not sure why ftml lint is failing, did Clippy get an update?

emmiegit:
Seeing that explanation in the PR description, good to know, but man side effects on import were a mistake.

emmiegit (approved):

emmiegit:
> Not sure why ftml lint is failing, did Clippy get an update?

Yeah I mentioned this in chat, there are new lints. Ignore it, I'll fix it in a PR.

Monkatraz:
> Seeing that explanation in the PR description, good to know, but man side effects on import were a mistake.

Eh, I'll defend it a _little bit_. The browser environment is _pretty frickin' weird_ compared to like, a backend server. It's very monolithic and honestly the closest thing I can compare it to is a video game. Plus with how JS works I don't think avoiding side effects on import was even possible.

You can also tell TypeScript to remove imports that are unused, which makes the `import "foo"` syntax how you get a side-effects import. However, I do prefer the explicit behavior and error that TS gives you with the current setup.

Monkatraz:
thanks @ammongit

PR #672, opened 2021-12-21
[ftml] Fix new clippy lints

emmiegit:
New lints have been added, which is causing the CI step to fail.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #673, opened 2021-12-22
Use Ziggy to support using Laravel route names on the frontend

Monkatraz:
This PR adds Ziggy, which allows us to use Laravel's named routes on the frontend. Ziggy does this by sticking a global `Ziggy` object with named route information in it in a `<script>` tag.

Here is an example of the kind of thing it generates:
```js
const Ziggy = {"url":"http:\/\/www.wikijump.localhost","port":null,"defaults":{},"routes":{"ignition.healthCheck":{"uri":"_ignition\/health-check","methods":["GET","HEAD"]},"ignition.executeSolution":{"uri":"_ignition\/execute-solution","methods":["POST"]},"ignition.shareReport":{"uri":"_ignition\/share-report","methods":["POST"]},"ignition.scripts":{"uri":"_ignition\/scripts\/{script}","methods":["GET","HEAD"]},"ignition.styles":{"uri":"_ignition\/styles\/{style}","methods":["GET","HEAD"]},"telescope":{"uri":"telescope\/{view?}","methods":["GET","HEAD"]},"login":{"uri":"user--services\/login","methods":["GET","HEAD"]},"register":{"uri":"user--services\/register","methods":["GET","HEAD"]},"password.request":{"uri":"user--services\/forgot-password","methods":["GET","HEAD"]},"password.reset":{"uri":"user--services\/reset-password\/{token}","methods":["GET","HEAD"]},"password.update":{"uri":"user--services\/reset-password\/{token}","methods":["POST"]},"verification.notice":{"uri":"user--services\/verify-email","methods":["GET","HEAD"],"bindings":{"user":"id"}},"verification.verify":{"uri":"user--services\/verify-email\/{id}\/{hash}","methods":["POST"]},"logout":{"uri":"user--services\/logout","methods":["GET","HEAD"]},"password.confirm":{"uri":"user--services\/confirm-password","methods":["GET","HEAD"]},"socialite-callback":{"uri":"social--providers\/callback","methods":["GET","HEAD"]},"account":{"uri":"--\/account","methods":["GET","HEAD"]},"account.profile":{"uri":"--\/profile","methods":["GET","HEAD"]},"account.settings":{"uri":"--\/settings","methods":["GET","HEAD"]},"account.messages":{"uri":"--\/messages","methods":["GET","HEAD"]},"account.notifications":{"uri":"--\/notifications","methods":["GET","HEAD"]},"user.profile":{"uri":"--\/u\/{user}","methods":["GET","HEAD"]},"docs":{"uri":"--\/docs","methods":["GET","HEAD"]},"terms":{"uri":"--\/terms","methods":["GET","HEAD"]},"security":{"uri":"--\/security","methods":["GET","HEAD"]},"privacy":{"uri":"--\/privacy","methods":["GET","HEAD"]},"report-flag":{"uri":"--\/report-flag","methods":["GET","HEAD"]}}};
```

Then, using the new `route` function from `@wikijump/api`, you can do stuff like:
```ts
const url = route("user.profile", { user: username })
```

Also. I added a bunch of dummy routes for stuff that doesn't exist yet but that I'm still linking to.

### DIFF
--- a/web/routes/web.php
+++ b/web/routes/web.php
@@ -123,6 +123,24 @@ Route::get('/user--avatar/{user}', function (User $user) {
 
 // -- WIKI
 
+// (names, not route paths)
+// TODO: account
+// TODO: account.profile
+// TODO: account.settings
+// TODO: account.messages
+// TODO: account.notifications
+// TODO: user.profile
+// TODO: docs
+
+// dummy routes so that these resolve on the frontend
+Route::view('--/account', 'next.test.page-test')->name('account');
+Route::view('--/profile', 'next.test.page-test')->name('account.profile');
+Route::view('--/settings', 'next.test.page-test')->name('account.settings');
+Route::view('--/messages', 'next.test.page-test')->name('account.messages');
+Route::view('--/notifications', 'next.test.page-test')->name('account.notifications');
+Route::view('--/u/{user}', 'next.test.page-test')->name('user.profile');
+Route::view('--/docs', 'next.test.page-test')->name('docs');
+
 if (GlobalProperties::$FEATURE_FRONTEND === 'next') {
     // Legacy special routes
     Route::any('/{special}:{path}', [OzoneController::class, 'handle'])
--- a/web/routes/web.php
+++ b/web/routes/web.php
@@ -131,6 +131,10 @@ Route::get('/user--avatar/{user}', function (User $user) {
 // TODO: account.notifications
 // TODO: user.profile
 // TODO: docs
+// TODO: terms
+// TODO: security
+// TODO: privacy
+// TODO: report-flag
 
 // dummy routes so that these resolve on the frontend
 Route::view('--/account', 'next.test.page-test')->name('account');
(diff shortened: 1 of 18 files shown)


web/routes/web.php
```diff
@@ -123,6 +123,32 @@
...
+// TODO: account.notifications
+// TODO: user.profile
+// TODO: docs
+// TODO: terms
+// TODO: security
+// TODO: privacy
+// TODO: report-flag
+
+// dummy routes so that these resolve on the frontend
+Route::view('--/account', 'next.test.page-test')->name('account');
```
emmiegit, line 140: What does `--` mean?

emmiegit (approved):
I had no idea this even existed lol. Good find

web/routes/web.php
Monkatraz: `--` is an invalid Wikidot route, right? I just picked something that wouldn't conflict.

I actually kind of like it because then I can put anything after it, like `login` or `register`. `user--services` is unwieldy, and we're gonna have to come up with equivalents for several categories (like accounts).

To be clear these are temporary URLs.

emmiegit: Makes sense. I was curious if it was a Laravel/Ziggy thing.

Monkatraz:
thanks @ammongit

PR #674, opened 2021-12-23
Development tooling madness

Monkatraz:
Another generalish PR:
- Made sure to copy over `.env` file when running Vite builds, as Vite can use environment variables
- Added a nicer NGINX `log_format` for when you're running it locally
- Added a `NGINX` service to PHP which has a `fetch()` method. I isolated this out because we haven't sorted out inter-container communication and I wanted a single, isolated place which we could change later.
- Made sure to import Svelte components asynchronously in a few places. Basically, I'm preventing the client from needing to download any Svelte components synchronously when they first visit the site. This decreases bundle size quite a bit, and improves page load performance.
- Added some `manualChunks` configuration for Rollup, in order to improve how the final asset bundles are built. An example of this is that CodeMirror gets imported in a single, big chunk, rather than in a million pieces.
- Made docker logs very pretty if you're running `pnpm dev`. This is accomplished by an absurd log parsing abomination that I made in `web/scripts/pretty-docker-logs.js`. If this ever breaks something, just get rid of it. It does make the logs much nicer to read though. If you need to view the raw logs, you can run `pnpm serve`.

image
image


Monkatraz:
I have no idea why the AWS builds are failing now. I don't think this PR did it?

web/app/Providers/AppServiceProvider.php
```diff
@@ -3,10 +3,10 @@
 
 namespace Wikijump\Providers;
 
-use Exception;
 use Illuminate\Support\Facades\Log;
 use Illuminate\Support\ServiceProvider;
 use Laravel\Telescope\TelescopeServiceProvider;
+use Wikijump\Services\NGINX\NGINX;
```
emmiegit: This is very pedantic, but I thought it made sense with `MJML` since it's an acronym, but `nginx` is one weird and PHP does a lot of `AdaCase` I feel like this should be `Wikijump\Services\Nginx\Nginx`. I've been feeling weird about `DEEPWELL` even though it's similar to `FTML` (all-caps or all-lower), that it too should be `Services\Deepwell`.

web/app/Services/NGINX/NGINX.php
```diff
@@ -0,0 +1,55 @@
...
+
+use Exception;
+use Illuminate\Support\Facades\Log;
+
+/**
+ * Static class that holds methods for interacting with the NGINX container.
+ */
+final class NGINX
+{
+    // TODO: Docker paths aren't reliable (or working at all) on AWS
```
emmiegit: Yeah this will need to be `.env` or something to work.

web/app/Services/NGINX/NGINX.php
```diff
@@ -0,0 +1,55 @@
...
+     * @param string $accept Accept header to send.
+     */
+    public static function fetch(string $path, string $accept = 'text/plain'): ?string
+    {
+        $url = self::NGINX_URL . $path;
+
+        $opts = [
+            'http' => [
+                'method' => 'GET',
+                'header' => "Accept: {$accept}",
```
emmiegit: I honestly don't understand PHP interpolation rules. I know `"Accept: $accept"` will do what you expect, is it smart enough to also see the braces here?

web/modules/components/src/lib/modal.ts
```diff
@@ -44,12 +50,24 @@ export class Modal<T extends typeof SvelteComponent> {
...
 
     const modals = document.getElementById("modals")
     if (!modals) throw new Error("Modals container not found")
 
+    this.loading = this.loadDialog(component, open, lazy)
+  }
+
+  /** Imports and constructs the dialog. */
+  private async loadDialog(component: T, open: boolean, lazy: boolean) {
+    component = component instanceof Promise ? (await component).default : component
```
emmiegit, line 63: Would this be worth moving into a helper?

web/scripts/pretty-docker-logs.js
```diff
@@ -0,0 +1,134 @@
+const chk = require("chalk")
```
emmiegit, line 1: > Made docker logs very pretty if you're running pnpm dev. This is accomplished by an absurd log parsing abomination that I made in web/scripts/pretty-docker-logs.js. If this ever breaks something, just get rid of it. It does make the logs much nicer to read though. If you need to view the raw logs, you can run pnpm serve.

You should add some variant of this disclaimer to the file.

emmiegit (approved):

web/app/Providers/AppServiceProvider.php
Monkatraz: Yeah, hm. I'll rename it.

web/app/Services/NGINX/NGINX.php
Monkatraz: Including the braces is a habit but also doesn't hurt anything. You use braces to interpolate expressions rather than just a variable.

emmiegit: Weird, because you can interpolate `$thing->method()` or `$thing->field`.

web/modules/components/src/lib/modal.ts
Monkatraz: Dealing with `import()` is always a mess, mostly due to bundlers. I'd rather just handle situations like this manually.

web/scripts/pretty-docker-logs.js
Monkatraz: Oh yeah, will do.

web/app/Services/NGINX/NGINX.php
Monkatraz: You can? God PHP is weird.

Monkatraz:
thanks @ammongit

PR #677, opened 2021-12-23
[WJ-954] Move frontend testing to Vitest

Monkatraz:
This PR switches us over to [Vitest](https://github.com/vitest-dev/vitest). Vitest is a testing framework _designed_ for Vite. Here are some things to note:
- Tests are ran in Node, not in browser. This used to be a massive issue but Vitest handles a lot of the hassle because it's _designed_ for this. Playwright is no longer needed, which means no big Chrome binary has to be installed.
- Code coverage is _way more accurate_. Most files are at 0% now. This will absolutely destroy our code coverage rating, but that's accurate as the frontend has very few tests.
- No more hacks are needed for testing.
- Less dependencies.
- Vitest has a watch mode using Vite's module graph and incremental compiling features. That makes actually running tests when developing very convenient.
- Vitest runs tests in parallel to an almost ludicrous degree for JavaScript. I mean like, with workers. It's multi-threaded.

I also fixed the `App/Services/Nginx` folder name because I just missed that somehow.


web/modules/ftml-wasm/tests/ftml.ts
```diff
@@ -86,8 +84,12 @@ describe("ftml-wasm", () => {
...
   })
 
-  it("renderHTML", async () => {
-    await lib.loading
+  // following two tests are failing, reason appears to be
+  // due to getrandom failing. I think this is an issue
+  // with how getrandom calls up to the browser for randomness.
+  // other tests don't fail, but these do, for some reason.
+
+  it.skip("renderHTML", () => {
```
emmiegit, line 92: Test mode doesn't compile ftml in test mode right? If it did this issue wouldn't be present, but if you use release or debug you'll have this issue. I should add an option for parse settings to make random ID generation deterministic for testing.

web/scripts/test-setup.js
```diff
@@ -0,0 +1,7 @@
+// neither JSDOM or happy-dom set a default language
+navigator.languages = ["en"]
+navigator.langauge = "en"
+
+// happy-dom doesn't seem to have matchMedia
+// the HTMLElement interface satisfies enough of it to make it work
+window.matchMedia = () => new HTMLDivElement()
```
emmiegit, line 7: lmao

emmiegit (approved):

web/modules/ftml-wasm/tests/ftml.ts
Monkatraz: That would work, or calling `getrandom()` yourself to provide the seed, as that returns a `Result`, unlike the `from_entropy()` method which throws. I'm not sure what you should use as your seed if `getrandom()` fails, but considering our use case it really wouldn't matter that much what you use as the fallback.

web/scripts/test-setup.js
Monkatraz: I just needed `addEventListener` okay

Monkatraz:
thanks @ammongit

PR #679, opened 2021-12-24
[WJ-949] Add Fluent support to DEEPWELL

emmiegit:
As part of epic [WJ-947](https://scuttle.atlassian.net/browse/WJ-947), we have been switching from ICU to Fluent as our localization system. This replaces the existing ICU / gettext-based method, which had some flaws, especially in Rust, where we could not implement message formatting.

This implementation is simple, because it doesn't need to see "components" like the frontend does, it simply copies all messages in `/locales/fluent` into each locale's respective `FluentBundle`.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #681, opened 2021-12-25
[WJ-949] Add attributes to DEEPWELL's Fluent implementation

emmiegit:
Follow-up to #679 

This adds supports for attributes to DEEPWELL's `locales` module. This includes properly parsing them out and returning them if present, and a new error case for missing attributes (as opposed to missing values). (Though, if this error isn't useful being split from the value one, I could merge them).

The change was made for both `HEAD /message/:locale/:key` and `POST /message/:locale/:key`, checking the existence of a template and actually formatting. (The latter is `POST` instead of `GET` because it requires passing in a JSON body to specify variables).

### DIFF
--- a/deepwell/src/locales/error.rs
+++ b/deepwell/src/locales/error.rs
@@ -59,4 +59,7 @@ pub enum LocalizationTranslateError {
 
     #[error("Message key was found, but has no value")]
     NoMessageValue,
+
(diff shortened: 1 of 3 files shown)


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #682, opened 2021-12-25
Add message about asset bundles after startup

emmiegit:
As discussed in chat, since this is not an error if it's occurring during startup, then it's probably not an issue because nginx is still initializing.

I explored using health checks to avoid the issue in container startup, but because nginx depends on php-fpm, we can't have php-fpm also depend on nginx.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #683, opened 2021-12-25
[WJ-950] Change PHP to use Fluent localizations

emmiegit:
This sets up PHP to call DEEPWELL for its translations instead of using the old ICU setup. The PHP Fluent library is too immature to use directly, and DEEPWELL works fine here. I also had to set up the ratelimit bypass secret in PHP since otherwise it would hit HTTP 429s and rendering any page would fail.

### DIFF
--- a/deepwell/src/api/v0.rs
+++ b/deepwell/src/api/v0.rs
@@ -34,6 +34,7 @@ pub fn build(mut app: ApiServer) -> ApiServer {
     app.at("/ping").all(ping);
     app.at("/version").get(version);
     app.at("/version/full").get(full_version);
+    app.at("/ratelimit-exempt").all(ratelimit_exempt);
     app.at("/teapot")
         .get(|_| async { error_response(StatusCode::ImATeapot, "🫖") });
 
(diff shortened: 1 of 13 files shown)


Monkatraz (approved):

emmiegit:
> #### warning GitGuardian has uncovered 2 secrets following the scan of your pull request.
> 
> Please consider investigating the findings and remediating the incidents. Failure to do so may lead to compromising the associated services or software components.
> mag_right Details of the secret
> hammer_and_wrench Guidelines to remediate hardcoded secrets
> ###### [GitGuardian](https://dashboard.gitguardian.com/auth/login/?utm_medium=checkruns&utm_source=github&utm_campaign=cr1) is an automated secrets detection service.We help developers and security teams secure the modern software development process.
> 
>  
> ###### Our GitHub checks need improvements? [Share your feedbacks](https://qxhaizuqzpk.typeform.com/to/KmeAPTMk)

Yeah...

Unfortunately our current secrets infrastructure doesn't really exist, so the local secret is just here. Neither dev nor prod have values. We will need a better way to add secrets to the codebase so this isn't an issue for us and we don't have anything in plaintext. Until then I'm ensuring anything published is local or doesn't matter.

emmiegit:
thanks @Monkatraz

PR #686, opened 2021-12-25
Use `picocolors` instead of `chalk` for coloring

Monkatraz:
This is a tiny PR to use the `picocolors` package instead of `chalk` for terminal coloring. Picocolors is a tiny dependency compared to Chalk. This doesn't actually remove Chalk, because we have other dependencies which use it, so...

Really, I just wanted to remove Chalk because it had a fixed dependency version, due to it being strictly ESM only past V5.


emmiegit (approved):
:paintbrush:

emmiegit:
I gotta say, I really like "chalk" as a library name though

Monkatraz:
thanks @ammongit

PR #689, opened 2021-12-25
[WJ-951] Add CI checks for new Fluent localization

emmiegit:
As the final part of our Fluent localization implementation, this PR adds a new GitHub workflow for testing our localization setup. This runs a new Rust utility which reads in all Fluent files, builds a catalog of their internals, and then validates various aspects of the different locales against the primary locales to catch typos, issues, and errors.

This also deletes all the old gettext / YAML localization files, as they are now unused.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #691, opened 2021-12-25
Add read-only database user

emmiegit:
This adds the database user `wikijump_ro`, which is like `wikijump` except it is read-only, and can make no changes to the database.

However, due to some technical nuisances with how permissions are inherited, despite a few hours of trying, I have not been able to revoke `CREATE` from the user (i.e., the ability to create and drop its own tables, independent of those belonging to `wikijump`).


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #694, opened 2021-12-26
Some tests for `wikijump/dom`

Monkatraz:
This PR adds some tests for `@wikijump/dom`. It doesn't test the entire module, just some parts of it.

I've also fixed some coverage issues.

### DIFF
--- a/web/scripts/vite-utils.js
+++ b/web/scripts/vite-utils.js
@@ -116,7 +116,8 @@ const BaseConfig = () => ({
 
   test: {
     root: ROOT,
-    environment: "happy-dom",
+    isolate: false,
+    environment: "jsdom",
     include: ["./modules/**/tests/**/*.{js,ts}"],
     setupFiles: ["./scripts/test-setup.js"],
     deps: {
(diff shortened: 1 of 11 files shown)


web/scripts/vite-utils.js
```diff
@@ -116,7 +116,8 @@ const BaseConfig = () => ({
 
   test: {
     root: ROOT,
-    environment: "happy-dom",
+    isolate: false,
```
emmiegit, line 119: Why is `isolate: false` here?

emmiegit (approved):

web/scripts/vite-utils.js
Monkatraz: It's a new option. If `isolate` is true, every test will be ran in an isolated environment. This adds some startup delay, among other things. I don't think we need this feature, as I'm not leaving garbage everywhere when I'm writing tests. At least for now, I mean, it could be needed later, but for right now keeping it off gives us a performance boost.

This will be in the Vitest docs eventually, things are a bit WIP over there lol

Monkatraz:
thanks @ammongit

PR #696, opened 2021-12-26
Install pgcrypto extension

emmiegit:
[`pgcrypto`](https://www.postgresql.org/docs/current/pgcrypto.html) is a standard PostgreSQL extension which permits use of cryptographic functions. Specifically here, I am interested in the SHA-512 function as part of content-addressable long string storage.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #703, opened 2021-12-30
Tooling madness

Monkatraz:
YET ANOTHER INCOHERENT PR

Changes:
* Added `editorconfig` stuff for HTML, Svelte, and CSS stuff
* Switched to a waaaay better Svelte ESLint parser/plugin
* Find out that `svelte-preprocess` has had `<style global>` this entire gosh darn time and I didn't know
* Use `<style global>` everywhere
* Add `.dic` and `.aff` files to Vite's static file handling list
* Use inline type imports in Svelte files (blocker was old ESLint plugin)
* Prepend SCSS abstracts import for Svelte files
* Run `pnpm serve` through the `dev.js` script, which basically prevents Vite from running


emmiegit (approved):

Monkatraz:
thanks @ammongit

PR #705, opened 2022-01-01
[WJ-871] User Dashboard Part Uno

Monkatraz:
I'm splitting this PR up because holy crap it got chunky. The vast majority of the changes come from the OpenAPI spec, which I've fully "schemaized", by which I mean I've typed out any object schemas that were previously defined inline.

I'll go into more detail about how the user dashboard works in the second PR. But, in short, it's an SPA. I'll need to be implementing a bunch of the `/user` API endpoints to get the dashboard to actually work, so that's why there are so many API related changes.


deepwell/src/methods/user.rs
```diff
@@ -36,7 +39,8 @@ pub async fn user_get(req: ApiRequest) -> ApiResponse {
     let reference = ItemReference::try_from(&req)?;
     let user = req.user(&txn).get(reference).await.to_api()?;
     txn.commit().await?;
-    build_user_response(&user, StatusCode::Ok)
+    let query: UserDetailsQuery = req.query().unwrap_or_default();
```
emmiegit: Since `req.query()` returns `Result<_>`, wouldn't this mean that a malformed details param yields implicit default instead of an error?

Annotate `detail` in `UserDetailsQuery` with `#[serde(default)]` and then you can just do `req.query()?` and handle errors but get defaults as expected.

This is very minor, but I'm not super big on `Default` being implemented for `UserDetailsQuery`.

emmiegit: Optional suggestion:
```suggestion
    let UserDetailsQuery { details } = req.query().unwrap_or_default();
```

deepwell/src/methods/user.rs
```diff
@@ -45,24 +49,28 @@ pub async fn user_put(mut req: ApiRequest) -> ApiResponse {
...
+    build_user_response(&user, UserDetails::default(), StatusCode::Created)
 }
 
 pub async fn user_delete(req: ApiRequest) -> ApiResponse {
     let txn = req.database().begin().await?;
     let reference = ItemReference::try_from(&req)?;
     let user = req.user(&txn).delete(reference).await.to_api()?;
     txn.commit().await?;
-    build_user_response(&user, StatusCode::Ok)
+    build_user_response(&user, UserDetails::default(), StatusCode::Ok)
```
emmiegit: The nice thing about having query params is we can add them to both of these routes. I think we should do the same `let query: UserDetailsQuery = req.query()?` for both of these and permit the caller to specify what shape they want the output in, since our code is generic over it anyways.

deepwell/src/methods/user.rs
```diff
@@ -45,24 +49,28 @@ pub async fn user_put(mut req: ApiRequest) -> ApiResponse {
...
+fn build_user_response(
+    user: &UserModel,
+    user_detail: UserDetails,
+    status: StatusCode,
+) -> ApiResponse {
+    // TODO: allow dumping the entire user model (internal API only)
+    let body = match user_detail {
+        UserDetails::Identity => Body::from_json(&UserIdentityOutput::from(user))?,
+        UserDetails::Info => Body::from_json(&UserInfoOutput::from(user))?,
+        UserDetails::Profile => Body::from_json(&UserProfileOutput::from(user))?,
```
emmiegit, line 74: Good use of `From`.

deepwell/src/services/user.rs
```diff
@@ -61,6 +62,78 @@ pub struct UpdateUser {
...
+    tinyavatar: Option<String>, // TODO
+    karma: u8,
+    role: String,
+}
+
+impl UserIdentityOutput {
+    pub fn from(user: &UserModel) -> Self {
+        Self {
+            id: user.id,
+            username: user.username.clone(),
```
emmiegit, line 78: We can avoid this clone.

In each method, we get a `UserModel` and are effectively passing that as a return value, we don't use it afterwards. Instead of borrowing it, if you passed the owned version of it (`user: UserModel`) then you can just move the `username` and any other allocated fields straight out without needing to reallocate something which is about to be destroyed anyways.

Hmm though looking at it, it does complicate the implementations which build off of the simpler versions, since they cannot also pass in the owned version, and Rust doesn't have a way of knowing that you aren't moving the same field twice anywhere.

deepwell/src/web/user_details.rs
```diff
@@ -0,0 +1,52 @@
...
+impl Default for UserDetails {
+    #[inline]
+    fn default() -> Self {
+        UserDetails::Identity
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Default, Clone)]
+pub struct UserDetailsQuery {
+    pub detail: UserDetails,
```
emmiegit, line 52: As noted above:
```suggestion
    #[serde(default)]
    pub detail: UserDetails,
```

locales/fluent/base/en.ftl
```diff
@@ -41,6 +49,8 @@ publish = Publish
 reveal-sidebar = Reveal Sidebar
 save = Save
 security = Security
+send = Send
+sent = Sent
```
emmiegit, line 53: Does Fluent have any guidelines/help on localizing tenses?

web/app/Http/Controllers/UserController.php
```diff
@@ -1,86 +1,64 @@
...
-     * @return Response
-     */
-    public function edit(User $user): Response
-    {
-        //
-    }
+        if ($path_type === 'slug') {
+            $user = User::where('slug', $path)->first();
+        } elseif ($path_type === 'id') {
+            $user = User::find($path);
```
emmiegit: Will this work if `$path` is a stringified int?

web/app/Models/User.php
```diff
@@ -202,8 +202,7 @@ protected function defaultAvatar(): string
             '&color=' .
             $colors[$scheme]['t'] .
             '&background=' .
-            $colors[$scheme]['b'] .
-            '&rounded=true';
+            $colors[$scheme]['b'];
```
emmiegit, line 205: rounderhouse is crying

web/app/Services/Deepwell/DeepwellService.php
```diff
@@ -67,24 +67,28 @@ public function translate(string $locale, string $key, array $values = []): ?str
     }
 
     // User
-    public function getUserById(int $id): ?object
+    public function getUserById(int $id, ?string $detail = null): ?object
     {
-        $resp = $this->client->get("user/id/$id");
+        $query = $detail ? "?detail=$detail" : '';
+        $resp = $this->client->get("user/id/$id$query");
```
emmiegit: Guzzle has a way of [passing in query parameters without manually building them](https://docs.guzzlephp.org/en/stable/quickstart.html#query-string-parameters). Also I don't see anything wrong with having `string $detail = 'identity'` and just passing the query parameter each time.

web/app/Services/Deepwell/DeepwellService.php
```diff
@@ -127,6 +138,20 @@ private static function readJson($resp): object
...
+     * @param array $properties The properties to convert.
+     */
+    private static function convertDateProperties(object $obj, array $properties): void
+    {
+        foreach ($properties as $property) {
+            if (property_exists($obj, $property)) {
+                $obj->$property = self::nullableDate($obj->$property);
+            }
+        }
+    }
```
emmiegit, line 155: Yeah I like this helper better than what I was doing lol

web/modules/api/src/api.ts
```diff
@@ -152,6 +152,25 @@ class WikijumpAPIInstance extends Api<void> {
...
+   * @param method - The API method to call.
+   * @param args - The arguments to pass to the method.
+   */
+  async try<
+    M extends keyof FilterFor<Api<void>, AnyFunction>,
+    F extends AnyFunction = FilterFor<Api<void>, AnyFunction>[M]
+  >(method: M, ...args: Parameters<F>): Promise<null | Awaited<ReturnType<F>>> {
+    try {
+      // @ts-ignore - TS can't quite figure this one out
+      return await this[method](...args)
```
emmiegit, line 168: I keep hearing that "return await" is bad but I still don't understand why.

web/modules/api/src/index.ts
```diff
@@ -13,77 +13,113 @@ export type {
...
   Username,
+  UsernameObject,
   UserProfile,
   UserProfilePatch,
-  UserRole,
   Vote,
   VoterList,
   Wikitext,
-  WikitextObj
+  WikitextObject
```
emmiegit, line 122: :+1:

web/resources/lib/components/dashboard/Dashboard.svelte
```diff
@@ -0,0 +1,115 @@
...
+  import { Route } from "tinro"
+  import RouteAnnouncer from "../RouteAnnouncer.svelte"
+  import DashboardLink from "./DashboardLink.svelte"
+  import DashboardPanel from "./DashboardPanel.svelte"
+  import ProfilePanel from "./panels/ProfilePanel.svelte"
+  import SettingsPanel from "./panels/SettingsPanel.svelte"
+
+  const t = Locale.makeComponentFormatter("dashboard")
+
+  // bit of a hack, unfortunately, but this is to prevent Tinro
```
emmiegit, line 26: I'm assuming here Tinro is the thing doing routing for the settings SPA?

web/routes/web.php
```diff
@@ -149,6 +149,12 @@
 Route::view('--/privacy', 'next.test.page-test')->name('privacy');
 Route::view('--/report-flag', 'next.test.page-test')->name('report-flag');
 
+// Dashboard
+Route::view('/--/dashboard/{path?}', 'next.wiki.dashboard')
```
emmiegit: This starts with `/--` instead of `--` (compared to the others), is this intentional?

emmiegit (approved):

deepwell/src/methods/user.rs
Monkatraz: Despite this being internal, I am following the external OpenAPI spec. Kind of. I mean, should a malformed query always return an error? I guess in an API context it should, but usually with query parameters you just ignore them if they don't make any sense.

deepwell/src/methods/user.rs
Monkatraz: Are these methods intended to always be internal or will we eventually make these external and match the spec? I didn't bother adding query parameters because in the spec these methods wouldn't be returning anything other than a response code.

deepwell/src/services/user.rs
Monkatraz: Yeah, I cloned it because hypothetically I don't know what you're going to do with these objects, nor what you're going to do with the `UserModel` afterwards.

web/modules/api/src/api.ts
Monkatraz: It's kind of bad. It also kind of doesn't matter, usually.

Promises can be nested, e.g. `Promise<Promise<Promise<string>>>`. Asynchronous functions implicitly wrap their returned value in a `Promise`. If I didn't do `await`, I would be returning `Promise<Promise<foo>>` instead of a `Promise<foo>`. I'd rather the return type be the same as the function you are "trying", so I `await` the function.

Arguably, the consumer should be the one bothering to `await` any promises, regardless of nesting. However, I don't know when this would actually matter, because a promise in JS runs its executor function regardless of if you're waiting on it or not. `await` blocks if the executor isn't resolved yet, or immediately resolves if it is.

locales/fluent/base/en.ftl
Monkatraz: No, I don't think so. Plus, these are only the same word in English. I don't know if another language would need to use different wording to get the same idea across.

web/app/Http/Controllers/UserController.php
Monkatraz: Good catch. I didn't catch this because my IDE can't type the `where` and `find` methods because of the weird insanity that is Laravel models.

web/app/Services/Deepwell/DeepwellService.php
Monkatraz: lmao yeah tbh I was being lazy because the `options` field was untyped in my IDE (because PHP arrays gdi)

web/resources/lib/components/dashboard/Dashboard.svelte
Monkatraz: Yes, it's a very tiny router.

web/routes/web.php
Monkatraz: lol no, I don't think it changes anything. I'll change it to match the others

deepwell/src/methods/user.rs
emmiegit: Ignoring unknown query parameters are fine but idk about ignoring an invalid value for a real key.

deepwell/src/services/user.rs
emmiegit: Well here it's being used when you're building a response, so there shouldn't be anything after that I think.

deepwell/src/methods/user.rs
Monkatraz: Also, with the `?` operator, what happens with an empty query string? Does `UserDetailsQuery.details` have to be `Option<UserDetails>` or will that implicitly have `default()` called on it?

deepwell/src/services/user.rs
Monkatraz: Sure, but it does feel a bit wrong to "mangle" the `UserModel` object because I'm _always_ assuming you're building a response with it and discarding the `UserModel` immediately.

I can already think of a few scenarios where I bet this will come up, because `UserIdentity` in particular is used all over the API.

emmiegit: I'm not sure what you mean by "mangle", when you are passing it as owned you're destructing it.

Honestly even if it's used elsewhere, `into()` of owned to owned is still preferable imo, because if you need a second copy just do `model.clone().into()`, which makes the cloning explicit.

deepwell/src/methods/user.rs
emmiegit: No if it's empty it gets filled in because you add `#[serde(default)]` to its field in the struct definition.

It says "when trying to deserialize, if this field is missing then just fill it with the default", which is what we want.

As for `?`, it is the same as the `try!()` macro, which is the same as: (effectively)

```rs
let item = try!(result);

// same as

let item = match result {
    Ok(item) => item,
    Err(error) => return Err(error.into()),
};
```

It's syntactic sugar for dealing with operations that can fail.

deepwell/src/services/user.rs
Monkatraz: That solution clones the _entire model_ in order to get a `UserIdentity`, rather than just a few `String` instances. I think it's a unnecessary microoptimization to avoid cloning here.

deepwell/src/methods/user.rs
emmiegit: Hmm, I suppose. We can change it later but I figured returning model data is something I can do, might make use of later, so why not.

I'm mostly suggesting this too because "why not", the generic helper function makes it easy to do, and we can get rid of it later if we think it's not worth it given how we call the route.

deepwell/src/methods/user.rs
Monkatraz: Okay, I'll do what you're suggesting then.

Monkatraz:
thanks @ammongit

deepwell/src/services/user.rs
lambda-fairy: It feels like these UserSomethingOutput structs are all *views* into the underlying UserModel. In that case, would a struct with a lifetime be appropriate here?

That said, I agree that this is not a big deal performance-wise.

emmiegit: You're right, I would've used borrowed strings everywhere if I could. I feel like there was some kind of ownership reason for why that wasn't possible, but I don't remember.

PR #707, opened 2022-01-02
Fix docker package name

emmiegit:
The `docker` package on Debian derived distros is an old weird desktop dock bar people don't use.

h/t @MajorRaccoon for noticing the issue in our docs


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #712, opened 2022-01-03
Update development documentation

emmiegit:
This updates our "getting started" type development docs, since it's been a while since they were updated and have fallen out of date in several respects. It also merges in the `frontend-development.md` file into the main one, since with the current `pnpm`-based deployment system, the differences between what the two files espoused was negligible.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #713, opened 2022-01-03
Use special enum for unset values

emmiegit:
For update methods, we need to distinguish between "the value is unset" and "the value is set, to null", since they have different semantics (i.e., leave value the same, modify value so that it is null).


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #715, opened 2022-01-04
[WJ-871] User Dashboard Part Dos

Monkatraz:
The dashboard isn't done, but this PR is huge and I need to go work on other things before I go insane. But, you can now change your profile settings, some of your account settings (notably not your username, email, or password), and your avatar.

Stuff I need to do later to this dashboard:
- Implement the notifications and messages views
- Signal to the UI that it needs to refresh due to saved changes
- Better selection box components
- Username, email, and password change flows
- Some less boring styling in places


deepwell/src/services/user.rs
```diff
@@ -100,7 +100,7 @@ impl From<&UserModel> for UserInfoOutput {
     fn from(user: &UserModel) -> Self {
         Self {
             identity: UserIdentityOutput::from(user),
-            about: user.about_page.clone(),
+            about: user.bio.clone(),
```
emmiegit, line 103: Shouldn't the field also be renamed to `bio`?

I know it will be a notable amount of effort so I'm not saying it has to be done now, mostly curious about your thoughts on the name.

web/app/Http/Controllers/AccountController.php
```diff
@@ -86,68 +97,189 @@ public function register(Request $request): Response
...
+        if (!$client) {
+            return new Response('', 401);
         }
 
-        $email = $request->input('email');
+        $accepts_invites = (bool) $client->get('receive_invitations');
+        $language = (string) $client->language ?? '';
+        $allow_messages = (string) $client->get('receive_pm');
+
+        // this is a complete guess
```
emmiegit, line 193: I would add a `TODO` here so we can find it later in greps as something to change

web/app/Http/Controllers/AccountController.php
```diff
@@ -86,68 +97,189 @@ public function register(Request $request): Response
...
+        if (!$client) {
+            return new Response('', 401);
         }
 
-        // send recovery email
+        if ($request->has('language')) {
+            $language = (string) $request->input('language');
+            try {
+                // we'll update via deepwell because it can properly handle
+                // language code validation
```
emmiegit, line 234: lol

emmiegit (approved):

deepwell/src/services/user.rs
Monkatraz: I'd rather rename `bio` to `about`, I think. I dunno, we'll need to do some more migrations on the user table so we can bike shed this then.

web/app/Http/Controllers/AccountController.php
Monkatraz: Will do in my next PR (want to get this merged and I'm on my phone lol)

Monkatraz:
thanks @ammongit

PR #720, opened 2022-01-06
Add Docker Compose version to helper

emmiegit:
The lack of an explicit Docker Compose version causes issues if the default compose file version is different, as there will be a version mismatch.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #725, opened 2022-01-07
Overhauled `pnpm dev` script

Monkatraz:
This PR overhauls the `pnpm dev` script, or really just replaces it. It's now a full on TypeScript package that gets ran with `esbuild-runner`. It's been designed to be _very paranoid_ and so it should be better behaved.


docs/development.md
```diff
@@ -80,16 +80,16 @@ You can run development mode using the following:
...
 
-You can just kill the terminal (`CTRL + C` usually) when you want to stop the server. (If you're on Windows, you may want to _hold_ `CTRL + C` to bypass annoying `Terminate batch job?` prompts, and if you do so, make sure to wait for the Docker containers to exit before restarting the development mode.)
+You can just kill the terminal (`CTRL + C` usually) when you want to stop the server.
 
-If for some reason the containers aren't shutting down, you can run the following:
+If for some reason development resources aren't shutting down correctly, you can run the following:
 
 ```sh
-$ pnpm compose stop
+$ pnpm dev clean
```
emmiegit, line 92: `stop` just brings the containers down, `clean` implies they're also deleted right? Or maybe I'm misunderstanding.

docs/modules.md
```diff
@@ -37,3 +37,36 @@ Here is how to use that template:
...
+
+### Publishing
+
+To publish modules, run this command:
+
+```
+$ pnpm modules:publish
+```
+
+This will start the publishing process for any modules whose version is higher than what is currently in the NPM registry. Make sure you are absolutely positive you want to take this step, and make sure you've ran `pnpm modules:version`.
```
emmiegit:
```suggestion
This will start the publishing process for any modules whose version is higher than what is currently in the NPM registry. **Make sure you are absolutely positive you want to take this step**, and make sure you've ran `pnpm modules:version`.
```

web/scripts/dev/cli.ts
```diff
@@ -0,0 +1,215 @@
...
+  stopped = false
+  starting = false
+
+  async clean() {
+    if (this.stopped) return
+    this.stopped = true
+
+    // try to stop anything that might still be running
+    for (const proc of processes) {
+      if (!proc.killed) {
```
emmiegit, line 38: Does this happen after you've sent `SIGTERM` and already given it some time to try and clean up on its own?

web/scripts/dev/cli.ts
```diff
@@ -0,0 +1,215 @@
...
+        // we're gonna _REALLY_ try to kill it
+        // `docker-compose` in particular is a pain
+        proc.kill("SIGINT")
```
emmiegit, line 41: I would suggest waiting a bit after `SIGINT` to see if it's cleaned itself up before proceeding with `SIGKILL`.

web/scripts/dev/cli.ts
```diff
@@ -0,0 +1,215 @@
...
+  private async hijackTerminal() {
+    if (process.stdin.isTTY) {
+      readline.emitKeypressEvents(process.stdin)
+      process.stdin.setRawMode(true)
+    }
+
+    process.stdin.on("keypress", async (_str, key) => {
+      // manually handle Ctrl+C
+      if (key.name === "c" && key.ctrl) {
+        // spamming Ctrl+C again will turn off raw mode,
```
emmiegit, line 195: :+1:

web/scripts/dev/mockoon.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  async close() {
+    if (this.stopped) return
+    this.stopped = true
+    await pnpm("mock-stop", false, "scripts/dev")
+  }
+
+  static async isRunning() {
+    try {
+      // have to use `cmdAsync`instead of `pnpm` because the `-s` flag
```
emmiegit:
```suggestion
      // have to use `cmdAsync` instead of `pnpm` because the `-s` flag
```

web/scripts/dev/vite.ts
```diff
@@ -0,0 +1,41 @@
...
+import { error } from "../pretty-logs"
+import { isBuild, isClean, isServe } from "./util"
+
+export class Vite {
+  declare instance: ViteDevServer
+
+  stopped = false
+
+  static async clean() {
+    // TODO: somehow check if Vite is running?
```
emmiegit, line 11: Recording pid on vite start? Checking the `3000` port?

emmiegit (approved):

emmiegit:
Thanks a lot for handling all this, sorry there have been so many build tool issues lately.

docs/development.md
Monkatraz: Clean is supposed to mean like "clean up" but not "reset". I suppose that could be confusing

web/scripts/dev/cli.ts
Monkatraz: I can't really tell if the process has ended or not. I have to attach an event listener but that event may never fire. `killed` just indicates if the process has been told to close.

web/scripts/dev/cli.ts
Monkatraz: What do you mean?

web/scripts/dev/vite.ts
Monkatraz: Vite isn't running another process. It's being ran directly. If Vite is still running, that's because there is _another_ `pnpm dev` running. I could check the port but I wouldn't be able to get the PID, I don't think.

Monkatraz:
thanks @ammongit

PR #727, opened 2022-01-08
[WJ-920] Move page links to DEEPWELL

emmiegit:
This PR removes `PageConnection` and similar model files from PHP and adds links and connections to DEEPWELL, which are updated via a specific endpoint. Now, all link updating occurs outside of PHP.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #728, opened 2022-01-08
[WJ-920] Replace PageContents with text table

emmiegit:
This removes the `PageContents` model and its `page_contents` table and replaces it with the `text` table pattern. This is a pattern for deduplicating large strings: it is a content-addressable (SHA-512) table that stores the string in question. A `CHECK` constraint ensures that the hash matches the contents for each row. A `UNIQUE` constraint is _not_ used because we do not directly query the `contents` column, and an index there would cause performance to suffer needlessly. The `CHECK` and `PRIMARY KEY` constraints already ensure uniqueness together (unless of course in the unlikely event that a SHA-512 collision is found).

Instead of storing the page source and compiled HTML again for every revision, it simply stores the hash, which references the same row in the table. The previous implementation meant that all revisions, including those which did not affect the wikitext, would result in that potentially long string being stored again.

There is presently a hack when it comes to saving revision data because `Outdater` needs a row to already exist, which it then updates separately. This will go away when we've moved `Outdater` functionality to DEEPWELL.

The useful utilities in `PageContents` and some other classes now exist in `PageService`, which makes the correct DEEPWELL calls in the background. I'm not sure what pattern we'll use in the long run, but I imagine it'll be something out of `\Wikijump\Services\Deepwell` where there are a few simple, high-level calls that handle the things the web server wants done.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #745, opened 2022-01-16
[WJ-920] Migrate primary page tables to DEEPWELL

emmiegit:
Apologies in advance for a very monstrous pull request. Unfortunately there wasn't much choice but to do everything in one go, since pages are a pretty fundamental concept in the code, and the related concepts (e.g. page categories) also needed to be addressed at the same time.

This PR reworks Wikijump's concept of pages entirely. Ownership and management of pages has moved entirely into DEEPWELL, and relevant methods have been added to interface with them. Follow-up Jira issues have been created for other revision operations which still need to be implemented. (See [WJ-982](https://scuttle.atlassian.net/browse/WJ-982)).

For information on the changes to the database schema regarding pages and related entities, see [WJ-920](https://scuttle.atlassian.net/browse/WJ-920).

The following code additions have been made:
* Adding a separate "internal" API version scheme. This is what's used by `DeepwellService` in PHP.
* Adds `JobService`, which runs tasks asynchronously. Presently the only job type available is rerendering a page.
* Adds `OutdateService`, which rerenders descendant pages depending on the source page and what kind of action was performed on it. Compare with [`Outdater.php`](https://github.com/scpwiki/wikijump/blob/c7b5611ae531b0225e1d9e42174f143ff434140a/web/php/Utils/Outdater.php).
* Adds `RenderService`, which invokes ftml.
* Adds `CategoryService`, which does CRUD for page categories. Most notable is `get_or_create()` which is invoked on page creation.
* Adds `SiteService`, which allows retrieving site data from the database. Read-only, since it's still owned by Ozone.
* Adds `ParentService`, a stub. No support for parents is currently in the code, this will be added later.
* "Direct" API methods. These permit retrieving an object using only its ID. These are intended as a temporary addition for old Wikidot code, and will be deprecated when those usages are no longer present. Direct methods should not be exposed publicly.
* The `PageService` and `RevisionService` has multiple stubs added for new operations.
* The concept of `RevisionTasks` is added, which determines what operations need to be carried out given what changes are present when a revision is being created. Compare with `OutdateService` (mentioned above) and `Outdater.php`.
* Retrieving page data via API returns a special object which gives some revision information in addition to the data in the `page` table.
 * This also includes the ability to pass in query parameters to modify the level of detail sent, similar to retrieving user information.
* Adds Ozone-owned forum tables to DEEPWELL's model file generation.
* Removes the `theme_preview` table and associated classes.
* Removes the `category` table and associated classes. The replacement, DEEPWELL category does not store configuration there, so a lot of things are currently on hardcoded defaults while we transition.
* Modify backlink reading and updating code.
* Adds DEEPWELL model classes in PHP for receiving and working with data received via the API.
* Adds temporary utility functions to convert JSON. Unfortunately, Sea-ORM does not support Postgres arrays, so for now we're doing these annoying conversions for JSON data in the database.
* Adds utility functions to process slugs with categories.
* Updates DEEPWELL and ftml to use directly-interpolated variables in string formatting, when sensible.
* Updates the local deploy ratelimit-exemption key.
 * This kept triggering the "you are committing secrets" check, when this is not meant to be private or sensitive.
* Adds `wikijump-health-check` as a helper script for container health checks.
* Lots of random PHP changes and deletions to get things to work and clean up dead code.

Some things are left undone here. For instance:
* `PageRevision` PHP classes and uses. A lot of the logic involving revisions in PHP is now obsolete or needs to be reimplemented.
* `PageRateVote` PHP classes and uses. [WJ-995](https://scuttle.atlassian.net/browse/WJ-995) covers his work.


deepwell/src/methods/mod.rs
```diff
@@ -18,21 +18,42 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
+//! Definitions of methods invoked by different
```
Monkatraz: Think you meant to remove this comment

deepwell/src/methods/page.rs
```diff
@@ -123,53 +227,114 @@ pub async fn page_links_to_missing_get(req: ApiRequest) -> ApiResponse {
...
+    let category =
+        CategoryService::get(ctx, page.site_id, Reference::from(page.page_category_id))
+            .await?;
+
+    // Get text data, if needed
+    let wikitext = get_text(ctx, details.wikitext, &revision.wikitext_hash).await?;
+    let compiled_html =
+        get_text(ctx, details.compiled_html, &revision.compiled_hash).await?;
+
+    let output = GetPageOutput {
```
Monkatraz, line 310: I think this would be more elegant if it used a `From` trait? `GetPageOutput::From(page)` or whatever

web/.gitignore
```diff
@@ -6,6 +6,7 @@ conf/wikijump.ini
 /vendor
 composer.phar
 node_modules
+.pnpm-debug.log
```
Monkatraz, line 9: idk why I resisted doing this for so long

web/app/Http/Controllers/PageController.php
```diff
@@ -36,21 +38,15 @@ public function show(): View
...
             $tags = $values['tags'] ?? null;
-            $revision = $page->getRevisionNumber();
-            $timestamp = $page->getDateLastEdited()->getTimestamp();
-            $page_title = $page->getTitleOrUnixName();
-            $title = $page_title;
-            $social_title = $page_title;
+            $revision = $page->revision_number;
+            $timestamp = $page->lastUpdated()->getTimestamp();
+            $title = $page->title;
+            $license = LicenseMapping::get('cc_by_sa_3'); // TODO hardcoded
```
Monkatraz, line 44: Is this still meant to be hardcoded?

web/app/Services/Deepwell/Models/DeepwellModel.php
```diff
@@ -0,0 +1,8 @@
+<?php
+declare(strict_types=1);
+
+namespace Wikijump\Services\Deepwell\Models;
+
+abstract class DeepwellModel
+{
+}
```
Monkatraz: What's this for?

Monkatraz (approved):
I ignored everything in `web/web`, `web/templates`, `web/php`, etc. I hit "viewed"  on _147_ files before I realized my browser scrollbar was only a quarter of the way down.

But, it's pretty clear that this makes pages not horrifying! I have so much work to do now...

deepwell/src/methods/page.rs
emmiegit: It actually originally was that, but at present it pulls from `PageModel`, `PageRevisionModel`, and two `Option<String>`s, so the `From` would be kind of messy (probably a tuple of all of them?)

web/app/Http/Controllers/PageController.php
emmiegit: Yes it's a category setting, which doesn't exist anymore, so we're just setting it to until we move license choice into site settings.

web/app/Services/Deepwell/Models/DeepwellModel.php
emmiegit: It's the base class for DEEPWELL models (basically outputs you get from the internal API). There isn't really anything specific to abstract I guess, but I still think it's worth having a parent so there's something in common in the class type system.

That said I did add the `id()` method since that is standard between them actually.

emmiegit:
thanks @Monkatraz

PR #759, opened 2022-01-18
Use replacement script instead of artisan key generation

emmiegit:
This replaces the `php artisan key:generate` command from the local container build only, to avoid issues with the PHP environment being incorrect and preventing a run of the key generation command. Since all this does is generate some random base64, we don't actually need to load the entire PHP package ecosystem to do this. This avoids an issue where the build will fail if an unrelated issue in the PHP environment prevents it from loading, which is already a difficult class of issue to debug. (The `php artisan config:cache` command is also removed because it is not required and would also have the same PHP issues).

### DIFF
--- a/install/files/php-fpm/setup-key.sh
+++ b/install/files/php-fpm/setup-key.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
 set -eux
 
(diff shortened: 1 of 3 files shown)


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #760, opened 2022-01-19
Revert "Use replacement script instead of artisan key generation"

emmiegit:
Reverts scpwiki/wikijump#759

The `APP_KEY` is more than just random bytes, it's formatted in some way to also represent the algorithm. Because it's more than trivial, we should restore the original PHP scripts and figure out dependency issues directly instead.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #831, opened 2022-02-28
Improve Sheaf's performance, fix bugs

Monkatraz:
This PR is mostly focusing on Sheaf's performance. This involved a lot of different changes. Sorry for the big PR.

Changes:
* Fix CodeMirror package duplication issues - required shameful dependency hoisting and manual `resolve.dedupe` configuration
* Fixed Sheaf's live preview not actually being live
* Migrate all workers to a new, slicker Comlink based system that is much simpler to use and much smaller than Threads
* Add a basic module worker polyfill so that _some_ workers kind of work on Firefox in development mode. This is probably temporary as Firefox is seemingly going to be getting module workers soon.
* Fix formatting issues with `<pre>` blocks in Svelte components
* Updated the `cm.ts` CodeMirror exports to be up to date
* Extended the `@wikijump/util` `animationFrame()` and `idleCallback()` methods to take executor functions
* Added a good `requestIdleCallback` polyfill that actually provides some performance benefit on Safari
* Updated the `@wikijump/ftml-components` code and math blocks to render on an idle callback
* Massively improve Sheaf live preview performance via CSS containment changes
* Added a super fast string UTF-8 byte length function to `@wikijump/util`, which is used to very quickly get the text size of Sheaf's contents
* Switch to using the tiny `micromorph` library for live preview morphing
* Moved the Prism highlighter off the main thread - now the library will only be downloaded if a code block has to be highlighted, which will be done asynchronously
* Updated ftml-wasm bundle
* Fixed a bunch of misc. bugs

Next PR will probably be me fixing a bunch of issues that are cropping up with production builds of the Vite stuff.


web/modules/cm-espells/src/MisspellingTooltip.svelte
```diff
@@ -102,9 +102,9 @@
 
   <hr />
 
-  <pre class="code cm-spellcheck-tip-source">
-      <code>{$t("#-source", { slice: word.word, from: word.from, to: word.to })}</code>
-  </pre>
+  <pre class="code cm-spellcheck-tip-source"><code
+      >{$t("#-source", { slice: word.word, from: word.from, to: word.to })}</code
+    ></pre>
```
emmiegit, line 107: lol, clever workaround to avoid whitespace

web/modules/cm-espells/src/espells/interface.ts
```diff
@@ -0,0 +1,128 @@
...
+  protected async createWorker() {
+    if (DICTIONARIES.hasOwnProperty(this.locale)) {
+      const { aff: affURL, dic: dicURL } = await DICTIONARIES[this.locale]()
+
+      // we can't use the static fromURL method on Espells because of Comlink limitations
+      // so it has to be reimplemented basically - we have to resolve the urls
+
+      const aff = await fetch(affURL).then(res => res.text())
+      const dic = Array.isArray(dicURL)
+        ? await Promise.all(dicURL.map(dic => fetch(dic))).then(res =>
```
emmiegit: This is a bit difficult to read, maybe worth adding a comment or two explaining what the parts of the callbacks are doing?

web/modules/comlink/src/index.ts
```diff
@@ -0,0 +1,124 @@
...
+   * automatically handles binding methods and worker creation.
+   *
+   * @param methods - The methods to bind to the class instance, which when
+   *   called will be passed the worker instance.
+   */
+  static of<T>(
+    methods: (keyof T)[]
+  ): abstract new () => AbstractWorkerBase<T> & Comlink.RemoteObject<T> {
+    // @ts-ignore
+    return class extends AbstractWorkerBase<T> {
```
emmiegit, line 56: You can do this? Incredible

web/modules/sheaf/src/components/Wikitext.svelte
```diff
@@ -70,53 +79,54 @@
...
+        element.innerText = ""
+        element.appendChild(fragment)
+      }
+    })
+
+    await animationFrame(() => {
+      // prepend style with a index comment so that each style string is unique
+      stylesheets = styles.map(
+        (style, idx) => `\n/* stylesheet ${idx + 1} */\n\n${style}\n`
+      )
```
emmiegit, line 127: Interesting, why is this necessary?

web/modules/sheaf/src/core.ts
```diff
@@ -72,7 +72,8 @@ export class SheafCore {
   }
 
   private update(update: ViewUpdate) {
-    if (!update.docChanged && !update.selectionSet) return
+    // if (!update.docChanged && !update.selectionSet) return
```
emmiegit, line 77: Why is this code commented out?

emmiegit (approved):

web/modules/cm-espells/src/MisspellingTooltip.svelte
Monkatraz: Yeah, that's what Prettier does. It is a cool workaround.

web/modules/comlink/src/index.ts
Monkatraz: yeah honestly when I got this horrifying thing working I was very surprised

web/modules/cm-espells/src/espells/interface.ts
Monkatraz: Yeah, or I should maybe rewrite it. It's just trying to be fast by doing all the fetch requests in parallel.

web/modules/sheaf/src/components/Wikitext.svelte
Monkatraz: The stylesheets are uniquely identified by their contents. What is uniquely identifying them is Svelte - below in the markup there is a `{#each}` block that is putting the stylesheet strings into the `<head>`. When the stylesheets array updates, it compares the previous list to the new one.

If you had two stylesheets with the same contents, it would confuse Svelte.

web/modules/sheaf/src/core.ts
Monkatraz: Ah, oops. That's temporary. With the original check, moving your cursor at all would cause the FTML to get rerendered. For now, the editor state will only update if the document changes.

Later, that will need to be properly fixed, because if we add sourcemapping there will need to be a bunch of editor state that updates with scrolling, active lines, etc.

Monkatraz:
thanks @ammongit

PR #838, opened 2022-02-28
Forbid "example" names.

emmiegit:
This change makes it so sites and users cannot be called `example`, or choose a name which starts with `example`. This way the name is reserved for documentation.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #839, opened 2022-03-02
Comlink wrapper improvements, more Sheaf bugfixes, performance boosts

Monkatraz:
Does what it says on the tin, basically.

In detail:
* Added timeouts for bound functions on `AbstractWorkerBase`
* Added the ability to provide fallback values for `AbstractWorkerBase`
* Limited bound functions to methods and classes for `AbstractWorkerBase`
* Added checks for `matchMedia` where ever it's used
* Added a new `wordCount` function to `ftml-wasm`, which uses the FTML AST to count the number of visible words
* Got rid of the `RenderHandler` class and the `Wikitext` component, replacing them with the faster and simpler `MorphingWikitext` component
* Added the `encode` and `decode` utility functions to `@wikijump/util`
* Made it so that the `Timeout` class has an _optional_ callback


web/modules/comlink/src/index.ts
```diff
@@ -17,62 +22,112 @@ export type {
...
+          throw new Error(`Method "${method}" timed out!`)
         }
       }
     }
+
+    return Derived
+  }
+
+  /** Tries to run a default method if the worker couldn't be started. */
+  private async _baseTryToGetDefault(method: keyof Methods<T>, args: any[]) {
```
emmiegit, line 118: Why is this needed?

web/modules/sheaf/src/components/MorphingWikitext.svelte
```diff
@@ -0,0 +1,105 @@
...
+    const result =
+      typeof wikitext === "string"
+        ? await FTML.renderHTML(wikitext, undefined, "draft")
+        : await wikitext.render()
+
+    timeCompile = measureCompile()
+    return result
+  })
+
+  const patch = createLock(async ({ html, styles }: RenderedHTML) => {
```
emmiegit, line 58: For my information, what does the `createLock` utility do?

web/modules/sheaf/src/state.ts
```diff
@@ -27,30 +53,121 @@ export class SheafState {
...
+  /** Retrieves the value of the editor as an `Uint8Array`. */
+  @Memoize()
   async buffer() {
     return await textBuffer(this.doc)
   }
 
+  /** Renders the editor with FTML. */
+  @Memoize()
+  async render() {
+    return await FTML.renderHTML(await this.value(), undefined, "draft")
```
emmiegit, line 71: Are these methods memoized because the value doesn't change, or is Typescript invalidating the cached result if any of the fields update?

emmiegit (approved):

web/modules/comlink/src/index.ts
Monkatraz: I wanted an easy way to provide fallback values if the worker can't start. It's a bit messy to do a override function and check for this yourself, so instead I provided a specific object that you can use for providing fallbacks. This function tries to run/retrieve a fallback value for a worker method, if it exists.

web/modules/sheaf/src/components/MorphingWikitext.svelte
Monkatraz: Spits out a function that when called multiple times, will wait for each previous call to finish. Basically, only one instance of the function can be active at a time.

web/modules/sheaf/src/state.ts
Monkatraz: The `SheafState` class is a immutable container basically. A new one is made when the editor updates. The idea with these functions being memoized is that if the editor state isn't changing but you're still requesting stuff from it, it won't do any unnecessary work.

An example of this is when you switch around the different preview tabs, like Result, HTML, back to Result, etc. It'll just use the previous rendered result.

Monkatraz:
thanks @ammongit

PR #846, opened 2022-03-03
[terraform] Make API ratelimit secret available in container

emmiegit:
The `api_ratelimit_secret` secret has been added to Terraform Cloud, so now we can make this value available in-container as an environment variable.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #847, opened 2022-03-03
Force DEEPWELL image build

emmiegit:
This should hopefully get an image built and sent to the ECR container so that we can deploy.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #848, opened 2022-03-04
[terraform] Connect php-fpm container to nginx

emmiegit:
The PHP-FPM container needs the ability to connect to nginx container so it can load the JS build manifest file. Locally this works but in our deployment it does not. Hopefully this change is sufficient to permit this.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #856, opened 2022-03-07
[WJ-966] Change ftml logging from slog to log

emmiegit:
I previously used `slog` (structured logging) since I thought it would be easier to read with a lot of information, but in practice the inherited key-value logging kind of made anything non-trivial a bit difficult to read. And I was thinking about whether a slog-to-log adapter may have been necessary, since `tide` uses `log` anyways, and so the root logger would not be the same kind so I may as well move over.

There are a lot of small annoying changes in the PR, apologies. Additionally, I remove a bunch of the features for logging, including the `console.log()` implementation, since they either weren't used, or the intended effect (lack of logging not impacting performace) are part of `log` itself.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #858, opened 2022-03-08
[WJ-996] Decommission FFI and Text_Wiki

emmiegit:
With parsing and rendering now happening via DEEPWELL, which can use ftml natively as it's also Rust, a lot of weird workarounds from FFI we've had are no longer needed.

This PR gets rid of a few things:
* ftml's FFI feature and associated code
* The `FtmlFfi` class and associated code
* The `WikitextBackend` pattern and associated classes
* The `WikiTransformation` class, which invokes `Text_Wiki`
* The legacy `Text_Wiki` library, which without the above cannot be used
* The wikitext backend feature flag

All existing calls to `WikitextBackend` have been replaced with calls to a `DeepwellService` stub method.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #860, opened 2022-03-08
[WJ-964] Implement compatibility IDs

emmiegit:
This PR alters prior migrations. You will need to recreate your local database.

As part of ensuring Wikidot compatibility, we want to ensure that Wikidot IDs can be accepted by Wikijump (if the corresponding data is imported that is). This is important because some legacy URLs include Wikidot IDs in them, and we do not want an ambiguous situation where some ID value could refer to one item in Wikidot and a different item in Wikijump.

Here, we set the starting value to a chosen high value that is beyond what Wikidot in production will likely ever reach. This way we can assign our own IDs without ever worrying about collisions with existing Wikidot items.

The default user IDs have low values since they are special constant users.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #861, opened 2022-03-09
[WJ-1004] Basic page API (only `pageGet`), render pages on `next` frontend without `RunData`

Monkatraz:
Does what it says on the tin. Next PR might be me finishing up the page API endpoints, or adding the "View Source" button. Also, 404s work correctly on the `next` frontend too.

### DIFF
--- a/ftml/Cargo.lock
+++ b/ftml/Cargo.lock
@@ -532,9 +532,9 @@ dependencies = [
 
 [[package]]
 name = "libz-sys"
-version = "1.1.4"
+version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "df2bf61678a0a521c3f7daf815d2e6717d85a272a7dcd02c9768272b32bd1e2a"
+checksum = "6f35facd4a5673cb5a48822be2be1d4236c1c99cb4113cab7061ac720d5bf859"
 dependencies = [
  "cc",
  "libc",
@@ -795,9 +795,9 @@ checksum = "d22b73985e369f260445a5e08ad470117b30e522c91b4820585baa2e0cbf7075"
 
 [[package]]
 name = "regex"
-version = "1.5.4"
+version = "1.5.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d07a8629359eb56f1e2fb1652bb04212c072a87ba68546a04065d525673ac461"
+checksum = "1a11647b6b25ff05a515cb92c365cec08801e83423a235b51e231e1808747286"
 dependencies = [
  "aho-corasick",
  "memchr",
--- a/ftml/Cargo.lock
+++ b/ftml/Cargo.lock
@@ -338,7 +338,6 @@ dependencies = [
  "lazy_static",
  "log",
  "maplit",
- "parking_lot",
  "pest",
  "pest_derive",
  "proptest",
@@ -349,7 +348,6 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_repr",
- "slog-bunyan",
  "sloggers",
  "str-macro",
  "strum",
@@ -426,17 +424,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "62b467343b94ba476dcb2500d242dadbb39557df889310ac77c5d99100aaac33"
 dependencies = [
  "libc",
-]
-
-[[package]]
-name = "hostname"
-version = "0.3.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3c731c3e10504cc8ed35cfe2f1db4c9274c3d35fa486e3b31df46f068ef3e867"
-dependencies = [
- "libc",
- "match_cfg",
- "winapi",
 ]
 
 [[package]]
@@ -556,15 +543,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "lock_api"
-version = "0.4.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "88943dd7ef4a2e5a4bfa2753aaab3013e34ce2533d1996fb18ef591e315e2b3b"
-dependencies = [
- "scopeguard",
-]
-
-[[package]]
 name = "log"
 version = "0.4.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -578,12 +556,6 @@ name = "maplit"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3e2e65a1a2e43cfcb47a895c4c8b10d1f4a61097f9f254f183aee60cad9c651d"
-
-[[package]]
-name = "match_cfg"
-version = "0.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ffbee8634e0d45d258acb448e7eaab3fce7a0a467395d4d9f228e3c1f01fb2e4"
 
 [[package]]
 name = "matches"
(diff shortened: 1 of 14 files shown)


ftml/Cargo.lock
```diff
@@ -338,7 +338,6 @@ dependencies = [
  "lazy_static",
  "log",
  "maplit",
- "parking_lot",
```
emmiegit, line 341: Why is this file getting changed?

web/app/Helpers/LegacyTools.php
```diff
@@ -116,6 +104,26 @@ public static function generateScreenVars()
...
+    {
+        /**
+         * Create a RunData instance.
+         */
+        $runData = new RunData();
+        $runData->init();
+        Ozone::setRunData($runData);
+        Log::debug('runData object created and initialized in LegacyTools::generateScreenVars()');
+
+        $site = static::getCurrentSite();
```
emmiegit, line 125: Is there a reason you're doing `static::xxx` instead of `self::x`? Is this something prettier is doing?

web/app/Http/Controllers/PageController.php
```diff
@@ -4,19 +4,205 @@
...
+        }
+
+        $site_id = $site->getSiteId();
+
+        $page = null;
+
+        if ($path_type === 'slug') {
+            $page = Page::findSlug($site_id, $path, $wikitext, $html);
+        } elseif ($path_type === 'id') {
+            $page = Page::findId($site_id, (int) $path, $wikitext, $html);
```
emmiegit, line 97: Huh, I didn't realize you could just cast. I've been using `intval()`, I should probably change that.

web/app/Http/Controllers/PageController.php
```diff
@@ -4,19 +4,205 @@
...
+
+            if ($updater !== null) {
+                $updater = $updater->toApiArray($avatars);
+            }
+
+            $output = array_merge($output, [
+                'id' => $page->id,
+                'slug' => $page->slug,
+                'category' => $page->page_category_id,
+                'parent' => null, // TODO
```
emmiegit, line 150: Should be:
```suggestion
                'parents' => [], // TODO
```

web/app/Http/Controllers/PageController.php
```diff
@@ -4,19 +4,205 @@
...
+                'children' => [], // TODO
+                'title' => $page->title,
+                'altTitle' => $page->alt_title,
+                'tags' => $page->tags,
+                'score' => 0, // TODO
+                'created' => $page->page_created_at,
+                'creator' => null, // TODO
+                'updated' => $page->lastUpdated(),
+                'updater' => $updater,
```
emmiegit, line 159: This isn't necessarily true, `$updater` here is the last person to make a revision on the page, which may not be why the page was most recently updated. (Though in our current implementation this assumption holds afaik)

If you want "last revision time" then you want `revision_created_at`, not `page_updated_at`.

web/app/Http/Controllers/PageController.php
```diff
@@ -4,19 +4,205 @@
...
+
+        if ($output_type['wikitext']) {
+            $output = array_merge($output, [
+                'wikitext' => $page->wikitext,
+            ]);
+        }
+
+        if ($output_type['html']) {
+            $output = array_merge($output, [
+                'html' => $page->compiled_html,
```
emmiegit: For single fields, isn't it more efficient just to do
```php
$output['html'] = $page->compiled_html;
```
?

web/app/Http/Controllers/PageController.php
```diff
@@ -4,19 +4,205 @@
...
-    public function show(): View
+    public function show(?string $path = null): View
     {
-        $values = LegacyTools::generateScreenVars();
+        // TODO: description, image, twitter, etc.
+        // TODO: site theming
+        // TODO: favicons
+        // TODO: header image/text + subtitle management
+        // TODO: navbar items
+        // TODO: private sites
```
emmiegit, line 182: Oof not really looking forward to this. It's not used a lot and will involve us really getting into the details of permissions and starting implementations there.

That said, I do look forward to actually making a cohesive permission system that is extensible and isn't brittle.

web/app/Http/Controllers/PageController.php
```diff
@@ -4,19 +4,205 @@
...
+        if ($site === null) {
+            abort(404);
+        }
+
+        $slug = LegacyTools::redirectToNormalUrl($site, $path ?? '', '');
+
+        $site_id = $site->getSiteId();
+
+        $page = Page::findSlug($site_id, $slug, false, true);
+        if ($page === null || $page->page_deleted_at !== null) {
```
emmiegit: This check is redundant. When getting pages there is already a filter for non-deleted pages. Fetching deleted pages requires either the direct fetch method (using page ID), or the fetch all method with deleted pages as part of your filter conditions.
```suggestion
        if ($page === null) {
```

Also consider that for deleted pages, there isn't necessarily one page for that slug anyways. I can create `/abc`, delete it, create another `/abc`, and delete that. Which one is the "real" `/abc`?

web/app/Http/Controllers/PageController.php
```diff
@@ -28,45 +214,29 @@ public function show(): View
         $timestamp = null;
         $tags = null;
 
-        if ($values['site']) {
-            $title = $values['site']->getName();
+        $title = $site->getName();
+
+        // TODO: this doesn't seem elegant?
+        $sidebar_page = Page::findSlug($site_id, 'nav:side', false, true);
```
emmiegit, line 216: It's not no, and right now we're hardcoding `nav:(top|side)` as the navigation pages. (Wikidot has them configurable, though honestly I'm not sure inherently how valuable being able to configure that is lol)

We will eventually want a utility that can pull navigation data for a site as a single encapsulated function. (Wikidot doesn't have that because it does a query every time for what the navigation pages are...)

web/app/Http/Controllers/PageController.php
```diff
@@ -28,45 +214,29 @@ public function show(): View
...
 
-        if ($values['wikiPage']) {
-            $page = $values['wikiPage'];
+        $page_content = $page->compiled_html;
+        $tags = $page->tags;
+        $revision = $page->revision_number;
+        $timestamp = $page->lastUpdated()->getTimestamp();
+        $title = $page->title;
+        $alt_title = $page->alt_title;
+        $social_title = $title;
```
emmiegit, line 227: "social"?

web/app/Services/Deepwell/Models/User.php
```diff
@@ -3,17 +3,19 @@
...
-    private string $detail;
-    private int $id;
-    private string $username;
-    private string $tiny_avatar;
-    private int $karma;
+    public string $detail;
+    public int $id;
+    public string $username;
+    public ?string $tiny_avatar;
+    public int $karma;
```
emmiegit, line 18: Oh I forgot to do this, thanks for catching this.

web/resources/views/next/wiki/page.blade.php
```diff
@@ -61,11 +65,6 @@
             </div>
         @endif
 
-        {{-- TODO: proper 404 page --}}
-        {{--
-            Wikidot handles missing pages basically by showing hardcoded content
-            in the $page_content variable. That is stupid.
```
emmiegit, line 67: lmao it did?

emmiegit (approved):

ftml/Cargo.lock
Monkatraz: I don't know, actually. I tried to match the lock on develop (or whatever it is with slightly updated dependencies) but it does seem pretty different.

web/app/Helpers/LegacyTools.php
Monkatraz: Uhhhhh I just do `static` I don't know if there is actually a difference

web/app/Http/Controllers/PageController.php
Monkatraz: It's matching the API spec, which hasn't changed yet

web/resources/views/next/wiki/page.blade.php
Monkatraz: yes. thank god it's dead. we will have a proper 404 page.

web/app/Http/Controllers/PageController.php
Monkatraz: Title that shows up on social media websites.

Monkatraz:
thanks @ammongit

PR #862, opened 2022-03-09
[WJ-980] Add document describing Wikijump's data limits

emmiegit:
The document will need to be grown as the platform does. The aim is to describe what the upper limits on what amounts of data can be stored.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #867, opened 2022-03-11
[WJ-999] Add build check for PHP integer width

emmiegit:
This adds a step in our `php-fpm` Dockerfiles to ensure that `PHP_INT_SIZE` (which gives integer width in bytes) is 8, meaning 64-bit. Because we depend on 64-bit IDs it is important that builds do not succeed in a 32-bit environment.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #868, opened 2022-03-12
[WJ-997] Removing english in text renderer

emmiegit:
The text renderer became rather descriptive for complex elements (e.g. images), which helped readability by adding English words. However this is counter to our internationalization policy, and instead of adding a million small words to our localization system for each, I'm just removing them and making this more internal content-based as it should be.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #876, opened 2022-03-17
Use `cross-env` to set DOCKER_BUILDKIT

Monkatraz:
This fixes running the compose commands on Windows.

### DIFF
--- a/web/svelte.config.cjs
+++ b/web/svelte.config.cjs
@@ -3,7 +3,11 @@ const mergeQueries = require("postcss-merge-queries")
 const autoprefixer = require("autoprefixer")
 const { globalStyle, postcss, scss, typescript } = require("svelte-preprocess")
 
-const abstracts = path.resolve(__dirname, "resources/css/abstracts.scss")
+let abstracts = path.resolve(__dirname, "resources/css/abstracts.scss")
+// removing the drive letter and normalizing the slashes is required for Windows,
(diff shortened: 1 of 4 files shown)


web/svelte.config.cjs
```diff
@@ -3,7 +3,11 @@ const mergeQueries = require("postcss-merge-queries")
 const autoprefixer = require("autoprefixer")
 const { globalStyle, postcss, scss, typescript } = require("svelte-preprocess")
 
-const abstracts = path.resolve(__dirname, "resources/css/abstracts.scss")
+let abstracts = path.resolve(__dirname, "resources/css/abstracts.scss")
+// removing the drive letter and normalizing the slashes is required for Windows,
+// which is absurd but that's just what Sass wants I guess
+abstracts = abstracts.replace(/^[A-Z]:/, "")
+abstracts = abstracts.replaceAll("\\", "/")
```
emmiegit, line 10: Technically this interferes with *nix platforms since backslashes are permitted as characters, but since that's not portable for Windows anyways it doesn't really matter.

emmiegit (approved):

Monkatraz:
thanks @ammongit

web/svelte.config.cjs
Monkatraz: huh, interesting, did not know that

PR #880, opened 2022-03-20
Disallow usernames that look like IP addresses

emmiegit:
We aren't supporting IP-based ("anonymous") edits, but it is still avoidable confusion if a user chooses a username which is (or appears as) an IP address. These regular expressions deny such usernames.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #902, opened 2022-04-01
Fix head type for checkbox

emmiegit:
Another oversight from when I added checkboxes: I copied the same config for both even though radio buttons require a name and checkboxes don't have one at all.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #910, opened 2022-04-06
[WJ-852] Add anchor block to ftml

emmiegit:
This adds the [`[[# anchor-name-block]]` syntax](https://www.wikidot.com/doc-wiki-syntax:links#anchors) syntax to ftml. This is not to be confused with `[#anchor Label here]` (links that point to these anchors), or `[[a]]` (constructing `<a>` elements).


ftml/src/parsing/parser.rs
```diff
@@ -424,6 +424,69 @@ impl<'r, 't> Parser<'r, 't> {
         (first, second, third)
     }
 
+    // Helpers to get individual tokens
+    pub fn get_token(
```
emmiegit, line 428: Not new methods, they were moved from `src/parsing/rule/impls/block/parser.rs`

emmiegit:
self-merge

PR #931, opened 2022-04-19
[WJ-832] Add ID isolation

emmiegit:
Wikidot adds `u-` prefixes to IDs specified by users to ensure uniqueness. However, because this has a few problems, this behavior was not previously copied to ftml as-is:

* The application is inconsistent. Elements in some blocks would have it added while others would not.
* The prefixing itself was incomplete, simply prepending `u-` to the string. So an `id="abc def"` would become `id="u-abc def"`, causing `def` to not be isolated in the same way.
* There are several syntax elements that allow interfacing with IDs directly. If I create a link to `[#apple Anchor]`, because of a `[[#apple]]`, the link won't work because it's not `u-apple`. Or, the `[[module CSS]]`, where all IDs in rules are specified directly, bypassing the `u-` prefixing requirement. (Neither of these are changed in this PR, it would be up to the user to add `u-`s where appropriate if the option is enabled.

This implements `u-` ID prefix isolation in ftml. It is set via a wikitext setting and not enabled by default.

### DIFF
--- a/ftml/src/id_prefix.rs
+++ b/ftml/src/id_prefix.rs
@@ -1,0 +1,39 @@
+/*
+ * id_prefix.rs
+ *
+ * ftml - Library to parse Wikidot text
+ * Copyright (C) 2019-2022 Wikijump Team
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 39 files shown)


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #937, opened 2022-04-22
[WJ-984] Add page deletion to DEEPWELL

emmiegit:
This implements the previously-stubbed method that allows for deleting pages via DEEPWELL. This works by adding a "tombstone" revision where the only change is the metadata field marking it as deleted. I consulted Wikidot's [`Deleter.php`](https://github.com/scpwiki/wikijump/blob/legacy/php/utils/Deleter.php) implementation for this, but there wasn't anything to carry over actually since we're not hard-deleting.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #938, opened 2022-04-22
Add tests for category slug parsing, fix up category handling

emmiegit:
Also adds (the possibly unneeded?) function `slug_is_valid()`, fixes `trim_default()`'s behavior with non-single-colon categories, and adds some tests.

This also enables non-single-colon categories, such as the category `abc:def` in `abc:def:ghi`. Previous the _first_ colon was used to separate the category/page, which I think is less useful because if you see `archived:component:page`, while [multicategories are _not_ supported](https://scuttle.atlassian.net/browse/WJ-355) (that said, WJ-355 says we should redirect `abc:def:ghi` -> `abc-def:ghi`? which makes sense but is out of scope for this?) it allows categories to nominally exist with that name, though they do not exhibit any kind of "parent"-like relationship.

### DIFF
--- a/deepwell/src/web/category.rs
+++ b/deepwell/src/web/category.rs
@@ -77,9 +77,31 @@ fn test_split_category() {
     }
 
     // Proper
+    check!("foo-bar", None, "foo-bar");
     check!("apple", None, "apple");
     check!("banana:apple", Some("banana"), "apple");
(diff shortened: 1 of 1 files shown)


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #940, opened 2022-04-23
[WJ-986] Add DEEPWELL method to retrieve a list of page revisions

emmiegit:
This implements the method to retrieve a series of page revisions. You can request up to 100 revisions (default 10) before or after a given revision number on a page. Like with the page requests, it conditionally retrieves wikitext and HTML if you specify those query parameter flags.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #943, opened 2022-04-26
[WJ-985] Add page undeletion / restoration to DEEPWELL

emmiegit:
This PR makes a number of changes related to page and revision handling to facilitate the addition of page undeletion.

This is the opposite operation of page deletion, but is a bit more involved since it's a cross between page creation and page editing. It also lays some groundwork for what page moving will involve.

------

While working on these changes, I realized that some prior ideas and approaches I had developed should be adapted. So this PR does strictly a bit more than just implementing the undeletion method:

* Adds the `revision_type` column
  * This represents the previously-fuzzy concept of a revision with special properties. This also allows us to cleanly represent the page creation event as something separate, which was previously handled in an unclear way.
  * So now, there are four revision types: `create` (for the initial revision), `delete` / `undelete` (for page deletion and restoration), `regular` (for regular edits)
  * This is implemented via Postgres type / enum
  * We use SeaORM's generated enum as the Rust enum in code
* Removes the `metadata` column
  * The `metadata` column was never really well-defined, which is problematic, but wasn't a huge issue until I started to try and use it. Originally conceived as a fuzzy precursor to customizable metadata, but also used for storing page deletion / restoration, this proved to be a problem because it wasn't clear how we determined between deletion events (revisions that delete) and deletion state (revisions that note the page is or isn't deleted, but aren't the action itself necessarily.
* Finally cleans up / consolidates `RevisionTasks` logic
* Fixes implementation issues with both `create_tombstone()` (deletion) and `create_resurrection()` (undeletion)
  * For instance, the slug of a restored page was actually the revision comments, rip
  * Page category wasn't properly created / set if the slug changed
  * We need to rerender the page on undelete
* The `changes` column properly determines what has changed, not just what is specified in the web request
* Updates to revisions (hiding revision data) disallows hiding wikitext if it's the latest revision
  * This is because the page contents are visible anyways as part of the current page, so it's not truly hidden

I also am starting to feel the need for more cohesive integration tests for page operations ([WJ-978](https://scuttle.atlassian.net/browse/WJ-978)), which will require some more work. At least we could start the code for it and worry about the CI step later lol.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #946, opened 2022-04-26
[WJ-991] Add methods to get, set, and delete page parents to DEEPWELL

emmiegit:
Essentially just a CRUD API, except our page parents are many-to-many instead of one-to-many like in Wikidot.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #958, opened 2022-05-04
[WJ-1024] Add separate feature flag for the HTML renderer.

emmiegit:
This adds a new feature to the `ftml` crate called `html`, enabled by default, which covers `HtmlRender` and related code. This way, a potential user who does not want the HTML renderer (either because they only want the syntax tree or have their own custom renderer) do not need to bring in what is a fairly large segment of the code.

### DIFF
--- a/ftml/Cargo.lock
+++ b/ftml/Cargo.lock
@@ -282,7 +282,7 @@ dependencies = [
 
 [[package]]
 name = "ftml"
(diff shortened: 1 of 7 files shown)


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #976, opened 2022-05-12
[WJ-1022] Add ruby blocks

emmiegit:
The [`<ruby>` annotation elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ruby) are HTML elements that assist in labelling characters in CJK languages. Currently on Wikidot, this is done by using custom CSS classes: `[[span class="ruby"]]` and `[[span class="rt"]]`. This adds them as native HTML blocks, and allows adding other attributes to them as expected.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #978, opened 2022-05-14
[WJ-995] [WJ-62] [WJ-22] Page votes and scoring

emmiegit:
Adds the `page_vote` table to DEEPWELL, as well as methods to create, get, delete, and disable votes.

Also contains the beginning of the scoring infrastructure, that is, taking vote to calculate a score which can be used to order pages.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

web/database/migrations/2022_05_13_040630_deepwell_page_vote.php
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public function up()
+    {
+        // This hands ownership of the page vote table to DEEPWELL
+
+        // Drop old tables
+        Schema::drop('page_rate_vote');
+
+        // Create new table
+        DB::statement("
```
danieltharp, line 23: Only thing I wanna note is stuff in this style does lock the solution in to Postgres. I don't think that's problematic right now, but if down the road there was a determination to use a different database this would have to be refactored.

This could be done with the typical schema statements and a conditional `if (env("DB_CONNECTION") == "pgsql"`  to run the check constraint.

emmiegit: This is true, though this is also an issue with our use of arrays for tasks such as storing tags.

For this specifically, I've mostly been using the raw SQL since it's not being managed by Eloquent and won't be retrieved as an Eloquent model.

PR #979, opened 2022-05-14
Fix packing script for web modules (and update ftml-wasm)

Zokhoi:
Fixes the copy method erroring due to fs-extra having been removed.
Also updates ftml-wasm and its npm scripts to latest ftml features.


emmiegit (approved):
Thanks for this!

PR #981, opened 2022-05-18
Create pl.ftl

Manganian7potasu:


emmiegit:
For some reason the locale validator check didn't activate. Not sure why, but I checked the branch locally and the validator passes.

locales/fluent/base/pl.ftl
```diff
@@ -0,0 +1,149 @@
...
+reset-password = Resetuj hasło
+
+remember-me = Zapamiętaj mnie
+
+create-account = Utwórz konto
+
+field-required = To pole jest wymagane
+
+characters-left = { $count ->
+  [1] Pozostał 1 znak
```
emmiegit, line 106: Why is this `P` capitalized when the other one is lowecase? Is this a property of Polish I'm unaware of?

Manganian7potasu: In english file there's:
```
characters-left = { $count ->
  [1] 1 character left
```
It could be translated as "1 znak pozostał", but it's grammarly incorrect in polish. Sentence "Pozostał 1 znak" sounds better, because verb, not noun, is the first word. Directly translating it to english it would be "Left 1 character" but it's also grammarly incorrect. In polish it's good.

Manganian7potasu: Oh, you meant:
```
pozostało { $count } znaków
}
```
Yeah, `P` should be capitalized. It's a mistake, thanks for noticing that

emmiegit (approved):

PR #982, opened 2022-05-19
Change "rating" field to f64

emmiegit:
This increases the size to match what DEEPWELL is using for the `ScoreService`.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #984, opened 2022-05-20
[WJ-989] Add page rollbacks

emmiegit:
This adds a DEEPWELL method to rollback a page to a previous specified revision. It copies all fields from that revision (except slug) and creates a new revision with that data. Like with edits, it only creates a revision if there are actual changes from the revision it's resetting to. It's essentially equivalent to replaying the edit at the specified revision number.

Also update rust version data.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #985, opened 2022-05-20
[WJ-988] Add page moves

emmiegit:
Depends on #984 	

This adds a DEEPWELL method to move / rename a page. The `RevisionService::create()` method already handled slug moves, but this was never invoked in any code.

The new `PageService::rename()` method (since `move` is a Rust reserved word) makes a revision which only alters the `slug`, and ensures the `page` row in the database is updated appropriately.

I have also added a new `revision_type` of `move` to correspond with these revisions which only encompass slug changes (otherwise they are indistinguishable from `regular` revisions). This revision type is automatically applied to any revision which changes the slug, but because only the `PageService::rename()` method does so, and does not alter any other fields, it is effectively solely reserved for page moves.

### DIFF
--- a/deepwell/Cargo.lock
+++ b/deepwell/Cargo.lock
@@ -910,7 +910,7 @@ dependencies = [
 
 [[package]]
 name = "deepwell"
-version = "0.6.3"
+version = "0.7.0"
 dependencies = [
  "anyhow",
  "async-std",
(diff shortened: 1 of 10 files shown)


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #986, opened 2022-05-21
Create vi.ftl

tiggre2225:


emmiegit:
Ran validator locally, looks fine.

locales/fluent/base/vi.ftl
```diff
@@ -0,0 +1,149 @@
...
+remember-me = Giữ đăng nhập
+
+create-account = Tạo Tài Khoản
+
+field-required = Mục này là bắt buộc
+
+characters-left = { $count ->
+  [1] 1 ký tự còn lại
+  *[other] { $count } ký tự còn lại
+}
```
emmiegit: It seems like Vietnamese doesn't have a separate form for plurals. So you could just do this instead, right?
```
characters-left = { $count } ký tự còn lại
```

tiggre2225: you mean the plural form like "s" in english?

emmiegit: Yes. The reason there is the `{ $count -> ... }` for English is because the translation is different if it's 1 vs many. It looks like "ký tự còn lại" is the same for either case, so the switch is not necessary.

tiggre2225: Vietnamese nouns only have 1 form for singular and plural meanings, so it stays the same

emmiegit: Okay cool. Then you can change the translation line to `characters-left = { $count } ký tự còn lại` instead to reflect the 1 form for both meanings.

tiggre2225: I just changed and saved it, thanks for telling me about that

emmiegit (approved):

PR #994, opened 2022-05-23
Decrease dependabot updating to once a week

emmiegit:
We've had it at daily for a while, and most updates are not urgent enough to require the PR spam that it creates. I think once a week is more than sufficient to balance frequency with notification generation.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #1000, opened 2022-05-23
Modify workflow activations to work properly on external PRs

emmiegit:
Apparently we've been doing this wrong. The [GitHub workflow documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore) has a separate activation for PRs. The `push` workflow type applies only when a commit is added to this repository, so it will not activate for external PRs, explaining our issue.

The `branches-ignore` for `pull_request` is different than `push`. We remove it here because otherwise it doesn't run on PRs _targeting_ that branch, which is the main point of most PRs.

This change also makes it so pushes to `develop` and `prod` run all the standard checks (e.g. `ftml.yaml`) on every merge. May as well do this, it both avoids duplicating the path globs, and ensures nothing incidentally breaks through some unexpected side channel.

### DIFF
--- a/.github/workflows/client.yaml
+++ b/.github/workflows/client.yaml
@@ -7,7 +7,6 @@ on:
       - "ftml/test/*"
       - ".github/workflows/client.yaml"
       - ".github/codecov.yml"
-
   push:
     branches:
       - develop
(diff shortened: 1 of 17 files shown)


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

rossjrw (approved):

PR #1002, opened 2022-05-23
Add include compatibility mode setting

emmiegit:
This adds the wikitext setting `use_include_compatibility`, which enables `[[include]]` as an alias for `[[include-messy]]`.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #1006, opened 2022-05-23
[chore] Fix locales build

emmiegit:
This PR:

* Clarifies how fluent functions are validated. (Constant array that is checked against)
* Renames the constant that's referenced.
* Don't skip over the primary locale during checking.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #1033, opened 2022-06-19
[WJ-1001] Add preliminary S3 and file support

emmiegit:
This PR adds S3 as a backing store that can be used, which is utilized via the new `BlobService`. This is a content-addressable blob store, similar to `TextService`. This is utilized by the new `FileService`, which stores individual file entries. Because they point to hashes for content, they automatically deduplicate, even if there are multiple files with the same content.

To ensure that file changes are not lost, like in Wikidot's simple CRUD system, the `FileRevisionService` stores rows in the `file_revision` table. Like `page_revision`, this represents the history for a particular file, and in addition to updating name/metadata, deletion, restoration, it also allows it to be moved between pages.

This is a preliminary PR, and some items are not settled yet. For instance, the exact method handlers for files is not yet specified, and will require some discussion. (Consider, web requests use the body for JSON input fields, but we also want a separate content type and blob body data, or a way of doing multipart uploads). I am finishing here because I implement the main skeleton as I intended, stubbing out relevant bits, and we are shifting towards the prioritized development plan that I highlighted in chat.


emmiegit:
I need to install libmagic in the containers, which is why builds are currently failing

Yossipossi1 (approved):

Monkatraz (approved):

emmiegit:
Finally, the builds passed lol

thanks @Yossipossi1 @Monkatraz

PR #1048, opened 2022-06-20
[WJ-1002] Add password and TOTP support 

emmiegit:
This adds the cryptographic basics for password verification (hashing via argon2), multi-factor authentication (MFA) using temporary one-time passwords (TOTP), and basic methods for setting each of those values. We will eventually need to encode the MFA secret in a QR code but that's out of scope at the moment.

The addition of cryptographic operations here provides a foundation for later authentication and related changes that will come as the account system develops.

The new services added are:
* `PasswordService` -- wraps the `argon2` crate by providing password hashing methods
* `MfaService` -- validates TOTP and recovery codes to supplement password auth, if enabled
* `AuthenticationService` -- calls the above services to perform high-level authentication operations

I have tried to be careful with my implementations, ensuring that I source random values from a CSPRNG, that any comparison operations are constant-time to avoid timing attacks, that failed authentications result in a constant time sleep to avoid brute-forcing, and that passwords are appropriately hashed before being stored in the database. (Relevant note there, the TOTP secret cannot be encrypted, as its exact value is needed on the server end. I did some research on this before and it seems the typical approach is to just store it as-is, like the password's salt.)


emmiegit:
Haven't gotten a response from https://github.com/TimDumol/rust-otp/issues/4. If enough time passes without interaction from the maintainer, I'm wondering if we should fork it and maintain it ourselves.

Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1069, opened 2022-07-05
[WJ-1037] [WJ-1038] Create generic content seeder framework

emmiegit:
This produces a simple Python script framework to output a SQL seed file based on a backup source, such as SCUTTLE. This PR also includes a preliminary implementation for Wikicomma (which is not complete because it lacks some necessary information).


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1071, opened 2022-07-10
[WJ-971] Add orange links 

emmiegit:
This adds the class `wj-link-missing` to any triple-bracket links which have missing pages. This would correspond to the `newpage` class in Wikidot which is used to make links "orange" (corresponding page doesn't exist yet).


Monkatraz (approved):

emmiegit:
Thanks @Monkatraz

PR #1072, opened 2022-07-10
[WJ-814] Run through all ftml tests

emmiegit:
Previously the AST test runner would quit as soon as it hit its first failure. This change now always runs through all the tests, so multiple runs are not required to debug test failures.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #1081, opened 2022-07-11
Add all the Korean translations (fluent)

Denevola:
Translated all files under the `locales/fluent` directory into Korean.


emmiegit (approved):
Everything seems to look fine to me, thanks for the contribution!

Denevola:
@emmiegit I have a question about contributing.

Is it okay to translate `locales/cmftml` as well? I wasn't sure since there are no translations yet and the locales validator doesn't support it.

emmiegit:
@Denevola You could I suppose, but I think we should wait until those are more set in stone first

Denevola:
I understood, thanks.

PR #1132, opened 2022-08-11
[WJ-1043] Use minio local container for S3

emmiegit:
This adds [minio](https://min.io/), a portable S3-compatible storage service to the local deployment. This is used for file storage in local environments, avoiding the mess of setting up AWS so that you can use S3 there.

Presently the data is _not_ in a Docker volume, but one can be added easily (add to `files:/data`). This means that file data is persisted along with database data, in the container itself, and can be easily wiped/reset as development efforts require.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1150, opened 2022-08-18
[WJ-1048] Prefix with # for hex colors, if missing

emmiegit:
@jewalky reported an issue with ftml, showing that hex colors without the `#` are not prefixed. That is, the following works:
```
###ccc|text here##
```
But without the explicit `#` it doesn't work:
```
##ccc|text here##
```
Since it produces `color: ccc`, which is not valid. This code changes it so that if a hex color is used, a `#` is prefixed it it does not already exist.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #1151, opened 2022-08-18
Replace leading non-standard spaces with regular spaces

emmiegit:
To resolve some issues with the use of non-breaking spaces instead of regular spaces, we replace those leading on a line with regular ones. This solves issues with paragraphs or lists which look for regular space tokens as part of parsing.


Yossipossi1 (approved):

Monkatraz (approved):

emmiegit:
thanks @Yossipossi1 @Monkatraz

PR #1152, opened 2022-08-18
[WJ-1047] Add CSS in <body>

emmiegit:
This is a pretty substantial change. The core of it is that inserted CSS now appears as `<style>` blocks within `<body>` rather than all being prepended in the `<head>`. However this entailed a lot of other changes to how the library works.

Because these styles had to be "bubbled up" to the top level rather than kept at the contextual element level, the `ParsingException` structure had its style variant removed, and the respective logic that stored styles as separate from elements were also changed. I also took this opportunity to rename `ParsingWarning` -> `ParsingError`, which is what it really is (even if "non-fatal" in the typical sense).

Now `ParsingException` was now the same as `ParsingError`, so I removed that wrapper structure everywhere and interfaced with a list of errors directly. Then with that settled, I updated all the AST tests, as they now insert the HTML styles into the output, have them as elements, and the `warnings` field was obviously renamed.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #1153, opened 2022-08-18
Fix issue with +* token

emmiegit:
The `heading` token looks for 1-6 `+` and an optional `*`, reflecting how headers work in Wikidot. For instance:
```
++ Sub-header
+++* Another header (but no table of contents)
```

However this causes a problem with this specific string: `**18+**`. It lexes it as `**` `18` `+*` `*`, which is incorrect, and causes it to not bold the relevant text.

This PR changes the lexer rule for `heading` so that it requires a non-`*` after in the `*` case. I also add an AST test to catch regressions.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #1154, opened 2022-08-19
[WJ-1051] Filter out URL contexts

emmiegit:
This ensures that the `href` and `src` attributes cannot output `javascript:` or `data:` values, which could lead to security issues.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #1155, opened 2022-08-19
[WJ-1055] Replace CUID (string) IDs for file with regular BIGSERIAL

emmiegit:
This changes the ID type for `file` to be a regular integer rather than the [`cuid`](https://github.com/ericelliott/cuid)-based string IDs.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1156, opened 2022-08-20
[WJ-1035] Sanitize (and minify) CSS

emmiegit:
Previously we would add the CSS, with HTML escaping. This caused issues in some cases because HTML escaping is not correct for CSS (though it does prevent some security issues). In order to prevent problems with invalid/problematic CSS, this PR uses [`parcel_css`](https://docs.rs/parcel_css/1.0.0-alpha.30/parcel_css/stylesheet/struct.StyleSheet.html) (based on [Firefox's CSS parser Rust library](https://docs.rs/cssparser/latest/cssparser/)) to parse out the CSS, then re-emit it.

Testing this locally, this prunes any invalid CSS which could be a security problem. And because we get it for free, I also minify the CSS. (This is toggleable based on a `WikitextSettings` flag).

### DIFF
--- a/ftml/Cargo.lock
+++ b/ftml/Cargo.lock
@@ -9,6 +9,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
 
 [[package]]
+name = "ahash"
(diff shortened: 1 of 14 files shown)


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #1178, opened 2022-09-02
[WJ-1074] Remove [[css]] alias

emmiegit:
It was a dumb thing to add in the first place, and since it doesn't already exist in Wikidot we can safely remove it without any issue.


Yossipossi1 (approved):

emmiegit:
thanks @yossipossi1

PR #1179, opened 2022-09-03
[WJ-1063] Swap inner() and contents()

emmiegit:
As discussed on the issue, in the ftml HTML builder, `inner()` is a better fit for what `contents()` does and vice eversa, so this PR swaps them.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #1181, opened 2022-09-05
[WJ-822] Add the bibliography

emmiegit:
This implements Wikidot's bibliography syntax:

* `((bibcite ...))`, as in Wikidot
* `[[bibcite ...]]`. since the `((` syntax for no reason is ridiculous. In this version, it also adds the frequently desired brackets around the number (e.g. `[1]`); this can be disabled by using `[[bibcite_ ...]]`.
* `[[bibliography]]`, displaying a set of references. At the recommendation of @jewalky, ftml supports multiple bibliography blocks, allowing separation in a way not currently possible with footnotes. If multiple labels are specified, then the first one takes precedence; I took pains to ensure this is true for the various forms it can occur in.
* For HTML rendering, I just copied our footnote implementation for now. [WJ-845](https://scuttle.atlassian.net/browse/WJ-845) covers the interactivity portion.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1193, opened 2022-09-07
Locale Validation (& fixing warnings)

Yossipossi1:
This PR adds a locale validation helper function and implements it for the SiteService. This also fixes a few warnings from the last PR by adding SiteService methods to the API proper.


deepwell/src/web/locale.rs
```diff
@@ -0,0 +1,29 @@
+/*
+ * web/file_details.rs
```
emmiegit:
```suggestion
 * web/locale.rs
```

deepwell/src/methods/mod.rs
```diff
@@ -54,7 +54,7 @@ pub mod category;
 pub mod file;
 pub mod file_revision;
 pub mod link;
-pub mod locales;
+pub mod locale;
```
emmiegit, line 57: Need to rename file in header comment.

emmiegit (approved):

PR #1194, opened 2022-09-07
[WJ-1084] Timeout ftml calls that take too long

emmiegit:
As a safety precaution (in case of a bug in the parser, some kind of malicious input, or just a very unusual processing delay), if rendering takes too long, then it should be prematurely terminated.

This PR adds this capability, returning an error if `RenderService::render()` takes longer than the configured amount of time (default 2000ms).


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1195, opened 2022-09-07
[WJ-1081] Allow invalid escapes

emmiegit:
In the interest of being more permissive with inputs (like Wikidot), strings do not require that only valid escapes are given, and any will be repeated as-is. So for instance `"a \n \z"` will return `a <newline> \z`. This also loosens up other checks which previously assumed ASCII boundaries (depending on the lexer to validate for it).


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1196, opened 2022-09-07
[WJ-1073] Simplify and clean up text renderer

emmiegit:
Previously the text renderer kind of pursued a "what if the page rendered to a terminal" psuedo-approach. However, because we actually use it internally (like in headings), it makes no sense for those sorts of elements to appear. See [WJ-1073](https://scuttle.atlassian.net/browse/WJ-1073), the issue which raised this.

So I've went ahead and culled `TextRender` of all the weird complexity and made it very simple: if it can clearly be copied as text, it is, with no formatting. If not, it's ignored.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1223, opened 2022-09-22
[WJ-1077] Framerail repository reorganization

emmiegit:
As part of [WJ-1077](https://scuttle.atlassian.net/browse/WJ-1077) and the [Framerail architecture plan](https://scuttle.atlassian.net/wiki/spaces/WD/pages/1570832385/Framerail-based+Architecture+Plan), this large PR reorganized our repository to prepare for this project change.

The following changes (among others) have been made:
* Move `web/` -> `legacy/`. As noted, the idea is to use it as a read-only reference, while slowly removing files from it as they're reimplemented.
  * Some files are retained in `web/`. These are primarily build scripts, javascript modules, and other frontend utilities which we need to move to `framerail/`. The `web/` directory will eventually be deleted.
* Update/remove devops flows associated with PHP.
* Move migrations from Laravel Eloquent to DEEPWELL. [[WJ-1025](https://scuttle.atlassian.net/browse/WJ-1026)]
  * Note that until we have a production deploy, we will be maintaining "clean migrations". That is, a single set of migrations to seed the database being amended with changes, with iterative new migrations when we actually have a database we care about.
  * Legacy Wikidot tables (prefixed with `wikidot__`) are not being retained. They appear in the migrations to ensure they work, but are immediately dropped after. Because we are no longer running the legacy PHP, there is no need for the old tables; when we want to (re)implement functionality, we will go through the schema revision procedure to create suitable table(s) for our needs.
* Set up new seeder system in DEEPWELL, based on JSON and text files.
* Some changes to the `site` table.
* Modified `docker-compose.yaml` to allow for local (web-less) deploys. This will seed the database and run the API service.
* Allow disabling logins for a user. [Stub]
* Provisional edits to the README.
* Edits to CODEOWNERS.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1224, opened 2022-09-23
[WJ-1097] Minor ftml buffer optimization

emmiegit:
Minor optimization: strings in Rust start allocation at zero (no allocation), and after they are first appended to they begin growing, doubling capacity each time. However since we know that the input wikitext length is correlated with the output HTML length, we can avoid excessive re-allocation (especially early on when the value is small) by basing our initial buffer size based on the input length in bytes.

I ran a script to look at our `test/` files, and I saw that, typically, the output HTML size stays below ~12% of the input wikitext byte length. (Though a value of 10% would likely also be fine.) So, during `HtmlContext` initialization, I use `String::with_capacity` with 112% of the input string's size, which should make allocations at all during the rendering process uncommon.

The `TextContext` initialization is similar, but it matches the input size (i.e. 100% of the value).

In various edge cases (and when retrieving from test files) this value is set to 0, since there is no change in correctness, just a (slight) decrease in speed.

### DIFF
--- a/ftml/src/render/text/mod.rs
+++ b/ftml/src/render/text/mod.rs
@@ -38,6 +38,7 @@ impl TextRender {
         elements: &[Element],
         page_info: &PageInfo,
         settings: &WikitextSettings,
+        wikitext_len: usize,
     ) -> String {
         self.render_partial_direct(
             elements,
@@ -46,6 +47,7 @@ impl TextRender {
             &[],
             &[],
             &BibliographyList::new(),
+            wikitext_len,
         )
     }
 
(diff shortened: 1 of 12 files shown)


ftml/src/render/text/mod.rs
```diff
@@ -101,13 +107,29 @@ impl Render for TextRender {
...
+#[derive(Debug)]
+struct RenderPartial<'a> {
+    elements: &'a [Element<'a>],
+    page_info: &'a PageInfo<'a>,
+    settings: &'a WikitextSettings,
+    table_of_contents: &'a [Element<'a>],
+    footnotes: &'a [Vec<Element<'a>>],
+    bibliographies: &'a BibliographyList<'a>,
+    wikitext_len: usize,
+}
```
emmiegit, line 135: Adding one extra paramater to this function means clippy complains about too many, so I created a wrapper structure for the `render_partial_direct()` call.

Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1232, opened 2022-09-23
[WJ-1076] Clean up some legacy files

emmiegit:
Removing files in `legacy/` which we aren't likely to use or reference in the near future.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1245, opened 2022-10-06
[WJ-962] User table refactoring

emmiegit:
This took a while, sorry.

This PR refactors the user table from Laravel into a new version fitting the DEEPWELL pattern, and adds new associated user handling infrastructure. The majority of the work here was for user aliases and user bot information. We allow a many-to-one relationship between bot owners and bots, so this is implemented as a table, rather than hardcoding the notion that there's a, say, `owner_user_id` in the `user` table for a bot. Aliases are also implemented, according to [WJ-1080](https://scuttle.atlassian.net/browse/WJ-1080).

If you are using postgres directly, note that you need to quote the user table (i.e. `"user"`, like `SELECT * FROM "user"`) because it's a special table in postgres.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1252, opened 2022-10-10
Update dependabot directory for npm

emmiegit:
`/web` is being moved over to `/framerail`, as discussed, and does not need dependency upgrades (like `/legacy`).


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #1263, opened 2022-10-26
[WJ-1056] Upgrade sea-orm to 0.10

emmiegit:
A number of build issues were preventing us from upgrading earlier. This should fix these.

PR #1264, opened 2022-10-26
[WJ-987] Use postgres arrays instead of JSON

emmiegit:
~~Blocked on #1263~~

Until now, we needed to use the `JSON` type to store lists of text, such as tags, within Postgres because [Sea-ORM](https://github.com/SeaQL/sea-orm) did not support However, as of [version 0.10, Sea-ORM supports PostgreSQL arrays](https://github.com/SeaQL/sea-orm/releases/tag/0.10.0) (under a crate feature). I have enabled this and changed over our tables to use `TEXT[]` instead of `JSON`, which allows us to eliminate the temporary JSON/array conversion functions that currently exist in our code.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1265, opened 2022-10-26
Address new clippy lints in ftml causing develop build breakages

emmiegit:
Same new lints as in deepwell.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #1266, opened 2022-10-26
Hotfix: Old SQL function in migration

emmiegit:
Some of the invocations were left behind and needed to be removed. A local deployment now builds and can run, so the issue is fixed here.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #1283, opened 2022-11-25
[WJ-1078] SvelteKit configuration changes

Monkatraz:
Changes:
- uses PNPM
- switched to an ESLint configuration similar to our original one (much better linting for Svelte files in particular)
- switched to a Prettier configuration similar to our original one
- added back Stylelint, pretty much exactly the same as our original one
- Ran linter and formatter across the PR


emmiegit (approved):
Thanks for doing this!

PR #1285, opened 2022-11-26
Upgrade chrono

emmiegit:
Supersedes #1273, #1274

This version deprecates some methods, so I change assertions to either be more explicit, add checking, or avoid the methods altogether.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #1292, opened 2022-11-29
[WJ-1089] [WJ-975] [WJ-1091] [WJ-1116] [WJ-185] Initial implementation of filters

emmiegit:
This PR implements the concept of "filters". which are configurable regular expressions which permit automated blocking of certain kinds of content. This is the Wikijump version of Wikidot's `ForbiddenNames` class.

Filters are divided into two kinds: platform and site. A platform filter applies to the entire platform, regardless of individual site policies. A site filter applies to only that particular site. (An enum variant of `FilterClass` exists to easily query both the platform and specific site filters, since both apply when considering a particular site's enforcement policies). The database allows creation of filters with a pattern and a "description" field. Which objects it applies to (e.g. pages, forum posts, etc.) are set via flag toggles. This way a common filter (say for a racial slur) does not need to be duplicated for each type.

Filters use `RegexSet` under the hood to check all the relevant filters at once. This way users can add any number of filters they need, and do _not_ need to perform optimizations such as grouping unrelated filters together with `(a...|b...)` in regex.

User filters work slightly differently. Platform user filters apply when the user is registering an account, and prevent it from being made if a violation is found. A site user filter prevents an existing user from joining a site.

All extant operations (users, pages, files) have code to use the new `FilterService` to check for violations, as appropriate. This applies to both creation and edits, so a user cannot circumvent a filter by creating the object with an acceptable value and then editing it to an inappropriate one.

Per [WJ-1116](https://scuttle.atlassian.net/browse/WJ-1116), the caller can suppress filter validation. In some contexts (staff on a site), the user can explicitly opt to override a filter after being told their normal operation violates it.

This PR also implements [WJ-185](https://scuttle.atlassian.net/browse/WJ-185) which performs name normalization to users on creation to remove leading or trailing whitespace.

It also adds `PageOrder` to represent the ways that requesting pages from `PageService` can come, since previously it came out unordered (probably actually last insertion/update order).


deepwell/src/models/sea_orm_active_enums.rs
```diff
@@ -23,7 +39,7 @@ pub enum PageRevisionType {
     Debug, Copy, Clone, PartialEq, Eq, EnumIter, DeriveActiveEnum, Serialize, Deserialize,
 )]
 #[sea_orm(rs_type = "String", db_type = "Enum", enum_name = "user_type")]
-#[serde(rename_all = "camelCase")]
+#[serde(rename_all = "kebab-case")]
```
Yossipossi1, line 42: Why kebab-case? And how does it affect this struct any differently from camelCase, considering all the enums are one word?

emmiegit: This is one of the things the autogeneration tool removes and I need to re-add. I guess it was originally camelCase though in practice all the enum values are intended to be kebab-case (but this hasn't come up since they're all one word).

If you're asking why this is necessary at all, the issue is otherwise it'll be stored as `Foo` and `Bar` not `foo` and `bar`.

Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1294, opened 2022-11-29
[WJ-1115] Pivot DEEPWELL towards being an internal API

emmiegit:
Prior to this, DEEPWELL was mostly designed to be internal but had some features (such as a rate-limit) which reflected a lack of certainty about whether it should in the end be a public or internal API. Since we are approaching this as an internal API in a VPC to be consumed only by the framerail web server.

This makes some changes in that vein, restructuring the project a bit.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1296, opened 2022-11-29
[WJ-1113] Add email filter

emmiegit:
This adds a new email filter type, which allows regular expressions to deny emails from known bad hosts. It operates on the same principle as other filters, but applies to user's registration emails.

### DIFF
--- a/deepwell/seeder/filters.json
+++ b/deepwell/seeder/filters.json
@@ -181,6 +181,18 @@
     },
     {
         "site": null,
+        "regex": "^sudo$",
+        "description": "Reserved name (sudo)",
+        "user": true
+    },
+    {
+        "site": null,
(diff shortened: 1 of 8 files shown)


deepwell/seeder/filters.json
```diff
@@ -161,6 +179,18 @@
...
+        "regex": "^sudo$",
+        "description": "Reserved name (sudo)",
+        "user": true
+    },
+    {
+        "site": null,
+        "regex": "^wheel$",
+        "description": "Reserved name (wheel)",
+        "user": true
+    },
```
emmiegit, line 193: Forgot these in the first filter PR.

Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1297, opened 2022-11-29
[WJ-1111] Add ScoreValue enum to ftml

emmiegit:
This needs to be a separate PR since DEEPWELL needs to depend on a new structure which presently does not exist in the target crate.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #1298, opened 2022-11-29
[WJ-1111] Uses ScoreValue in DEEPWELL

emmiegit:
~~Requires #1297~~

Uses the new `ScoreValue` structure in DEEPWELL, after it was added and published in the ftml crate.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #1299, opened 2022-12-01
[WJ-1121] [WJ-364] Sessions

emmiegit:
This adds the `session` table to the database and implements `SessionService`, with associated new routes for logging in and initiating sessions. Throughout I reviewed [OWASP Session Management Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html). Particular care is paid to the main login endpoint, which runs `AuthenticateService::run_password()`, which is modified to prevent timing attacks in the case of an unknown/invalid user. (Meaning it pulls up a fake user, runs calculations on that, and discards the output, so an attacker cannot tell from the time delay if the user exists or not).

There are a few services of note here, so I'll summarize them:
* `PasswordService` -- Verifies if a password matches a password hash, and can create new password hashes.
* `MfaService` -- Verifies TOTP (timed one-time passwords) and recovery codes, and can create new recovery codes.
* `AuthService` -- Invokes the prior two to authenticate a particular user. There are separate calls for the password and MFA stage. If a user has MFA, then they are given only a temporary restricted session in the mean time.
* `SessionService` -- Creates, invalidates, and expires sessions. These have long, securely-generated session tokens as the primary key, and are associated with the user ID and the login origin information (IP address and user-agent).

The new `SessionService` has several methods, not just for the creation and deletion of sessions, but also for listing all sessions (to see current login activity), to invalidate all _other_ sessions (see [WJ-364](https://scuttle.atlassian.net/browse/WJ-364)), and getting the associated user from a session. Bear in mind that the `restricted` flag is important: a restricted session should _not_ be treated as if they have no access. It is used for temporary states, such as logging in.

In general, sessions expire quickly (30 minutes) and are expected to be frequently exchanged, a process which should be invisible to the end client. A `renew()` method is provided for this purpose.

There is a task to remove expired sessions from the database regularly, however even if they built up it will not pose a security risk, as expired sessions are excluded via query condition.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1308, opened 2022-12-11
Deny [[iframe]]s which specify a non-URL

emmiegit:
As a means of XSS prevention, using the `is_url()` helper function. Bumps version so we can ensure dependents have the fix.


Monkatraz (approved):

emmiegit:
thanks @Monkatraz

PR #1315, opened 2022-12-18
Remove git sources from [[embed]]

emmiegit:
Due to potential security concerns over GitHub and GitLab snippets and XSS, I'm removing these features from ftml. Since there are no users of these features yet, there should be no incompatibility problems. This includes a version bump.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1389, opened 2023-03-13
[WJ-1136] Add "unknown user"

emmiegit:
Instead of using "anonymous" for this purpose, per [WJ-1136](https://scuttle.atlassian.net/browse/WJ-1136) this adds an "unknown user" sentinel which specifically represents users whose true identity are not known or represented, rather than being intentionally concealed.

Thankfully because of our new seeder infrastructure, making this change was very simple.

I also noticed that the `location` field had been inadvertantly excluded. I have added it, along with starter seed values, and the relevant code changes to also insert that data.

[WJ-1136]: https://scuttle.atlassian.net/browse/WJ-1136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deepwell/seeder/users.json
+++ b/deepwell/seeder/users.json
@@ -10,6 +10,7 @@
         "real_name": "Administrator",
         "gender": null,
         "birthday": null,
+        "location": null,
         "biography": "Root platform administrator for this Wikijump instance",
         "user_page": null,
         "aliases": [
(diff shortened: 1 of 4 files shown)


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1405, opened 2023-03-30
[WJ-1135] Add wikidot origin database column, create ImportService

emmiegit:
This PR adds `from_wikidot` as a column to the database, which is set in the new `ImportService`, which is meant to be used during one-time Wikidot data ingestion.

This also updates the Sea-ORM models because of the new update. Email uniqueness checks have been changed, it's only enforced against regular users and is not constrained in the database (allowing deleted users to "reuse" emails).


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1406, opened 2023-03-31
[WJ-1142] Web setup for local instances

emmiegit:
Here I've set up our `docker-compose` build system to serve the SvelteKit web server and permits visiting a local site representing your instance.

This involved adding two new containers, `web` (node web server), and `nginx` (reverse proxy). The goal is for nginx to serve dynamic certificates (later on), and for now, do basic route-untangling and redirection, and then the local SvelteKit server will receive more well-formed paths. The nginx instance also redirects HTTP -> HTTPS and `www.` -> root. (Though internally the "main" wiki will still be that with slug `www`.)

For local, the docker image creates a snake oil / self-signed certificate, like we had previously. For dev and prod this will need to handle custom domains via Let's Encrypt certificate creation and distribution.

`install/files/nginx/ffdhe2048.txt` is just the contents of https://ssl-config.mozilla.org/ffdhe2048.txt.

Screenshot of `https://wikijump.localhost/`:
image


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1414, opened 2023-04-07
[WJ-1146] Svelte routing, add first connection to DEEPWELL

emmiegit:
This is a proof of concept for framerail, having it successfully make a call to DEEPWELL (in this case, ping) and send the response in the generated HTML:

image


framerail/src/lib/deepwell.server.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export async function ping(): string {
+  // TODO for now, dump as string
+  const response = await axios.put(
+    `http://${DEEPWELL_HOST}:${DEEPWELL_PORT}/api/trusted/ping`,
+     { timeout: 500 },
+  );
+
+  return response.data;
+}
```
Monkatraz, line 14: Axios isn't really needed depending on the complexity of the API, simpler more verbose `fetch` calls could be used. The old web used an OpenAPI JS generator to handle the grunt work here, which you could use. Not important for this PR really.

Monkatraz (approved):
lfg

framerail/src/lib/deepwell.server.ts
emmiegit: Yeah I figured. I'm not as experienced with the JS ecosystem, and I have no special attachment here, I'm totally fine changing this to be better. Besides this is hardly a full API fetch framework here that we'd lose.

emmiegit:
thanks @monkatraz

PR #1415, opened 2023-04-09
[WJ-1079] DEEPWELL internal API structure

emmiegit:
This PR is pretty big, and improves, rejuvenates, and fixes several aspects of DEEPWELL.

The primary  change is changing the DEEPWELL API (meant for trusted consumers, i.e. framerail) to better fit its role as an internal API. It also takes advantage of a known, singular API consumer context to simplify some requests. Particularly, it removes most route parameters and instead uses request bodies to store all the information. This is useful since, instead of there needing to be glue logic to attach route parameters to items from JSON bodies, we can have the entire request data be in one JSON object and not have to do much massaging in the `methods/` implementations beyond that.

This also makes a number of other miscellaneous changes, such as finally renaming `revision` -> `page_revision`, removing `CuidReference` (which had already been somewhat altered), adds `get_id()` methods for convenience `Reference` extraction, and others.

I ran through all the current methods to make sure they basically functioned as expected. Taking this opportunity, I fixed several things that had been developed but not quite working, such as user recreation (i.e. deleting a user and making one with the same slug) and uploading files (specifically uploading a user avatar blob).


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1416, opened 2023-04-10
[WJ-1148] Add deepwell.toml configuration file

emmiegit:
See [WJ-1148](https://scuttle.atlassian.net/browse/WJ-1148) for the issue description. This PR implements this by creating a `Secrets` structure, which is read only from environment variable, and a `Config` structure, which comes from a TOML configuration file (with some overridable via command-line flag). This TOML file is where configurable fields will go in the future, and I've moved out existing constants from the codebase into this file so that they can be configurable without needing to recompile.

I've created `/deepwell/config.example.toml`, which illustrates all the recommended values and provides documentation for each configuration key:

<details>
<summary>Sample documented deepwell.toml configuration file</summary>

```toml
[logger]

# Whether the logger is overall enabled or not.
# Passing "-q" completely silences all logging.
enable = true

# The log level to emit at.
#
# Possible values:
# - "error"
# - "warning"
# - "info"
# - "debug"
# - "trace"
level = "info"


[server]

# The IP and port to bind to when the server starts.
address = "[::]:2747"

# The path to write the pid file
# If excluded or empty, then no pid file is written.
pid-file = ""


[database]

# Whether to run migrations on startup (if needed).
#
# If false, then an out-of-date database will yield errors,
# you must migrate manually.
run-migrations = true

# Whether to run the seeder on startup (if needed).
#
# If false, then an empty database may yield errors,
# you must seed manually.
run-seeder = true

# The path containing data files for the seeder.
# Corresponds to /deepwell/seeder in the repository.
seeder-path = "seeder"


[security]

# The time to sleep in between failed authentication requests.
#
# If a request successfully authenticates then it is responded to
# without added delay. But if the request fails to authenticate,
# then this amount of delay (in milliseconds) is slept before
# returning the "authentication failed" response.
#
# This helps prevent brute-forcing by making repeatedly guessing invalid
# passwords slow, as opposed to legitimate users who only occasionally
# mistype their password and for whom this delay is not a significant issue.
#
# See https://unix.stackexchange.com/a/2127
authentication-fail-delay-ms = 100

[security.session]

# All session tokens are prefixed with this string.
# It makes a session token immediately reecognizable
# as it is otherwise a long random string.
token-prefix = "wj:"

# How long session tokens should be.
#
# Tokens are generated using rand::distributions::Alphanumeric,
# meaning they are cryptographically secure strings composed of
# ASCII characters A-Z, a-z, and 0-9.
#
# This starts counting at random bytes.
# No matter how long your prefix is (see above),
# the random portion of the session token will be this long.
# This means the *actual* length of a token is length(token_prefix) + token_length.
token-length = 64

# How long, in minutes, a user session token lasts.
#
# This is for the internal session token, which is automatically refreshed
# by framerail when it sees the token is close to expiry.
duration-session-minutes = 30

# How long, in minutes, a user has to enter their MFA code after initial login.
#
# When logging in with MFA, the user is given a "restricted session token".
# This session token is only useful for logging in with MFA (at which point they
# are given a regular session token), and is useless for anything else.
#
# This field determines how long such session tokens should last before expiry.
duration-login-minutes = 5

[security.mfa]

# The number of recovery codes to have available at any given time.
#
# After MFA setup, this number of recovery codes are generated for the user.
# If they ever use a code, it is exhausted (single-use), and on refresh any
# remaining recovery codes are invalidated and this number of new recovery
# codes are generated.
recovery-code-count = 12

# The length in bytes for each recovery code.
#
# It is suggested to make this value even, as a dash is inserted in the
# middle of each recovery code to make it easier to read.
recovery-code-length = 8

# The amount of time that each TOTP should last.
#
# We use 30 seconds because this is standard with helpers
# such as Google Authenticator and Authy.
#
# It balances between giving the user enough time to enter a code,
# but short enough to make bruteforcing values impractical.
time-step = 30

# The allowed leniency value to account for clock skew.
#
# This represents the number of seconds (positive or negative)
# that a TOTP is offset by in determining whether the authentication
# was accepted.
#
# See https://github.com/TimDumol/rust-otp/blob/master/src/lib.rs#L56
time-skew = 1


[job]

# How long, in milliseconds, to sleep in between jobs.
#
# The intent is to avoid overloading the database by allowing a
# short sleep period in between background tasks.
#
# See the JobService (src/service/job/service.rs) for information
# on the job queue and main loop.
delay-ms = 5

# The period, in seconds, to prune all expired sessions.
#
# This is *not* needed to actually invalidate sessions, so
# setting this to a long period is not a security risk.
# It is merely for clearing the database of already-expired tokens.
prune-session-secs = 600  # 5 minutes


[locale]

# The path containing Fluent translation files.
# Corresponds to /locales in the repository.
path = "../locales"


[ftml]

# How long (in milliseconds) to allow a render job to run before terminating it.
#
# This is to ensure that a parser bug or malicious input cannot
# crash or freeze the backend. This value should not be too
# aggressive, but still not extremely long.
render-timeout-ms = 2000


[user]

# The number of name changes a user has by default.
default-name-changes = 2

# The maximum number of name changes a user has.
#
# Attempting to add additional name change tokens to
# a user will top out at this amount.
max-name-changes = 3

# Every this many days, all users get another
# name change token (up to the cap). Checked hourly.
#
# Set to 0 to disable.
refill-name-change-days = 90
```

</details>

I was considering naming this `wikijump.toml`, like the old `wikijump.ini`, but I thought it would be better to have it consistent with the service name to avoid confusion.

[WJ-1148]: https://scuttle.atlassian.net/browse/WJ-1148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1419, opened 2023-04-11
[WJ-1150] Add "test" site in seeder

emmiegit:
The idea of this site is simply that it's for whatever pages, edits, etc. that don't affect any "actual" sites.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1421, opened 2023-04-11
Remove user details from internal API 

emmiegit:
This resolves a TODO, and simplifies the user method by simply passing the full model. The public framerail API can implement this filtration logic.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1425, opened 2023-04-12
[WJ-1152] Support site aliases

emmiegit:
This adds DEEPWELL support for site aliases. Meaning, a site slug can "point to" its actual destination, for instance `scpwiki.wikijump.com` transparently redirecting to `scp-wiki.wikijump.com`. This also adds support for sites to change slugs, which like users automatically leaves behind a site alias (though unlike users this does not consume a rename token as sites on the instance are assumed to be more controlled and not created by potential bad actors.)

This is accomplished by changing the `UserAliasService` to be the `AliasService`, which supports `AliasType::Site` and `AliasType::User`. When doing your usual methods, e.g. `UserService::get()`, then if a slug is used and it's an alias, that is redirected into the actual user it points to.

I tested by creating both a site and a user, modifying them, then changing their slugs and ensuring that the alias situation was correct. For instance, you can change a user back to a previous slug and it will not cause a database conflict (and also not deduct a rename token).

There is currently no support for site deletion, so deleting site aliases is not a concern.

I also added site aliases to the database seeder, so they can be specified for the initial database setup.

Note: If this PR needs to be rebased, amend the comment starting with "Get rid of ..." to fix the commit hash it reference.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1435, opened 2023-04-18
[WJ-1154] Add basic page route

emmiegit:
This is a first through-connection between a web request and the database. From a raw web request (e.g. `GET https://www.wikijump.localhost/start`), to Framerail (web server), to DEEPWELL, to the database, this is the proof-of-concept that delivers a JSON payload to Framerail containing all the relevant data for the page starting from the domain, slug, and session token. It also processes site redirects (aliases) and page redirects (slug normalization), and it has a stub for user permissions.

Currently it just dumps the JSON to the screen for debugging purposes:
image

But obviously this shows that all the data is there for Framerail to process and produce a proper page display. There are other restrictions in this version, such as not handling missing pages or sites properly yet.

This is implemented with a new `ViewService`, which provides high-level data to be consumed directly by Framerail from minimal data. It calls other services to turn say, a domain into a site ID, then the slug into a page ID, etc., and dumps it all into JSON to be consumed and rendered server-side.

This adds methods to `DomainService` to convert hostnames into sites (e.g. both `foo.wikijump.com` and `customdomain.com`). This requires adding the main and files domains into the configuration for portability.

The PR also has miscellaneous necessary changes:
* Removed the `.to_api()` method. This was intended to provide better error statuses, but has several issues, and it's not publicly-exposed anyways so it hardly matters.
* Renamed `methods/` -> `endpoints/` for clarity.
* Because SvelteKit doesn't allow GET requests with bodies, I've changed those internal routes to use PUT with a name change (e.g. `GET /user` -> `PUT /user/get`).


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1436, opened 2023-04-18
[WJ-1137] Add punycode-like names to default filter

emmiegit:
In HTTP, [punycode](https://en.wikipedia.org/wiki/Punycode) domains begin with `xn--`. We shouldn't allow this as it could be used for confusability attacks (making something that looks like `scp-wiki` but in Cryllic). This adds it to the default filter set.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1437, opened 2023-04-18
Prune legacy files

emmiegit:
It's been a bit since we've last done this.

This PR removes a bunch of old PHP stuff corresponding to things that are now implemented in DEEPWELL and/or Framerail. We no longer need them for reference, so we remove them to symbolize our progress in that respect.

As always, all legacy PHP files can be permanently found in the [`legacy-php`](https://github.com/scpwiki/wikijump/tree/legacy-php) branch.


Yossipossi1 (commented):
You love to see it.

Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1439, opened 2023-04-18
Fix mentions of "Wikidot"

emmiegit:
I noticed that the seeder made mention of "Wikidot" in a couple places we missed. We originally took these pages from the old gabrys repo start page, so that is why those mentions were left in here.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1440, opened 2023-04-19
Prune more legacy files

emmiegit:
I performed a more thorough search of `/legacy` than the last PR (#1437) to remove things which we already have implemented, or files so small that they're trivial and not of use when looking at old code.


Yossipossi1:
YES!!! FUCK YEAH!!!!

Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1441, opened 2023-04-19
[WJ-1171] Replace unmaintained chrono with time

emmiegit:
Within DEEPWELL, we previously used the `chrono` crate, which is unfortunately unmaintained, and as a dependency uses an old vulnerable version of `time`. This PR entirely removes the `chrono` dependency and replaces it with a modern version of `time`, which is both maintained, and has essentially the same feature set. All around it's an improvement.

ftml also made use of the `chrono` crate, though much less extensively. This dependency has also been replaced.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1445, opened 2023-04-21
Update rust dependencies (2)

emmiegit:
Well it turns out more than one sub-dependency still uses chrono, so that sucks. But we can still update ftml as bumped last PR.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1447, opened 2023-04-22
Remove deprecated dependencies of ftml

notgull:
- `lazy_static` has generally fallen out of favor due to `once_cell`, which will be soon incorporated into the standard library.
- `void` is no longer maintained by its original author and has been replaced by `std::convert::Infallible`.

### DIFF
--- a/ftml/Cargo.toml
+++ b/ftml/Cargo.toml
@@ -28,9 +28,9 @@ cfg-if = "1"
 enum-map = "2"
 entities = "1"
 latex2mathml = { version = "0.2", optional = true }
-lazy_static = "1"
 log = "0.4"
 maplit = "1"
+once_cell = "1.17.1"
 parcel_css = { version = "1.0.0-alpha.32", optional = true }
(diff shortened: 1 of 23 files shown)


emmiegit:
Huh, didn't know that. I hadn't really checked up on those crates, they've never caused build problems which is how I usually find out about issues.

This said regarding `void`, I have a counterpoint: it sounds cool.

notgull:
> This said regarding `void`, I have a counterpoint: it sounds cool.

Mmm, I just reviewed the public API, and I just realized that `void` is exposed in the public API. So we might not want to actually remove it to prevent a breaking change.

Not sure if `lazy_static` appears anywhere in the public API...

emmiegit:
I think it's fine to change void. Unfortunately there are some big changes that still need to happen and I've been playing kind of loose with semver. I do want to bump to 2.0 after some key things are changed but yeah this crate is mostly internal-use only.

emmiegit (approved):

emmiegit:
thanks @notgull!

notgull:
Contributor badge acquired

PR #1451, opened 2023-04-26
Fix wording in compatibility ID document 

emmiegit:


Yossipossi1 (approved):
No sure... this seems like a pretty big change to me.

emmiegit:
thanks @Yossipossi1

PR #1452, opened 2023-04-26
[WJ-1118] Add case-sensitivity logic to filter regex.

Yossipossi1:
This PR adds logic for specifying case-sensitivity for filters. This is performed by adding the `(?i)` flag to the beginning of the case-insensitive regex in the database, which specifies everything else in the regex is case-insensitive (or until it reaches a `(?-i)` flag). See [WJ-1118](https://scuttle.atlassian.net/browse/WJ-1118) for additional details.

[WJ-1118]: https://scuttle.atlassian.net/browse/WJ-1118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deepwell/src/services/filter/service.rs
+++ b/deepwell/src/services/filter/service.rs
@@ -118,7 +118,7 @@ impl FilterService {
                         model_regex.insert_str(0, "(?i)");
                     }
 
-                    model.regex = Set(model_regex)
+                    model.regex = Set(model_regex);
                 }
 
                 _ => {}
--- a/deepwell/src/services/filter/service.rs
+++ b/deepwell/src/services/filter/service.rs
@@ -121,6 +121,7 @@ impl FilterService {
                     model.regex = Set(model_regex);
                 }
 
+                // If the regex is being changed but is case-sensitive, do not touch it.
                 _ => {}
             }
         };
--- a/deepwell/src/services/filter/service.rs
+++ b/deepwell/src/services/filter/service.rs
@@ -109,7 +109,8 @@ impl FilterService {
 
                 // If the regex is not being changed, add/remove the flag from the database regex.
                 ProvidedValue::Unset => {
-                    let mut model_regex = model.get(filter::Column::Regex).as_ref().to_string();
+                    let mut model_regex =
+                        model.get(filter::Column::Regex).as_ref().to_string();
 
                     if case_sensitive {
                         model_regex = str!(model_regex.trim_start_matches("(?i)"));
@@ -331,7 +332,11 @@ impl FilterService {
                 Condition::all()
                     .add(filter::Column::SiteId.eq(site_id))
                     // Check for both case sensitive and insensitive variants
-                    .add(filter::Column::Regex.eq(regex).or(filter::Column::Regex.eq(format!("(?i){regex}"))))
+                    .add(
+                        filter::Column::Regex
+                            .eq(regex)
+                            .or(filter::Column::Regex.eq(format!("(?i){regex}"))),
+                    )
                     .add(filter::Column::DeletedAt.is_null()),
             )
             .one(txn)
--- a/deepwell/src/services/filter/service.rs
+++ b/deepwell/src/services/filter/service.rs
@@ -107,14 +107,14 @@ impl FilterService {
                     regex.insert_str(0, "(?i)")
                 }
 
-                // If the regex is not being changed, add/remove the flag from the database regex.
+                // If the regex is not being changed, remove (and conditionally readd) the
+                // case-insensitivity flag from the database's regex.
                 ProvidedValue::Unset => {
-                    let mut model_regex =
-                        model.get(filter::Column::Regex).as_ref().to_string();
+                    let mut model_regex = str!(model.get(filter::Column::Regex).as_ref());
 
-                    if case_sensitive {
-                        model_regex = str!(model_regex.trim_start_matches("(?i)"));
-                    } else {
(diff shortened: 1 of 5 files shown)


deepwell/src/services/filter/service.rs
```diff
@@ -91,6 +99,32 @@ impl FilterService {
             ..Default::default()
         };
 
+        // Handle case-sensitivity logic
+        if let ProvidedValue::Set(case_sensitive) = case_sensitive {
+            match regex {
+                // If the regex is being changed, add case-insensitivity flag if case-insensitive.
+                ProvidedValue::Set(ref mut regex) if !case_sensitive => {
+                    regex.insert_str(0, "(?i)")
```
emmiegit, line 107: What about cases where it is case sensitive and the `(?i)` should be removed? Seems like there's a 2×2 here, was/wasn't case sensitive and is/isn't case sensitive, like below.

deepwell/src/services/filter/service.rs
```diff
@@ -91,6 +99,32 @@ impl FilterService {
...
+
+                // If the regex is not being changed, add/remove the flag from the database regex.
+                ProvidedValue::Unset => {
+                    let mut model_regex =
+                        model.get(filter::Column::Regex).as_ref().to_string();
+
+                    if case_sensitive {
+                        model_regex = str!(model_regex.trim_start_matches("(?i)"));
+                    } else {
+                        model_regex.insert_str(0, "(?i)");
```
emmiegit, line 121: How does the logic verify this isn't inserted twice?

deepwell/src/services/filter/service.rs
```diff
@@ -91,6 +99,32 @@ impl FilterService {
...
+                    }
+
+                    model.regex = Set(model_regex)
```
emmiegit: Missing semicolon?

deepwell/src/services/filter/service.rs
```diff
@@ -91,6 +99,32 @@ impl FilterService {
...
+                }
+
+                _ => {}
```
emmiegit: If `ProvidedValue` isn't set? I don't understand this case.

Would a comment be helpful here?

deepwell/src/services/filter/service.rs
Yossipossi1: I'm under the impression that if someone adds `(?i)` to the start of a case-sensitive regex, they intend to make it case-insensitive. For this reason, it isn't removed if the case-sensitive option is toggled. It will be interpreted by the database as case-insensitive and would be indistinguishable from regularly toggling the case sensitivity.

deepwell/src/services/filter/service.rs
Yossipossi1: Good catch, since an update to the case sensitivity doesn't necessarily imply it is changing the case sensitivity, just that a new value was provided. Will fix.

deepwell/src/services/filter/service.rs
Yossipossi1: Yep, will fix.

deepwell/src/services/filter/service.rs
Yossipossi1: This catches if `ProvidedValue` is `Set(_)` but case_sensitivity is true. See prior comments. This is just to satisfy the compiler by telling it to do nothing in this circumstance.

deepwell/src/services/filter/service.rs
emmiegit: Okay, now that I'm at my desktop and thinking this through, this make sense. For ease of readability, consider a `Provided::Value::Set(_) => (),` branch underneath with a comment for this case (then the `_ => ()` can be removed too).

deepwell/src/services/filter/service.rs
```diff
@@ -91,6 +99,33 @@ impl FilterService {
...
+                ProvidedValue::Set(ref mut regex) if !case_sensitive => {
+                    regex.insert_str(0, "(?i)")
+                }
+
+                // If the regex is not being changed, remove (and conditionally readd) the
+                // case-insensitivity flag from the database's regex.
+                ProvidedValue::Unset => {
+                    let mut model_regex = str!(model.get(filter::Column::Regex).as_ref());
+
+                    model_regex = str!(model_regex.trim_start_matches("(?i)"));
```
emmiegit, line 118: Instead of re-allocating, since you already get an owned string on line 113, you can instead do:
```suggestion
                    if model_regex.starts_with("(?i)") {
                        // SAFETY: In this branch we know the string starts with 4 ASCII characters
                        model_regex.drain(..4);
                    }
```

emmiegit (approved):

deepwell/src/services/filter/service.rs
Yossipossi1: I don't know if this is necessarily better. I feel that `trim_start_matches()` is more understandable than `drain(..4)`. Additionally I don't know if this saves much on computation time, but I don't have any hard statistics on this. I also think it's needlessly verbose comparatively, and also doesn't strip the prefix multiple times if present like `trim_start_matches` does.

deepwell/src/services/filter/service.rs
Yossipossi1: Sounds good to me

deepwell/src/services/filter/service.rs
emmiegit: Okay I'll add it as a helper function instead then.

PR #1455, opened 2023-04-28
[WJ-1122] Add minimum byte length for user names/slugs.

Yossipossi1:
This PR adds a config option for the minimum bytes a user's name/slug must be, and adds checks for if this condition is not satisfied. See [WJ-1122](https://scuttle.atlassian.net/browse/WJ-1122).

Please be aware I am very sleep deprived, so I may have missed something important. Thorough review would be appreciated (there are only a few lines).

[WJ-1122]: https://scuttle.atlassian.net/browse/WJ-1122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deepwell/config.example.toml
+++ b/deepwell/config.example.toml
@@ -197,3 +197,6 @@ max-name-changes = 3
 #
 # Set to 0 to disable.
 refill-name-change-days = 90
+
+# The minimum length for a user's name in bytes.
+min-byte-length = 3
--- a/deepwell/config.example.toml
+++ b/deepwell/config.example.toml
@@ -199,4 +199,13 @@ max-name-changes = 3
 refill-name-change-days = 90
 
 # The minimum length for a user's name in bytes.
+# A user who attempts to register must have a slug which is at least this long.
+#
+#  The intent is to prevent the proliferation of very short, unidentifiable usernames
+# such as single Latin letters (e.g. "N"). However, we check *bytes* specifically to
+# avoid Anglo-centricism, as many full names in other languages fit within two letters
+# (consider CJK for instance). Single characters outside of the main ASCII block
+# usually take up two or three bytes themselves, making it a non-issue for those languages. 
+#
+# See WJ-1122.
 min-byte-length = 3
--- a/deepwell/src/services/alias/service.rs
+++ b/deepwell/src/services/alias/service.rs
@@ -76,6 +76,9 @@ impl AliasService {
         // Then we check that the new slug doesn't already exist.
         // This also checks aliases, though we also verify down below that
         // it actually finds conflicts properly.
+        //
+        // If the alias is for a user, also ensures that it is over
+        // the minimum name length.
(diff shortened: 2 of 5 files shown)


deepwell/config.example.toml
```diff
@@ -197,3 +197,6 @@ max-name-changes = 3
 #
 # Set to 0 to disable.
 refill-name-change-days = 90
+
+# The minimum length for a user's name in bytes.
```
emmiegit, line 201: I would suggest a bit of background for this option.

```suggestion
# The minimum length for a user's name in bytes.
# A user who attempts to register must have a slug which is at least this long.
#
#  The intent is to prevent the proliferation of very short, unidentifiable usernames
# such as single Latin letters (e.g. "N"). However, we check *bytes* specifically to
# avoid Anglo-centricism, as many full names in other languages fit within two letters
# (consider CJK for instance). Single characters outside of the main ASCII block
# usually take up two or three bytes themselves, making it a non-issue for those languages. 
#
# See WJ-1122.
```

deepwell/src/services/alias/service.rs
```diff
@@ -76,6 +76,9 @@ impl AliasService {
         // Then we check that the new slug doesn't already exist.
         // This also checks aliases, though we also verify down below that
         // it actually finds conflicts properly.
+        //
+        // If the alias is for a user, also ensures that it is over
```
emmiegit:
```suggestion
        // If the alias is for a user, also ensures that it is at least
```

"over" means `>`, "at least" means `>=`

deepwell/src/services/alias/service.rs
```diff
@@ -112,6 +115,11 @@ impl AliasService {
 
                     return Err(Error::Conflict);
                 }
+
+                if slug.bytes().len() < ctx.config().minimum_name_bytes {
+                    tide::log::error!("User's name does not contain enough bytes.");
```
emmiegit: This could be more helpful. At least I think "not long enough" is a better error message. (This would need to be copied to the other locations as well.)
```suggestion
                    tide::log::error!(
                        "User's name is not long enough ({} < {})",
                        slug.bytes().len(),
                        ctx.config().minimum_name_bytes,
                    );
```

deepwell/src/services/alias/service.rs
```diff
@@ -112,6 +115,11 @@ impl AliasService {
 
                     return Err(Error::Conflict);
                 }
+
+                if slug.bytes().len() < ctx.config().minimum_name_bytes {
```
emmiegit: Fyi, `string.len()` is the same as `string.bytes().len()`. However I'm fine with making it explicit since it is more intentional and compiles to the same code.

emmiegit (approved):

PR #1457, opened 2023-05-01
[WJ-1172] Page route follow-ups

emmiegit:
Implement the various leftover issues from [WJ-1172](https://scuttle.atlassian.net/browse/WJ-1172), which primarily concern properly handling missing data. For instance, if a site doesn't exist, then display the "this site doesn't exist" page, or pulling `_404` for a site as the "this page doesn't exist" response.

This _does not_ implement HTTP status codes for these errors (e.g. returning 403 along with `_public` if you can't see a page, see [WJ-1180](https://scuttle.atlassian.net/browse/WJ-1180).

**Issue tasks:**
* Make `PageArguments` case-insensitive.
* Consume any unused page arguments and emit warnings for each.
* Add `/rerender` page argument key.
* Catch and display error message if site is not found.
* Catch and display error message if page is not found.
* Verify page slug normalization.
* Add stubbed check for page access, display error message if not. (Tested with hardcoded "cannot access")

**Changes:**
* Fix configuration key inconsistent naming issues with #1455 preventing a local deploy from running.
* Fix typo in `filters.json` seeder file preventing local running.
* Add both "dotted" and "undotted" forms of the domain in the runtime configuration, for convenience.
 * For instance, "undotted" would be `wikijump.com` and "dotted" would be `.wikijump.com`, for concatenation with a site slug name for instance.
 * Either form can be entered in the configuration, and it will be handled appropriately.
* Adds `fluent_str!` macro to easily make `FluentValue` strings which are borrowed.
* Disable the clippy lint about large enum variants.
* Replace `dotenv` (unmaintained) with `dotenvy`.
* Upgrade `wikidot-path` (supersedes #1449).
* Modify `DomainService` so that it passes back some information on missing domains, for better error messages (specifically, custom domain vs no such site for slug).
* Create the `SpecialPageService`, which gets the wikitext for a special page (if exists, falling back to a language-specific default if not) and then renders and returns it for display.
* Add configuration keys so that values like `_404` are not hardcoded.
* Add localization keys for fallback error messages if no special pages are specified.
 * Because "no such site" messages are rare and have no site from which to draw upon special pages, nor a site to render in, they are not part of the `SpecialPageService` and instead draw rendered HTML directly from the localization, rather than rendering wikitext.

[WJ-1172]: https://scuttle.atlassian.net/browse/WJ-1172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WJ-1180]: https://scuttle.atlassian.net/browse/WJ-1180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1461, opened 2023-05-08
[WJ-1177] Add DEEPWELL configuration validation to CI

emmiegit:
To avoid issues where typos or not-yet-updated documentation (i.e. `config.example.toml`) cause runtime issues with using such a configuration file in a deployment, here we add a CI step to validate configuration files before merge.

This is implemented by adding a concept of "special action" to deepwell, set via env variable, which can specify the executable to perform a special task instead of running a server. In this case, we have it coherency check all the configuration files passed in on the command line.

### DIFF
--- a/deepwell/src/config/special_action.rs
+++ b/deepwell/src/config/special_action.rs
@@ -18,9 +18,54 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
-#[derive(Debug)]
-pub enum SpecialAction {}
+//! Perform a "special action" instead of normal server execution.
+//!
+//! This is useful in contexts such as CI, where we want DEEPWELL to
+//! not run as a daemon, but instead perform a special action or check,
+//! as if motivated by a script.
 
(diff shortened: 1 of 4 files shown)


deepwell/src/config/special_action.rs
```diff
@@ -0,0 +1,72 @@
...
+use std::path::PathBuf;
+use std::{env, process};
+
+pub fn run_special_action() {
+    // Get action name, if specified.
+    // Otherwise return and perform normal execution.
+    let action_name = match env::var("DEEPWELL_SPECIAL_ACTION") {
+        Ok(value) => value,
+        Err(_) => return,
+    };
```
Yossipossi1: I believe patterns like this could be simplified into just this one-liner:
```suggestion
    let Ok(action_name) = env::var("DEEPWELL_SPECIAL_ACTION") else { return };
```

Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1466, opened 2023-05-10
Add checks to disallow empty slugs

emmiegit:
Turns out that you were allowed to make users/sites/pages with an empty slug, since it was technically already normalized. Oops!

Tested with local deployment.


Yossipossi1 (approved):
The classic blunder!

emmiegit:
thanks @Yossipossi1

PR #1469, opened 2023-05-15
Add special page for banned users

emmiegit:
Currently if you are banned from a Wikidot site, you are shown a generic, un-customizable page permissions error. This is poor UX, as EN at least doesn't mind banned users _reading_ the site (since it's public anyways), and the error message suggests it's an issue on _our_ end and not _their_ end, resulting in unnecessary inquiries.

We won't be carrying this behavior over, but we can add a site setting to, if the site staff decide to, disallow banned users from having read access. In this case, we should provide a special message for banned users, and allow it to be customized.

This is done by adding a `_ban` special page which is like `_public` but displays for banned users, specifically where site settings also disallow them from reading the site. If they are banned but have read-only access, this error is _not_ displayed.

### DIFF
--- a/deepwell/config.example.toml
+++ b/deepwell/config.example.toml
@@ -224,6 +224,13 @@ missing = "_404"
 # Fallback string is localized as "wiki-page-private"
 private = "_public"
 
+# When a user is banned from the site.
+# This is only shown if the site disallows banned users from
+# being able to see pages.
+#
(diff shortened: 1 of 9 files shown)


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1470, opened 2023-05-15
Add specific category support to missing special pages

emmiegit:
When you visit a page that doesn't exist, it pulls the message to display from `_404`. But to be more specific, it first checks the _current category's_ `_404` page, then if that doesn't exist it goes to the site-wide one.

I implement this by changing the single slug to check into a list, and then the first one to exist is used. Then if all are missing, it uses the localization fallback (this has been split into a helper method). For missing pages, I have a little routine that adds the category-specific slug if contextually relevant.

Missing page, category-specific `_404`:
image

Missing page, no category-specific `_404`:
image


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1493, opened 2023-06-16
Add cdn.wikijump.com to default filter

emmiegit:
Similar to other technical-related terms like "login" and "account", we should disallow `cdn.wikijump.com` from being a site slug.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1503, opened 2023-06-19
[WJ-1133] Implement MailCheck Email Validation

Yossipossi1:
[Link to Issue](https://scuttle.atlassian.net/browse/WJ-1133)

This PR adds email verification using the MailCheck API, as well as the EmailService.

I'll be honest in that I feel like I missed something or did something wrong somewhere, so I would appreciate a more thorough review if possible.

### DIFF
--- a/deepwell/src/models/user.rs
+++ b/deepwell/src/models/user.rs
@@ -24,6 +24,7 @@ pub struct Model {
     pub last_renamed_at: Option<OffsetDateTime>,
     #[sea_orm(column_type = "Text")]
     pub email: String,
+    pub email_is_alias: Option<bool>,
     pub email_verified_at: Option<OffsetDateTime>,
     #[sea_orm(column_type = "Text")]
     pub password: String,
--- a/deepwell/src/services/email/service.rs
+++ b/deepwell/src/services/email/service.rs
@@ -1,0 +1,28 @@
+/*
+ * services/email/service.rs
+ *
+ * DEEPWELL - Wikijump API provider and database manager
+ * Copyright (C) 2019-2023 Wikijump Team
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+ use super::prelude::*;
+
+ #[derive(Debug)]
+pub struct EmailService;
+
+impl EmailService {
+
+}
--- a/deepwell/src/services/email/service.rs
+++ b/deepwell/src/services/email/service.rs
@@ -25,10 +25,7 @@ pub struct EmailService;
 
 impl EmailService {
     /// Validates an email through the MailCheck API.
-    pub async fn validate(
-        _ctx: ServiceContext<'_>,
-        email: String,
-    ) -> Result<EmailValidationOutput> {
+    pub async fn validate(email: &String) -> Result<EmailValidationOutput> {
         // Sends a GET request to the MailCheck API and deserializes the response.
         let mailcheck = surf::get(format!("https://api.mailcheck.ai/email/{email}"))
             .send()
--- a/deepwell/src/services/email/service.rs
+++ b/deepwell/src/services/email/service.rs
@@ -44,7 +44,7 @@ impl EmailService {
             // Invalid request.
             400 => {
                 tide::log::error!(
-                    "MailCheck API returned status 400: {:?}",
+                    "MailCheck API request failed with bad response: {:?}",
                     mailcheck.error,
                 );
                 return Err(Error::BadRequest);
@@ -53,7 +53,7 @@ impl EmailService {
             // Exceeded rate limit.
             429 => {
                 tide::log::error!(
-                    "MailCheck API returned status 429: {:?}",
+                    "MailCheck API hit ratelimit: {:?}",
                     mailcheck.error,
                 );
                 return Err(Error::RateLimited);
--- a/deepwell/src/services/email/service.rs
+++ b/deepwell/src/services/email/service.rs
@@ -18,11 +18,24 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
- use super::prelude::*;
+use super::prelude::*;
 
  #[derive(Debug)]
 pub struct EmailService;
 
 impl EmailService {
+    // Validates an email through the MailCheck API. 
+    pub async fn validate(
+        ctx: ServiceContext<'_>,
+        email: String,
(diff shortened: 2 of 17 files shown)


deepwell/src/models/user.rs
```diff
@@ -24,6 +24,7 @@ pub struct Model {
     pub last_renamed_at: Option<OffsetDateTime>,
     #[sea_orm(column_type = "Text")]
     pub email: String,
+    pub email_is_alias: Option<bool>,
```
emmiegit, line 27: So the three options here are:
* `true` -- Email is known to be an alias. (e.g. `foo+name@gmail.com`)
* `false` -- Email is not known to be an alias.
* `null` -- Email has not been run through the verification service.
?

deepwell/src/services/email/service.rs
```diff
@@ -0,0 +1,95 @@
...
+        // Check request status.
+        match mailcheck.status {
+            // Valid request.
+            200 => {}
+
+            // Invalid request.
+            400 => {
+                tide::log::error!(
+                    "MailCheck API returned status 400: {:?}",
+                    mailcheck.error
```
emmiegit:
```suggestion
                    mailcheck.error,
```

deepwell/src/services/email/service.rs
```diff
@@ -0,0 +1,95 @@
...
+                );
+                return Err(Error::BadRequest);
+            }
+
+            // Exceeded rate limit.
+            429 => {
+                tide::log::error!(
+                    "MailCheck API returned status 429: {:?}",
+                    mailcheck.error
```
emmiegit:
```suggestion
                    mailcheck.error,
```

deepwell/src/services/email/service.rs
```diff
@@ -0,0 +1,95 @@
...
+
+        // Check request status.
+        match mailcheck.status {
+            // Valid request.
+            200 => {}
+
+            // Invalid request.
+            400 => {
+                tide::log::error!(
+                    "MailCheck API returned status 400: {:?}",
```
emmiegit: Since we're checking individual status codes rather than just doing `MailCheck API returned {}`, I'd rather these messages be like `MailCheck API request fails with bad response:` and `MailCheck API ratelimited:`

deepwell/src/services/email/service.rs
```diff
@@ -0,0 +1,95 @@
...
+                );
+                return Err(Error::RateLimited);
+            }
+
+            // Other statuses.
+            _ => {
+                tide::log::warn!(
+                    "MailCheck API returned status {}: {:?}",
+                    mailcheck.status,
+                    mailcheck.error
```
emmiegit:
```suggestion
                    mailcheck.error,
```

deepwell/src/services/email/service.rs
```diff
@@ -0,0 +1,95 @@
...
+
+        // Check if the domain has any MX records.
+        if !mailcheck.mx {
+            output.valid = false;
+            output.classification = EmailClassification::Invalid;
+        }
+
+        if mailcheck.did_you_mean.is_some() {
+            output.did_you_mean = mailcheck.did_you_mean;
+        }
```
emmiegit: Normally wouldn't this be `output.did_you_mean = mailcheck.did_you_mean` unconditionally if we're copying `Option` for `Option` anyways? Otherwise it seems like a more canonical conditional would be
```rust
if let Some(suggestion) = mailcheck.did_you_mean {
    output.did_you_mean = Some(suggestion);
}
```
?

Between the two I'd rather do unconditional copy since that avoids a branch altogether.

deepwell/src/services/email/service.rs
```diff
@@ -0,0 +1,95 @@
...
+
+        Ok(output)
```
emmiegit, line 89: This function is a nice clean way of extracting the relevant information and placing it into the output struct, I approve.

deepwell/src/services/user/service.rs
```diff
@@ -159,14 +161,57 @@ impl UserService {
...
+
+            match email_validation_output.classification {
+                EmailClassification::Normal => {
+                    tide::log::info!("User {slug}'s email was verified successfully");
+                    Some(false)
+                }
+
+                EmailClassification::Alias => {
+                    tide::log::info!(
+                        "User {slug}'s email was verified successfully (as an alias)"
```
emmiegit:
```suggestion
                        "User {slug}'s email was verified successfully (as an alias)",
```

deepwell/src/services/user/service.rs
```diff
@@ -159,14 +161,57 @@ impl UserService {
...
+                EmailClassification::Disposable => {
+                    tide::log::error!(
+                        "User {slug}'s email is disposable and did not pass verification"
+                    );
+                    return Err(Error::DisallowedEmail);
+                }
+
+                EmailClassification::Invalid => {
+                    tide::log::error!(
+                        "User {slug}'s email is invalid and did not pass verification"
```
emmiegit:
```suggestion
                        "User {slug}'s email is invalid and did not pass verification",
```

deepwell/src/services/user/service.rs
```diff
@@ -159,14 +161,57 @@ impl UserService {
...
+                EmailClassification::Alias => {
+                    tide::log::info!(
+                        "User {slug}'s email was verified successfully (as an alias)"
+                    );
+                    Some(true)
+                }
+
+                EmailClassification::Disposable => {
+                    tide::log::error!(
+                        "User {slug}'s email is disposable and did not pass verification"
```
emmiegit:
```suggestion
                        "User {slug}'s email is disposable and did not pass verification",
```

deepwell/src/services/user/service.rs
```diff
@@ -159,14 +161,57 @@ impl UserService {
...
         };
 
+        // Perform email verification.
+        //
+        // If the email is either disposable or invalid, propogate the error upwards and
+        // stop the account creation. If the email passes validation, mark if it's an alias
+        // or not.
+        //
+        // The assigned variable is also used to check whether email validation occurred, as it
+        // will always be `Some` if validation occurred and `None` otherwise.
```
emmiegit, line 177: Good comment.

deepwell/src/services/user/service.rs
```diff
@@ -159,14 +161,57 @@ impl UserService {
...
+
+                EmailClassification::Invalid => {
+                    tide::log::error!(
+                        "User {slug}'s email is invalid and did not pass verification"
+                    );
+                    return Err(Error::InvalidEmail);
+                }
+            }
+        } else {
+            None
```
emmiegit, line 210: Since it's a long branch, a comment could be helpful to illustrate what this case is:
```suggestion
            // Skipping email verification
            None
```

emmiegit (approved):

emmiegit:
Have you been able to test this locally with sample user creation / email calls?

For the merge conflict I would suggest rebasing and simply rebuilding the `Cargo.lock` at each conflict.

deepwell/src/models/user.rs
Yossipossi1: Yep.

emmiegit: Oh I forgot to mention in my review, you need to add this column to the [migration](https://github.com/scpwiki/wikijump/blob/develop/deepwell/migrations/20220906103252_deepwell.sql#L17) too.

Yossipossi1:
> Have you been able to test this locally with sample user creation / email calls?
> 
> For the merge conflict I would suggest rebasing and simply rebuilding the `Cargo.lock` at each conflict.

Added a route for it and also tested it (thank you :) )

PR #1504, opened 2023-06-23
Update development.md

Yossipossi1:
This PR aims to update the `development.md` file, which contains instructions for setting up a local instance of Wikijump.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -92,6 +92,7 @@ $ docker images        # List images
 $ docker rmi [ID]      # Remove the image with this ID
 ```
 
+You can also use `docker system prune`, which deletes *everything* unused, though this should be used sparingly.
 
 ## Entering the container
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -45,7 +45,7 @@ Node, NPM, and PNPM are well-behaved on Windows and Linux, and the difference in
 
 ## Setup: Configuration
 
-There are a couple of configuration files that need to be initialized prior to running your instance of Wikijump. These will be the `config.toml` and `.env` files, both located in the `deepwell` subdirectory. Both of these files can be copied from their `.example` counterparts without changing them, though it is worth looking through them briefly to understand what can be configurated for your instance.
+There are a couple of configuration files that need to be initialized prior to running your instance of Wikijump. These will be the `config.toml` and `.env` files, both located in the `deepwell` subdirectory. For a typical local deploy, both of these files can be copied from their `.example` counterparts without changing them, though it is worth looking through them briefly to understand what can be configurated for your instance.
 
 There is also a Docker configuration file that configures the various containers that host Wikijump in the local environment. You can find this file in `install/local/dev/`, named `docker-compose.yaml`, alongside `docker-compose.dev.yaml` (which provides various helpful tools for developing locally).
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -49,7 +49,7 @@ There are a couple of configuration files that need to be initialized prior to r
 
 There is also a Docker configuration file that configures the various containers that host Wikijump in the local environment. You can find this file in `install/local/dev/`, named `docker-compose.yaml`, alongside `docker-compose.dev.yaml` (which provides various helpful tools for developing locally).
 
-Notice that in `docker-compose.yaml`, there are configuration options for the domains to use. For development purposes, these are set to `wikijump.localhost`. This is the domain you will be connecting to, e.g. `http://www.wikijump.localhost`. The TLD `.localhost` is just like the usual `localhost` domain. Even when running locally, HTTPS is used. Because this certificate is self-signed, you will need to dismiss the certificate warning.
+Notice that in `docker-compose.yaml`, there are configuration options for the domains to use. For development purposes, these are set to `wikijump.localhost`. This is the domain you will be connecting to, e.g. `https://www.wikijump.localhost`. The TLD `.localhost` is just like the usual `localhost` domain. Even when running locally, HTTPS is used. Because this certificate is self-signed, you will need to dismiss the certificate warning.
 
 ## Setup: Dependencies
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -19,8 +19,6 @@ The recommended way to install Wikijump is via Docker. Docker is a way of contai
 You will need [Docker](https://www.docker.com/) installed and running:
 
 <pre>$ sudo apt install docker.io</pre>
-Then start the Docker daemon via dockerd:</p>
-<pre>$ sudo dockerd</pre></p>
 
 ## Setup: Utilities and Programs
 
@@ -37,7 +35,7 @@ Node, NPM, and PNPM are well-behaved on Windows and Linux, and the difference in
 
 There are a couple of configuration files that need to be initialized prior to running your instance of Wikijump. These will be the `config.toml` and `.env` files, both located in the `deepwell` subdirectory. Both of these files can be copied from their `.example` counterparts without changing them, though it is worth looking through them briefly to understand what can be configurated for your instance.
 
-There is also a Docker configuration file that configures the various containers that host Wikijump in the local environment. You can find this file in `install/local/dev/`, named `docker-compose.yaml`.
+There is also a Docker configuration file that configures the various containers that host Wikijump in the local environment. You can find this file in `install/local/dev/`, named `docker-compose.yaml`, alongside `docker-compose.dev.yaml` (which provides various helpful tools for developing locally).
 
 Notice that in `docker-compose.yaml`, there are configuration options for the domains to use. For development purposes, these are set to `wikijump.localhost`. This is the domain you will be connecting to, e.g. `http://www.wikijump.localhost`. The TLD `.localhost` is just like the usual `localhost` domain. Currently, the domain uses `HTTPS`, which can result in a certificate error as it is self signed — this can just be ignored.
 
@@ -101,7 +99,7 @@ If you want to enter a container to make temporary changes, you can do so by ent
 $ docker exec -it [container id] sh
 ```
 
-...where `[container id]` is the ID of the corresponding container from `docker ps`.
+...where `[container id]` is the ID of the corresponding container from `docker ps`. (`sh` can be replaced with a different command.)
 
 One reason you may need to enter the container is to manually adjust the Wikijump config. For example, if you use a port other than 80 for your Docker container, you will need to edit `site.custom_domain` to add the port number (e.g. "`www.wikijump.localhost:8080`"). Alternatively, use curl to set the domain directly (e.g. "`-H 'www.wikijump.localhost'`")
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -6,32 +6,21 @@ This document will explain how to set up Wikijump on your machine for local deve
 
 The `install` folder has everything you need to run a local Wikijump install either in a container or on metal or a VM.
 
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -6,32 +6,21 @@ This document will explain how to set up Wikijump on your machine for local deve
...
-    </td>
-    <td>
-      <p>Install <a href="https://docs.docker.com/docker-for-windows/install-windows-home">Docker Desktop</a>, which you'll be using as the Docker daemon, and leave it running in the background.</p>
-      <p>If you have set <code>appendWindowsPath=false</code> in your WSL config, then you may hit an error along the lines of <code>"docker-credential-desktop.exe": executable file not found in $PATH</code>. In this case you should either add <code>/mnt/c/Program\ Files/Docker/Docker/resources/bin</code> to your PATH, or <a href="https://github.com/rossjrw/dotfiles/blob/3c5445abb138b735cc3caf61f070c9125fa87d2f/.profile#L28">create a symlink</a> in a directory that is in your PATH already.</p>
-    </td>
-  </tr></tbody>
-</table>
+<pre>$ sudo apt install docker.io</pre>
+Then start the Docker daemon via dockerd:</p>
+<pre>$ sudo dockerd</pre></p>
```
emmiegit: We should be recommending they start the docker service, not running the daemon manually. I remember talking about a table or similar which showed systemd vs WSL2 `service` invocations for starting the Docker daemon.

Presently this assumes the user is using Ubuntu (or specifically WSL2?), this should get table-ized or put into the "WIndows" quote, or something like that.

docs/development.md
```diff
@@ -46,13 +35,11 @@ Node, NPM, and PNPM are well-behaved on Windows and Linux, and the difference in
...
+There are a couple of configuration files that need to be initialized prior to running your instance of Wikijump. These will be the `config.toml` and `.env` files, both located in the `deepwell` subdirectory. Both of these files can be copied from their `.example` counterparts without changing them, though it is worth looking through them briefly to understand what can be configurated for your instance.
 
-Notice that in `docker-compose.yaml`, there are configuration options for the domains to use. For development purposes, these are set to `wikijump.localhost`. This is the domain you will be connecting to, e.g. `http://www.wikijump.localhost`. The TLD `.localhost` is just like the usual `localhost` domain. Your browser trusts both of these domains implicitly so there is no need to use `HTTPS`.
+There is also a Docker configuration file that configures the various containers that host Wikijump in the local environment. You can find this file in `install/local/dev/`, named `docker-compose.yaml`.
```
emmiegit: I would also mention `docker-compose.dev.yaml`, which provides various helpfuls for developing locally (mapping local volumes and such).

Basically, `docker-compose.yaml` is the basic setup describing all the (local) containers and how they can talk to each other, and `docker-compose.dev.yaml` adds various conveniences on top, but is (at least should be) optional from the base one.

docs/development.md
```diff
@@ -64,58 +51,57 @@ $ pnpm install
...
 ## Building
 
-You can now finally build the Docker images using the following command:
-
-```sh
-$ pnpm dev build
-```
-
-_This might take a surprisingly long time_. Thankfully, Docker's build step is _heavily_ cached. It won't take this long again (unless of course you need to change one of those early steps).
+You can now finally build the Docker images. First, start docker:
```
emmiegit:
```suggestion
You can now finally build the Docker images. First, start Docker:
```

docs/development.md
```diff
@@ -64,58 +51,57 @@ $ pnpm install
...
 
-## Development
+<table>
+<thead><tr><th>Systemd Distros</th><th>WSL2</th></tr></thead>
```
emmiegit:
```suggestion
<thead><tr><th>systemd Distros</th><th>WSL2</th></tr></thead>
```

docs/development.md
```diff
@@ -64,58 +51,57 @@ $ pnpm install
...
+<tbody valign="top">
+<td><p><pre>$ sudo systemctl --user start docker-desktop</pre></p></td>
```
emmiegit: Oh I see the confusion now. Hmm, not sure about `docker-desktop` specifically? I think it would be better, in setup, to suggest turning on Docker as a _daemon_ (i.e. `sudo systemctl enable --now docker.service`) and then also mentioning that Docker desktop is a thing later.

"Start" means run now (but on restart it'll be gone).
"Enable" means automatically run on startup.
"Enable now" means automatically run on startup, and if it's not running right now, also start it.

docs/development.md
```diff
@@ -64,58 +51,57 @@ $ pnpm install
...
 
-This command will run everything in watch mode, and start the Docker containers. However, it will first ask you if you want to build the containers, as in what `pnpm dev build` does. This is asked because building may take a long time.
-
-Once everything has started, you can connect to `http://www.wikijump.localhost/`. Changes you make to the codebase should automatically be applied to the containers, as your machine's filesystem has been "bound" to the containers' filesystem. This is one-way, so a container can't modify your filesystem.
+_This might take some time_. Thankfully, Docker's build step is _heavily_ cached. 
 
-You can just kill the terminal (`CTRL + C` usually) when you want to stop the server.
+> ### For Windows:
+>
+> You may encounter various errors involving file permissions if using Windows-based tools alongside WSL2. It is recommend that you fix these issues by either granting the correct file permissions to any that Windows may have modified, or by redownloading the repository from github.
```
emmiegit:
```suggestion
> You may encounter various errors involving file permissions if using Windows-based tools alongside WSL2. It is recommend that you fix these issues by either granting the correct file permissions to any that Windows may have modified, or by re-cloning the repository purely within WSL2.
```

docs/development.md
```diff
@@ -64,58 +51,57 @@ $ pnpm install
...
 ```
 
-## Using `sudo`
-
-If your environment requires docker commands be run using `sudo`, the normal `pnpm dev` and `pnpm compose` commands will fail. You can use the `sudo` variants of these commands, `pnpm dev sudo` and `pnpm compose-sudo`.
 
 ## Entering the container
 
-If you want to enter the container to make temporary changes, you can do so by entering it with a CLI. From Docker Desktop, after running the containers, find the Wikijump app and within it the container you wish to enter, then click the 'CLI' button. Or from the command line:
+If you want to enter a container to make temporary changes, you can do so by entering it with a CLI. From Docker Desktop, after running the containers, find the Wikijump app and within it the container you wish to enter, then click the 'CLI' button. Or from the command line:
```
emmiegit, line 99:
```suggestion
If you want to enter a container to make temporary changes, you can do so by entering it with a CLI. From Docker Desktop, after running the containers, find the Wikijump app and within it the container you wish to enter, then click the 'CLI' button. Or from the command line: (you can replace `sh` with a different command)
```

docs/development.md
```diff
@@ -64,58 +51,57 @@ $ pnpm install
...
 
 ```
-$ docker exec -it [name of container] sh
+$ docker exec -it [container id] sh
 ```
 
-...where `[name of container]` is the name of the container from `docker container ls`.
+...where `[container id]` is the ID of the corresponding container from `docker ps`.
```
emmiegit: Doesn't `docker container ls` also work?

Yossipossi1: Yes, though `docker ps` is shorter. They are functionally identical, though I figured having it be the shorter command would make it easier to remember. I can revert this change if you think `container ls` is more recognizable though.

docs/development.md
```diff
@@ -46,13 +45,11 @@ Node, NPM, and PNPM are well-behaved on Windows and Linux, and the difference in
 
 ## Setup: Configuration
 
-Wikijump has a lot of configuration options that can set. The primary configuration file can be found in `web/conf/`, named `wikijump.ini`. Freshly cloned, your Wikijump repo won't have this file. However, it will have an example configuration, `wikijump.ini.example`, which you can just copy and rename to `wikijump.ini` for the time being.
+There are a couple of configuration files that need to be initialized prior to running your instance of Wikijump. These will be the `config.toml` and `.env` files, both located in the `deepwell` subdirectory. Both of these files can be copied from their `.example` counterparts without changing them, though it is worth looking through them briefly to understand what can be configurated for your instance.
```
emmiegit:
```suggestion
There are a couple of configuration files that need to be initialized prior to running your instance of Wikijump. These will be the `config.toml` and `.env` files, both located in the `deepwell` subdirectory. For a typical local deploy, both of these files can be copied from their `.example` counterparts without changing them, though it is worth looking through them briefly to understand what can be configurated for your instance.
```

docs/development.md
```diff
@@ -46,13 +45,11 @@ Node, NPM, and PNPM are well-behaved on Windows and Linux, and the difference in
...
 
-Notice that in `docker-compose.yaml`, there are configuration options for the domains to use. For development purposes, these are set to `wikijump.localhost`. This is the domain you will be connecting to, e.g. `http://www.wikijump.localhost`. The TLD `.localhost` is just like the usual `localhost` domain. Your browser trusts both of these domains implicitly so there is no need to use `HTTPS`.
-
-Finally, you will need a `.env` file in the `web/` root. There is a `.env.example` file that you can copy and rename.
+Notice that in `docker-compose.yaml`, there are configuration options for the domains to use. For development purposes, these are set to `wikijump.localhost`. This is the domain you will be connecting to, e.g. `http://www.wikijump.localhost`. The TLD `.localhost` is just like the usual `localhost` domain. Even when running locally, HTTPS is used. Because this certificate is self-signed, you will need to dismiss the certificate warning.
```
emmiegit:
```suggestion
Notice that in `docker-compose.yaml`, there are configuration options for the domains to use. For development purposes, these are set to `wikijump.localhost`. This is the domain you will be connecting to, e.g. `https://www.wikijump.localhost`. The TLD `.localhost` is just like the usual `localhost` domain. Even when running locally, HTTPS is used. Because this certificate is self-signed, you will need to dismiss the certificate warning.
```

docs/development.md
```diff
@@ -64,58 +61,47 @@ $ pnpm install
...
+> ### For Windows:
+>
+> You may encounter various errors involving file permissions if using Windows-based tools alongside WSL2. It is recommend that you fix these issues by either granting the correct file permissions to any that Windows may have modified, or by re-cloning the repository purely within WSL2.
 
-Once everything has started, you can connect to `http://www.wikijump.localhost/`. Changes you make to the codebase should automatically be applied to the containers, as your machine's filesystem has been "bound" to the containers' filesystem. This is one-way, so a container can't modify your filesystem.
+Once everything has started, you can connect to `http://www.wikijump.localhost/`. Changes you make to the codebase should automatically be applied to the containers, as your machine's filesystem has been "bound" to the containers' filesystem. This is one-way, so a container can't modify your filesystem. Adding new dependencies, however, will require a rebuild.
```
emmiegit:
```suggestion
Once everything has started, you can connect to `http://www.wikijump.localhost/`. Changes you make to the codebase should automatically be applied to the containers, as your machine's filesystem has been "bound" to the containers' filesystem (see `docker-compose.dev.yaml` below). This is one-way, so a container can't modify your filesystem. Adding new dependencies, however, will require a rebuild.
```

docs/development.md
```diff
@@ -64,58 +61,47 @@ $ pnpm install
...
 
-You can also bring containers up without the JS/TS development mode.
+It's useful to keep track of existing Docker images and containers, and destroy them when you no longer need them, so you don't waste space rebuilding the same image over and over. If you are using Docker Desktop, you can manage containers and images from the GUI. Otherwise, using the command line:
+
 ```
-$ pnpm dev serve
+$ docker container ls  # List containers
+$ docker rm [ID]       # Destroy the container with this ID
+$ docker images        # List images
+$ docker rmi [ID]      # Remove the image with this ID
```
emmiegit, line 92: New one you could add (just thought of), `docker system prune`, which goes and deletes _everything_ unused. It cleans up a lot of stuff and is normally only used every couple months kind of thing.

emmiegit (approved):

PR #1505, opened 2023-06-25
[WJ-1167] Site Membership Functionality

Yossipossi1:
[Link to Issue](https://scuttle.atlassian.net/browse/WJ-1167)

This PR intends to add basic functionality to allow for the adding, removing, and querying of users as site members. While not exhaustive, this should serve enough functionality for an MVP and serve as a good basis for future increments in functionality to this area. Endpoints are present and tested as well.


deepwell/migrations/20220906103252_deepwell.sql
```diff
@@ -101,6 +101,20 @@ ALTER TABLE site
...
     FOREIGN KEY (custom_domain) REFERENCES site_domain(domain);
 
+--
+-- Site Membership
+--
+
+CREATE TABLE site_member (
+    membership_id BIGSERIAL PRIMARY KEY,
+    user_id BIGSERIAL NOT NULL REFERENCES "user"(user_id),
+    site_id BIGSERIAL NOT NULL REFERENCES site(site_id),
```
emmiegit: `BIGSERIAL` is for primary keys, since it creates an autoincrement counter. Here `user_id` and `site_id` are not getting assigned new values, they're foreign keys to an existing table. These should be `BIGINT`.

deepwell/src/api.rs
```diff
@@ -166,6 +166,14 @@ fn build_routes(mut app: ApiServer) -> ApiServer {
         .get(site_custom_domain_retrieve);
     app.at("/site/fromDomain/:domain").get(site_get_from_domain);
 
+    // Site Membership
+    app.at("/site/members").get(membership_site_retrieve);
+    app.at("/site/members/get").get(membership_retrieve);
+    app.at("/site/members/add").put(membership_put);
+    app.at("/site/members/remove").delete(membership_delete);
```
emmiegit: The root here should be `/site/member`, singular.

deepwell/src/services/site_member/service.rs
```diff
@@ -0,0 +1,158 @@
...
+
+impl SiteMemberService {
+    /// Add a user to a site.
+    pub async fn add(
+        ctx: &ServiceContext<'_>,
+        SiteMembership { site_id, user_id }: SiteMembership,
+    ) -> Result<Option<SiteMemberModel>> {
+        let txn = ctx.transaction();
+        tide::log::info!(
+            "Adding membership of user with ID {user_id} to site ID {site_id}"
```
emmiegit:
```suggestion
            "Adding membership of user with ID {user_id} to site ID {site_id}",
```

deepwell/src/api.rs
emmiegit: These route/methods are not correct. The only reason `/get` is a separate thing here is because in HTTP you cannot submit a body with a GET request, so that is `PUT /foo/get` in order to emulate a "GET with a body" (PUT can have a body and is idempotent so that's why it's used here).

The others should not have separate routes. It should be `GET /site/member`, `PUT /site/member`, `DELETE /site/member`.

deepwell/src/services/site_member/structs.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub site_id: i64,
+    pub user_id: i64,
+}
+
+#[derive(Deserialize, Debug, Copy, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct SiteMembershipHistory {
+    pub user_id: Option<i64>,
+    pub site_id: Option<i64>,
+    pub current_members: Option<bool>,
```
emmiegit: Why is this plural?

deepwell/src/endpoints/site_member.rs
```diff
@@ -0,0 +1,87 @@
...
+    let txn = req.database().begin().await?;
+    let ctx = ServiceContext::new(&req, &txn);
+
+    let input: SiteMembership = req.body_json().await?;
+    let created = SiteMemberService::add(&ctx, input).await?;
+    txn.commit().await?;
+
+    match created {
+        Some(model) => build_membership_response(&model, StatusCode::Created),
+        None => Ok(Response::new(StatusCode::NoContent)),
```
emmiegit, line 46: :+1:

deepwell/src/services/site_member/service.rs
emmiegit:
```suggestion
            "Adding membership of user with ID {user_id} to site ID {site_id}",
```

deepwell/src/services/site_member/service.rs
```diff
@@ -0,0 +1,158 @@
...
+        );
+
+        // If the user is already a member of the target site, discontinue.
+        if Self::get_optional(ctx, SiteMembership { site_id, user_id })
+            .await?
+            .is_some()
```
emmiegit: Since this check is actually used, generally this is done using an `exists()` method. (See `PageService` for an example)

deepwell/src/services/site_member/service.rs
```diff
@@ -0,0 +1,158 @@
...
+            return Ok(None);
+        }
+
+        // TODO: Check for membership qualifications (e.g. bans).
+
+        // Insert new membership.
+        let model = site_member::ActiveModel {
+            user_id: Set(user_id),
+            site_id: Set(site_id),
+            deleted_at: Set(None),
```
emmiegit: Remove this, unnecessary.

deepwell/src/services/site_member/service.rs
```diff
@@ -0,0 +1,158 @@
...
+            ..Default::default()
+        };
+
+        let membership = model.insert(txn).await?;
+
```
emmiegit:
```suggestion
```

deepwell/src/services/site_member/service.rs
```diff
@@ -0,0 +1,158 @@
...
+            Some(member_model) => {
+                let mut model = member_model.into_active_model();
+                model.deleted_at = Set(Some(now()));
+                model.update(txn).await?
+            }
+
+            // If no membership is found, return BadRequest error.
+            None => {
+                tide::log::error!(
+                    "Could not remove user ID {user_id} from site ID {site_id} as they are not a member."
```
emmiegit:
```suggestion
                    "Could not remove user ID {user_id} from site ID {site_id} as they are not a member.",
```

deepwell/src/services/site_member/service.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /// Remove a user from a site.
+    pub async fn remove(
+        ctx: &ServiceContext<'_>,
+        SiteMembership { site_id, user_id }: SiteMembership,
+    ) -> Result<SiteMemberModel> {
+        let txn = ctx.transaction();
+        tide::log::info!(
+            "Removing the membership of user ID {user_id} from site ID {site_id}"
```
emmiegit:
```suggestion
            "Removing the membership of user ID {user_id} from site ID {site_id}",
```

deepwell/src/services/site_member/service.rs
```diff
@@ -0,0 +1,158 @@
...
+                model.deleted_at = Set(Some(now()));
+                model.update(txn).await?
+            }
+
+            // If no membership is found, return BadRequest error.
+            None => {
+                tide::log::error!(
+                    "Could not remove user ID {user_id} from site ID {site_id} as they are not a member."
+                );
+                return Err(Error::BadRequest);
```
emmiegit: Couldn't this return `None` for non-members? I was thinking add and removal operations could both be idempotent.

deepwell/src/services/site_member/service.rs
```diff
@@ -0,0 +1,158 @@
...
+            .filter(
+                Condition::all()
+                    .add(site_member::Column::SiteId.eq(site_id))
+                    .add(site_member::Column::UserId.eq(user_id))
+                    .add(site_member::Column::DeletedAt.is_null()),
+            )
+            .one(txn)
+            .await?;
+        Ok(model)
+    }
```
emmiegit, line 117: Add `exists()`, as noted above.

deepwell/src/services/site_member/service.rs
```diff
@@ -0,0 +1,158 @@
...
+            .filter(
+                Condition::all()
+                    .add(site_member::Column::SiteId.eq(site_id))
+                    .add(site_member::Column::DeletedAt.is_null()),
+            )
+            .order_by_asc(site_member::Column::MembershipId)
+            .all(txn)
+            .await?;
+
+        Ok(models)
```
emmiegit, line 138: This method needs a TODO stub for pagination. Sites like EN have tons of members, returning them all in one request doesn't make sense.

deepwell/src/services/site_member/service.rs
```diff
@@ -0,0 +1,158 @@
...
+                Condition::all()
+                    .add(site_member::Column::UserId.eq(user_id))
+                    .add(site_member::Column::DeletedAt.is_null()),
+            )
+            .order_by_asc(site_member::Column::MembershipId)
+            .all(txn)
+            .await?;
+
+        Ok(models)
+    }
```
emmiegit, line 159: For a `get_membership_history()` method, it would be `user_id: i64`, `site_id: Option<i64>`, and then returning all the membership events in chronological order. Probably should have a TODO for pagination?

deepwell/src/services/site_member/structs.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub start_id: i64,
+    pub limit: u64,
+}
```
emmiegit: Where is this struct used?

deepwell/src/api.rs
Yossipossi1: First time for everything 😅
will fix

emmiegit: No worries

emmiegit: Also site membership retrieve is its own thing and also takes a body, make that `PUT /site/member/site/get` or something

deepwell/src/services/site_member/structs.rs
Yossipossi1: It was used in the membership history methods before I removed them. I'm not sure why clippy didnt give off a warning for them.

deepwell/src/services/site_member/service.rs
Yossipossi1: Sure, I'll add a TODO

deepwell/migrations/20220906103252_deepwell.sql
Yossipossi1: Wasn't aware, I'll fix

deepwell/src/services/site_member/service.rs
Yossipossi1: Got it

deepwell/src/services/site_member/service.rs
Yossipossi1: Right, forgot that Option implements Default

deepwell/src/services/site_member/service.rs
Yossipossi1: Sure, I could change it to that. I was just modelling it after what VoteService did.

deepwell/src/services/site_member/service.rs
emmiegit: It's not `Default` (Rust trait), it's `DEFAULT` (database column setting).

All nullable columns have an implicit `DEFAULT NULL` basically.

deepwell/src/services/site_member/service.rs
emmiegit: Hmm. Yeah no I think you're right it should stay like this like for vote service.

deepwell/src/services/site_member/service.rs
Yossipossi1: Well yeah but also that `Set(None)` would automatically be included in `..Default::default` is what I meant

emmiegit: Actually no, the `Default` here is `Unset`, which is analogous to a SQL query where the value is simply excluded.

But yeah I get you.

emmiegit (approved):

PR #1595, opened 2023-09-09
[WJ-1180] Fix loading page data in +error.svelte

Zokhoi:
Turns out you need to do `$page` for loading values from store.


emmiegit (approved):

emmiegit:
Good catch! Such an easy-to-mistake detail

PR #1596, opened 2023-09-10
Frontend related features [1]

Zokhoi:
This tracks the Jira issues and features for frontend I am trying to tackle.

Done:
* [WJ-1179] Add svelte component for page view
* [WJ-1155] First basic UI interaction to edit a page
* Page creation, deletion, move
* Page history
* Sigma-esque page layout

[WJ-1179]: https://scuttle.atlassian.net/browse/WJ-1179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WJ-1155]: https://scuttle.atlassian.net/browse/WJ-1155?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deepwell/src/services/view/service.rs
```diff
@@ -100,12 +100,12 @@ impl ViewService {
...
         // Get page, revision, and text fields
-        let (category_slug, page_slug) = split_category(page_slug);
+        let (category_slug, page_clean_slug) = split_category(page_slug);
         let page_info = PageInfo {
-            page: cow!(page_slug),
+            page: cow!(page_clean_slug),
             category: cow_opt!(category_slug),
             site: cow!(&site.slug),
-            title: cow!(page_slug),
+            title: cow!(page_clean_slug),
```
emmiegit: "Clean" because the slug is normalized? Is this the cause of the improper normalization issue you reported in chat?

framerail/src/routes/+page.svelte
```diff
@@ -1,32 +1,6 @@
 <script lang="ts">
   export let data
+  import Page from "./[slug]/[...extra]/page.svelte"
```
emmiegit, line 3: :+1:

framerail/src/routes/[slug]/[...extra]/+server.ts
```diff
@@ -0,0 +1,17 @@
...
+export async function POST(event) {
+  let data = await event.request.formData()
+
+  let page = parseInt(data.get("page-id")?.toString() ?? "1")
+  let siteId = parseInt(data.get("site-id")?.toString() ?? "1")
+  let comments = data.get("comments")?.toString() ?? ""
+  let wikitext = data.get("wikitext")?.toString()
+  let title = data.get("title")?.toString()
+  let altTitle = data.get("alt-title")?.toString()
+  let tags = data.get("tags")?.toString().split(" ") ?? []
```
Comment (emmiegit) [inline framerail/src/routes/[slug]/[...extra]/+server.ts]: The "space splitting" convention is not something we should copy from Wikidot on the server side. It's fine for now as a _client-side_ means of specifying tags, but on the server it should just be dealing with a JSON list of strings. This way we get rid of ambiguities around tag separations and can eventually have a better tag picker UI.

deepwell/src/services/view/service.rs
Zokhoi: It was because the later lines receive the per category slug and not the full slug, and couldn't resolve pages with category properly, so I had to separate the variables for them to work. The variable name was intended to mean something like "stripped down slug".

framerail/src/routes/[slug]/[...extra]/+server.ts
Comment (Zokhoi) [reply framerail/src/routes/[slug]/[...extra]/+server.ts]: Yeah this is related to the tag input in the simple makeshift editor being just a single line of input. When the editor gets some more work this should be changed.

Comment (emmiegit) [reply framerail/src/routes/[slug]/[...extra]/+server.ts]: Oh you're right. I am too tired, I thought this was in deepwell. This is fine for now.

Zokhoi:
Will have to merge these changes to dev for now and continue working off of the new changes in dev

emmiegit:
For `framerail/src/lib/assets/wikijump-banner-solid.min.svg` should it be in `assets/`?

framerail/src/lib/server/page/history.ts
```diff
@@ -0,0 +1,29 @@
...
+    method: "PUT",
+    headers: {
+      "Content-Type": "application/json"
+    },
+    body: JSON.stringify({
+      siteId,
+      pageId: pageId,
+      revisionNumber: revisionNumber ?? -1,
+      revisionDirection: "before",
+      limit: limit ?? 20
```
emmiegit: I think the API already has a default. I think these kinds of decaults should live there, in one place?

framerail/src/routes/+error.svelte
```diff
@@ -1,18 +1,71 @@
...
+    goto(`/${$page.params.slug}`, {
+      noScroll: true
+    })
+  }
 </script>
 
 <h1>UNTRANSLATED:Svelte Error</h1>
 
-<p><textarea class="debug">{JSON.stringify($page)}</textarea></p>
+<p><textarea class="debug">{JSON.stringify($page, null, 2)}</textarea></p>
```
emmiegit, line 28: What is the `2` here? If it's a constant we should give it a name so we can tell what it means in the future.

framerail/src/routes/+error.svelte
```diff
@@ -1,18 +1,71 @@
...
   import { page } from "$app/stores"
+  import { goto } from "$app/navigation"
+
+  function cancelEdit() {
+    goto(`/${$page.params.slug}`, {
+      noScroll: true
+    })
+  }
+
+  async function saveEdit() {
```
emmiegit: I understand that site etc. information is in `error` because it's an error page. I don't think I understand why edit utilities are here. If you're making a page that doesn't exist?

Zokhoi:
Probably yes, or in `static/` or some other place. When I was including that I wasn't aware how to import from a path relative to project root or other custom directories. I think I could set up an alias for it in svelte or vite though.

emmiegit:
Sounds good, thanks. If it's not possible we should create a custom directory for frontend for assets. (Maybe symbolic links would work?)

framerail/src/lib/server/page/edit.ts
```diff
@@ -0,0 +1,42 @@
...
+  wikitext: string,
+  title: string,
+  altTitle: string,
+  tags: string[]
+): Promise<object> {
+  let endpoint = "/page"
+  if (!pageId) {
+    // Assume page creation
+    endpoint += "/create"
+  }
```
emmiegit: Interesting. Maybe this would be cleaner?
```ts
const endpoint = pageId ? '/page' : '/page/create'
```

framerail/src/lib/server/page/edit.ts
```diff
@@ -0,0 +1,42 @@
...
+  const response = await wellfetch(endpoint, {
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json"
+    },
+    body: JSON.stringify({
+      siteId,
+      page: pageId ?? slug,
+      slug,
```
emmiegit, line 23: Does create require page ID at all?

framerail/src/lib/server/page/index.ts
```diff
@@ -0,0 +1,6 @@
+import { pageDelete } from "$lib/server/page/delete"
+import { pageEdit } from "$lib/server/page/edit"
+import { pageHistory } from "$lib/server/page/history"
+import { pageMove } from "$lib/server/page/move"
+
+export { pageDelete, pageEdit, pageHistory, pageMove }
```
emmiegit: Minor point but I think you can do `export { name } from 'somewhere/else'`

framerail/src/routes/[slug]/[...extra]/+server.ts
```diff
@@ -0,0 +1,69 @@
+import * as page from "$lib/server/page"
+
+export async function POST(event) {
+  let data = await event.request.formData()
+  let slug = event.params.slug
+
+  let extra = event.params.extra
+    ?.toLowerCase()
+    .split("/")
+    .filter((flag) => flag.length)
```
Comment (emmiegit) [inline framerail/src/routes/[slug]/[...extra]/+server.ts:10]: There is already logic to parse wikidot-paths. Presently it's used as part of the `ViewService`, to do page parsing/presentation. It returns a list of flags with all the extra fields already parsed.

framerail/src/routes/[slug]/[...extra]/+server.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/** Delete page. */
+export async function DELETE(event) {
+  let data = await event.request.formData()
+  let slug = event.params.slug
+
+  let pageIdVal = data.get("page-id")?.toString()
+  let pageId = pageIdVal ? parseInt(pageIdVal) : null
+  let siteId = parseInt(data.get("site-id")?.toString() ?? "1")
```
Comment (emmiegit) [inline framerail/src/routes/[slug]/[...extra]/+server.ts:65]:
```suggestion
  let siteId = parseInt(data.get("site-id")?.toString() ?? "1") // TODO get site id
```

framerail/src/routes/[slug]/[...extra]/page.svelte
```diff
@@ -0,0 +1,339 @@
...
+      body: fdata
+    })
+    invalidateAll()
+  }
+
+  function navigateEdit() {
+    let options: string[] = []
+    if (data.options.noRender) options.push("norender")
+    options = options.map((opt) => `/${opt}`)
+    goto(`/${data.page.slug}${options.join("")}/edit`, {
```
Comment (emmiegit) [inline framerail/src/routes/[slug]/[...extra]/page.svelte:26]: Maybe we should expose a method of converting options back into a URL path.

framerail/src/routes/[slug]/[...extra]/page.svelte
```diff
@@ -0,0 +1,339 @@
...
+      noScroll: true
+    })
+  }
+
+  async function handleMove() {
+    let form = document.getElementById("page-move")
+    let fdata = new FormData(form)
+    let newSlug = fdata.get("new-slug")
+    if (!newSlug) {
+      moveInputNewSlugElem.style.outline = "1px solid red"
```
Comment (emmiegit) [inline framerail/src/routes/[slug]/[...extra]/page.svelte]:
```suggestion
      moveInputNewSlugElem.style.outline = "1px solid red" // TODO add proper CSS class/styling
```

framerail/src/routes/[slug]/[...extra]/page.svelte
```diff
@@ -0,0 +1,339 @@
...
+  UNTRANSLATED:Response <textarea class="debug">{JSON.stringify(data, null, 2)}</textarea>
+</p>
+
+<h2>{data.pageRevision.title}</h2>
+
+<hr />
+
+<div class="page-content">
+  {#if data.options?.noRender}
+    UNTRANSLATED: Content not shown.
```
Comment (emmiegit) [inline framerail/src/routes/[slug]/[...extra]/page.svelte:106]: This is fine for now. Eventually we should show a message and the page source underneath.

framerail/src/routes/[slug]/[...extra]/page.svelte
```diff
@@ -0,0 +1,339 @@
...
+    {/each}
+  </ul>
+</div>
+
+{#if data.options?.edit}
+  <form id="editor" class="editor" method="POST" on:submit|preventDefault={saveEdit}>
+    <input
+      name="title"
+      class="editor-title"
+      placeholder="title"
```
Comment (emmiegit) [inline framerail/src/routes/[slug]/[...extra]/page.svelte]:
```suggestion
      placeholder="UT:Title"
```

framerail/src/routes/[slug]/[...extra]/page.svelte
```diff
@@ -0,0 +1,339 @@
...
+      type="text"
+      value={data.pageRevision.title}
+    />
+    <input
+      name="alt-title"
+      class="editor-alt-title"
+      placeholder="alternative title"
```
Comment (emmiegit) [inline framerail/src/routes/[slug]/[...extra]/page.svelte]:
```suggestion
      placeholder="UT:Alternative Title"
```

emmiegit (approved):
Left a lot of comments, but mostly fine to go. You can change these in a follow-up PR to maintain momentum.

framerail/src/routes/+error.svelte
Zokhoi: This is the indentation space number for formatting.

emmiegit: Oh, right. For some reason I thought it was an API call.

framerail/src/lib/server/page/edit.ts
Zokhoi: This method is combined for page creation and edit, so I made it to use pageId first for page edit, and fall back to slug for page creation.

framerail/src/routes/[slug]/[...extra]/+server.ts
Comment (Zokhoi) [reply framerail/src/routes/[slug]/[...extra]/+server.ts:10]: Yes, but this bit is the web client making request to svelte server instead of to deepwell, and this flag is for the svelte server to tell if the web operation is edit, move or history.

framerail/src/routes/+error.svelte
Zokhoi: Right, this should really be page creation.

framerail/src/routes/[slug]/[...extra]/+server.ts
Comment (emmiegit) [reply framerail/src/routes/[slug]/[...extra]/+server.ts:10]: This is in a `+server.ts` file, couldn't it ask deepwell to process the route? Or other relevant data?

Comment (Zokhoi) [reply framerail/src/routes/[slug]/[...extra]/+server.ts:10]: It probably comes down to if framerail making requests to deepwell just for parsing these flags being economical or not.

Comment (emmiegit) [reply framerail/src/routes/[slug]/[...extra]/+server.ts:10]: Yeah. One reason I have `ViewService` is so you pass all the data once and have it process everything and return the response, instead of going back and forth.

The main reason I'm interested in having the logic there is so that the parsing of options is consistent. There are some edge cases that you have to be careful with.

framerail/src/routes/[slug]/[...extra]/+server.ts
Comment (Zokhoi) [reply framerail/src/routes/[slug]/[...extra]/+server.ts:65]: Currently it already has the site id passed from web client, this bit is just some leftovers.

Comment (emmiegit) [reply framerail/src/routes/[slug]/[...extra]/+server.ts:65]: So remove the `?? "1"`?

Comment (Zokhoi) [reply framerail/src/routes/[slug]/[...extra]/+server.ts:65]: If we can handle the NaN on deepwell side then probably
Or we can make it the same thing as pageId and give deepwell null instead

PR #1623, opened 2023-09-25
[WJ-1186] Multiple translation requests per call

emmiegit:
This replaces the `PUT /message/:locale/:message_key` call (which translates one message key) with `PUT /message/:locale/translate` (which translates many).

The request body is the following:
```json
{
    "message-key-1": {
        "variable-1": 100,
        "variable-2": "foo"
    },
    "message-key-2": {}
}
```

This will return a JSON object where each key corresponds to the translated message:
```json
{
    "message-key-1": "100 new foo objects have been successfully initialized.",
    "message-key-2": "Create new page"
}
```

### DIFF
--- a/deepwell/src/api.rs
+++ b/deepwell/src/api.rs
@@ -132,7 +132,7 @@ fn build_routes(mut app: ApiServer) -> ApiServer {
 
     // Localization
     app.at("/locale/:locale").get(locale_get);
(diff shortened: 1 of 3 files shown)


Zokhoi (approved):

emmiegit:
thanks @Zokhoi

PR #1625, opened 2023-09-28
Add locale folder and config file to cargo watch

Zokhoi:
So that we can hot reload Deepwell when locale files are changed.


emmiegit (approved):

Zokhoi:
Thanks @emmiegit

PR #1626, opened 2023-09-29
[WJ-1187] Automatic reload on configuration change

emmiegit:
`cargo watch` will rebuild `deepwell` when any watched files change. However this can take a bit even if no source files have changed, as it rebuilds the binary. I have implemented a feature, `-w` or `--watch-files`, which will restart the process (via self-`exec`) if the configuration TOML file or anything under `locales` changes. This option is CLI-only, and is not found in the configuration files. It only applies to a `local` feature flagged build only.

However, I cannot get this to work with Docker. None of the events are coming through like with a local instance.

For now, I have re-enabled the `cargo watch` approach for deploy, and changed to use the `RecommendedWatcher` (inotify on linux). (If you want to try to get this to work on Docker, keep in mind `PollWatcher` must be used, and you need to disable the `-w [path]` flags in `cargo watch`.) This works fine, but is slower. To use `-w` you can manually stop the container and then run deepwell locally on your host machine. This will restart on file change nearly instanteously, it's quite nice.

### DIFF
--- a/.github/workflows/deepwell.yaml
+++ b/.github/workflows/deepwell.yaml
@@ -45,8 +45,11 @@ jobs:
       - name: System Dependencies
         run: sudo apt update && sudo apt install libmagic-dev
 
-      - name: Build
-        run: cd deepwell && cargo build
+      - name: Build (local)
+        run: cd deepwell && cargo build --features local
+
+      - name: Build (deploy)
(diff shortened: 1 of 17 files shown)


Zokhoi (approved):

emmiegit:
thanks @Zokhoi

PR #1627, opened 2023-09-29
[WJ-1188] Remove void crate

emmiegit:
Replace with `std::convert::Infallible`, which is in the standard library. The `void` crate is now removed after a bump to the ftml version.


emmiegit:
Lint issue should be ignored, known bug in `cargo clippy` which has already been patched upstream (but not yet released).

Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1629, opened 2023-09-29
[WJ-1190] Replace surf with ureq

emmiegit:
We want to avoid dependencies on native TLS due to build/linking issues. Additionally surf is not as actively maintained as ureq.

[`ureq`](https://crates.io/crates/ureq) is not async but for our use I think it works fine. It has the features we need, uses rustls, and unlike reqwest it does not depend on tokio.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1630, opened 2023-09-29
Remove local / deploy features from deepwell

emmiegit:
Presently there is just the one feature, `watch`, and the `local` / `deploy` psuedo-features are not helpful. Removing for now, if we need them in the future we can add it then.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1631, opened 2023-09-30
Upgrade sea-orm and sea-query

emmiegit:
The long-procrastinated SeaQL crate upgrade PR. Supersedes #1614 and #1615.

PR #1632, opened 2023-09-30
Add serde rename to model script

emmiegit:
This is a follow-up to #1631 which adds an option to the `generate-models.sh` script to add the `#[serde(rename_all = "camelCase")]` automatically. (The ones for the enums still have to be added manually though.)


Zokhoi (approved):

emmiegit:
thanks @Zokhoi

PR #1633, opened 2023-09-30
[WJ-1161] [WJ-373] [WJ-770] [WJ-1128] Add message support

emmiegit:
* [WJ-1161](https://scuttle.atlassian.net/browse/WJ-1161) &mdash; Direct Messages epic
* [WJ-373](https://scuttle.atlassian.net/browse/WJ-343) &mdash; Allow archiving read PMs
* [WJ-770](https://scuttle.atlassian.net/browse/WJ-770) &mdash; Allow sending to multiple recipients (to, cc, bcc)
* [WJ-1128](https://scuttle.atlassian.net/browse/WJ-1128) &mdash; Allow sending messages to yourself (notes)

This PR adds initial support for messages, allowing exact later implementations of functionality to be added. The key design here is that there are three main objects:

* Message drafts (`message_draft`)
* Message records (`message_record`)
* Message copies (`message`)

In order to avoid duplicating message rendering, etc. logic, all message operations require a draft. Sending a message simply requires you pass a draft ID. Here the message is rendered and metadata is stored, then when you go to send it, the draft row is deleted and message table rows are created.

Then there is a distinction between _message records_ and _messages_. A message record stores all the bulk of a message, such as the body, subject, and sender. There is only one of these. Then there are one or more message copies. Each only has a particular user who received / can see the message, as well as data applicable to their particular copy.

For instance, let's say that Alice sends Bob and Charlie a message. This creates **1** `message_record`, but **3** `message`s (two inbox and one outbox).

Then each `message` can be mutated in user-specific ways. Maybe Bob tags his message, or trashes it, or whatever. This updates the flags in that row, but doesn't update the `message_record`, since the subject etc. stays the same regardless of actions he takes like that.

An important detail is that `message` IDs are internal-only, and not exposed outside DEEPWELL. Instead, `message_record` IDs are used to interact with messages. These are CUIDs such as `kjvpdq21zlllxxwabllkkxs5`. This way, because messages are private, we can avoid leaking information about message activity based on the change in message ID values (if assigned sequentially).

Then, these public CUID values are used to identify the message. This way you can link directly to a message you've sent by sharing the same CUID that you see in your outbox that your recipient got in their inbox.

That is, `(message_record_id, user_id)` can uniquely identify a `message` (primary key is an internal-only sequential ID). And since you're logged in, the second portion is already included, so Alice, Bob, and Charlie can all use `kjvpdq21zlllxxwabllkkxs5` to identify the message Alice sent, since it looks at the record ID, and not their individual `message` IDs. And if Eve tries to visit the URL for message `kjvpdq21zlllxxwabllkkxs5`, the system will see that she is not a recipient and deny her read access.

[WJ-1161]: https://scuttle.atlassian.net/browse/WJ-1161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WJ-373]: https://scuttle.atlassian.net/browse/WJ-373?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WJ-770]: https://scuttle.atlassian.net/browse/WJ-770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WJ-1128]: https://scuttle.atlassian.net/browse/WJ-1128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Zokhoi (approved):

emmiegit:
thanks @Zokhoi

PR #1641, opened 2023-10-06
[WJ-1193] Implement interactions

emmiegit:
This implement's bluesoul's notion of "interactions" in deepwell, as long as macros to autogenerate most of the relevant code. One minor issue is that no-data interaction still need to have a `()` passed in, but overall I think this macro system reduces a lot of boilerplate. Existing relations tables like `site_member` have been removed and replaced with an interaction implementation.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1!

PR #1643, opened 2023-10-06
Bump deepwell version

emmiegit:
It's been a while since the last version bump, and since we merged #1641, it seems like a good time to increase the version number.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1644, opened 2023-10-09
Replace REST with JSON-RPC

emmiegit:
Since we've finalized use of DEEPWELL as an internal API only, external-facing requirements no longer apply. We've already removed things like per-IP ratelimiting and now we've been facing a limitation of REST APIs, which is our desire to send more rich error data than one of a handful of relevant HTTP status codes. As such, we're shifting to a [JSON-RPC](https://www.jsonrpc.org/specification)-based internal API for DEEPWELL, which gets rid of the unnecessary HTTP complexities we don't need, and instead essentially has a remote method invocation system, with `serde_json`-backed serialization for input/output. `tide` (the web server) has been entirely removed, and so a lot of the lines changed are from log line diffs.

Likewise, I have taken the existing `services::Error` object and made it have a larger number of more specific error cases, since we are now sending those over the wire instead of bunching them into much more vague HTTP statuses.

I have also been able to use the transaction closure, which is wrapped outside of each method. Thus, boilerplate related to setting up the transaction and ensuring it is committed is now removed. For most methods, this means the wrapper code is very simple. Take the `message_draft_send` method:

```rs
pub async fn message_draft_send(
    ctx: &ServiceContext<'_>,
    params: Params<'static>,
) -> Result<MessageRecordModel> {
    let SendMessageDraft { message_draft_id } = params.parse()?;
    info!("Sending message draft with ID {message_draft_id}");
    MessageService::send(ctx, &message_draft_id).await
}
```

For local testing of DEEPWELL (for instance, using curl), the endpoint is now `POST http://localhost:2747/jsonrpc`. This always returns HTTP 200, error information is passed in the JSON result, in accordance with the JSON-RPC standard. I have modified the necessary code in framerail as well, and refactored it a bit. There, we are now using the [`json-rpc-2.0`](https://www.npmjs.com/package/json-rpc-2.0) library to act as a JSON-RPC client.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1651, opened 2023-10-15
Avoid false positive lint failures

emmiegit:
One issue we've had historically is the locales build failing when a new lint causes a conflict in ftml. This is unhelpful in terms of build failures, since it seems that there is an issue with the locale process when there is nothing of the sort.

Here, I add `--no-deps` to the `cargo clippy` invocations so that any dependencies are not linted. Instead, we will rely on the unconditional runs for our Rust crates on push to catch issues. (That is, our GitHub Workflow configurations cause all crates to be built on merge to `develop` or `prod` regardless of whether any such files were actually changed.)

### DIFF
--- a/.github/workflows/deepwell.yaml
+++ b/.github/workflows/deepwell.yaml
@@ -129,4 +129,4 @@ jobs:
         run: cd deepwell && cargo fmt --all -- --check
 
       - name: Clippy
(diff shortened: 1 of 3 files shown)


Zokhoi (approved):

emmiegit:
thanks @Zokhoi

PR #1669, opened 2023-10-27
[WJ-1191] Implement fallback locales

emmiegit:
This PR resolves [WJ-1191](https://scuttle.atlassian.net/browse/WJ-1191), which calls for two different, but related features:
* **Being able to pass in a list of locales when translating.** If a translation cannot be found for the first locale, the second is tried, then the third, etc., until a match is found.
* **Automatically trying generic forms of a locale.** One issue we presently have is that, if you request a translation in the `en_US` locale, you won't actually get anything, even though we have lots of `en` translations. This is because all those translations are for `en` specifically, so because there are no exact matches which include the region `US`, the search fails. Instead, we try increasingly generic forms of the locale until we find a match, before moving onto the next fallback locale. The order we use is taken from below:

locale fallback order

### Present locale, no fallback

Two fallback locales were specified, but neither was used.

```json
{
  "locales": [
    "zh_Hans",
    "en_CA",
    "fr"
  ],
  "messages": {
    "about": {},
    "navigated-to": {
      "path": "/start"
    }
  }
}
```
yielded
```json
{
  "about": "关于",
  "navigated-to": "导航至 ⁨/start⁩"
}
```

### Missing locale, generic locale

The first locale was not present, so a fallback was used. However this too was not present, so increasingly general variants of this locale (`en-US-valencia` → `en-US` → `en`) were checked until it found  a match.

```json
{
  "locales": [
    "hy",
    "en_US_valencia"
  ],
  "messages": {
    "about": {},
    "navigated-to": {
      "path": "/start"
    }
  }
}
```
yielded
```json
{
  "about": "About",
  "navigated-to": "Navigated to ⁨/start⁩"
}
```

### Mixed case

Each message is fetched separately, so which fallback locale is used depends on the particular message. This way partial translations of a language are supported. (For testing purposes the Vietnamese translation for login strings were removed.)

```json
{
  "locales": [
    "vi",
    "zh_Hans",
    "en"
  ],
  "messages": {
    "login": {},
    "login.toast": {},
    "logout": {},
    "logout.toast": {}
  }
}
```
yielded
```json
{
  "login": "登入",
  "login.toast": "您已登入。",
  "logout": "Đăng xuất",
  "logout.toast": "Bạn đã được đăng xuất."
}
```

### No locales specified
A new error, `NoLocaleSpecified` (error code 4105) has been introduced for this case.

[WJ-1191]: https://scuttle.atlassian.net/browse/WJ-1191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deepwell/src/endpoints/locale.rs
+++ b/deepwell/src/endpoints/locale.rs
@@ -32,9 +32,9 @@ pub struct LocaleOutput {
 }
 
 #[derive(Deserialize, Debug, Clone)]
-pub struct TranslateInput<'a> {
-    locale: &'a str,
-    messages: HashMap<String, MessageArguments<'a>>,
+pub struct TranslateInput {
+    locales: Vec<String>,
+    messages: HashMap<String, MessageArguments<'static>>,
 }
 
 type TranslateOutput = HashMap<String, String>;
@@ -58,18 +58,29 @@ pub async fn translate_strings(
     ctx: &ServiceContext<'_>,
     params: Params<'static>,
 ) -> Result<TranslateOutput> {
-    let TranslateInput {
-        locale: locale_str,
-        messages,
-    } = params.parse()?;
+    let TranslateInput { locales, messages } = params.parse()?;
+
+    if locales.is_empty() {
+        error!("No locales specified in translate call");
+        return Err(ServiceError::BadRequest);
+    }
 
     info!(
-        "Translating {} message keys in locale {locale_str}",
+        "Translating {} message keys in locale {} (or {} fallbacks)",
         messages.len(),
+        &locales[0],
+        locales.len() - 1,
     );
 
-    let locale = LanguageIdentifier::from_bytes(locale_str.as_bytes())?;
     let mut output: TranslateOutput = HashMap::new();
+    let locales = {
+        let mut langids = Vec::new();
+        for locale in locales {
+            let langid = LanguageIdentifier::from_bytes(locale.as_bytes())?;
+            langids.push(langid);
+        }
+        langids
+    };
 
     for (message_key, arguments_raw) in messages {
         info!(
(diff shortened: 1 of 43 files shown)


Zokhoi:
How is the locale list ordered if some of the generic locales has already been specified in the list?
e.g.
```json
{
  "locales": [
    "en-US",
    "zh-hans",
    "en"
  ],
  ...
}
```
Does it still try to insert `en` before `zh-hans`?

deepwell/src/endpoints/locale.rs
```diff
@@ -58,18 +58,29 @@ pub async fn translate_strings(
...
-    let locale = LanguageIdentifier::from_bytes(locale_str.as_bytes())?;
     let mut output: TranslateOutput = HashMap::new();
+    let locales = {
+        let mut langids = Vec::new();
+        for locale in locales {
+            let langid = LanguageIdentifier::from_bytes(locale.as_bytes())?;
+            langids.push(langid);
+        }
+        langids
+    };
```
emmiegit, line 83: Forgot this was used here, I should move the helper function below out somewhere public so it can get imported and used here too.

emmiegit:
> How is the locale list ordered if some of the generic locales has already been specified in the list? e.g.
> 
> ```json
> {
>   "locales": [
>     "en-US",
>     "zh-hans",
>     "en"
>   ],
>   ...
> }
> ```
> 
> Does it still try to insert `en` before `zh-hans`?

Yes, for each locale it tries all the generic forms before going to the next one. So it would check `en` before _and_ after `zh-hans`.

Zokhoi (approved):

emmiegit:
thanks @Zokhoi

PR #1668, opened 2023-10-24
[WJ-1145] Add initial redis support

emmiegit:
This PR adds Redis support to DEEPWELL, secrets retrieval, a local Redis container (similar to minio) for development, and the GitHub workflows for redis and minio.

I ensure DEEPWELL builds with Redis, and the `ping` method now checks for the Redis connection in addition to the existing Postgres connection check.

PR #1671, opened 2023-10-30
[WJ-1176] Job queue

emmiegit:
Ground work for a persistent job queue was set up in #1668.

In this PR, I use the added `rsmq_async` crate to persist jobs to the queue, and receive jobs from the queue in a job worker (configurable quantity), before deleting them post-success, to enable failure retries. The previous job queue implementation, an in-memory channel, was removed. This simplifies `ServerState` setup a bit.

Additionally, to implement recurring jobs such as the periodic pruning work, this system permits jobs to submit a "follow-up job", which is added to the queue as part of the job's work. Combined with a delay, this allows things such as "run this operation every six hours".

I've added configuration fields for all these various durations and values.

I also change the concept of "directly fetching pages" to not require a site ID, and have an option for more clear fetching of deleted pages or not. This also updates the `OutdateService`, which had a bug in it regarding use of an incorrect `site_id` on pages.

NOTE: There is a known issue where the value for "one day" (86400 seconds) gets multiplied by 1000 and overflows rsmq's limit. I will investigate further and file a bug in the upstream, but for now this PR is ready.


Yossipossi1 (approved):

Zokhoi (approved):

emmiegit:
thanks @Zokhoi @Yossipossi1

PR #1678, opened 2023-10-30
Avoid false positive lint failures (2)

emmiegit:
Follow-up to #1651

Despite the `--no-deps`, our locale workflow is picking up errors not from its own code being linted but because clippy is annoying about deepwell's use of star imports in prelude and the like.


Zokhoi (approved):

emmiegit:
thanks @Zokhoi

Yossipossi1 (approved):

emmiegit:
thanks yossi

PR #1679, opened 2023-11-03
A lot of frontend stuff [2]

Zokhoi:
This PR contains the frontend related stuff that I have been working on since the last one, including:
* Fixing all the JSON-RPC stuff, and the remaining camelCase to snake_case
* [WJ-1175](https://scuttle.atlassian.net/browse/WJ-1175) - i18n related
* Login and logout
* Viewing page source, page revision, and basic user profile
* Error popup

[WJ-1175]: https://scuttle.atlassian.net/browse/WJ-1175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Zokhoi:
...A lot of checks are failing with repo not found?

emmiegit:
> ...A lot of checks are failing with repo not found?

GitHub seems to have recovered from its outage, I'll manually restart all your CI jobs.

deepwell/src/endpoints/locale.rs
```diff
@@ -93,7 +93,7 @@ pub async fn translate_strings(
             ctx.localization()
                 .translate(&locales, &message_key, &arguments)?;
 
-        output.insert(message_key, translation.to_string());
+        output.insert(message_key, Some(translation.to_string()));
```
emmiegit: What does null here represent? If a translation key cannot be found, it returns an error, not `None`. Do you think we should change this behavior?

deepwell/src/services/view/service.rs
```diff
@@ -288,6 +289,90 @@ impl ViewService {
...
+
+        let viewer = Viewer {
+            site,
+            redirect_site,
+            user_session,
+        };
+
+        // Helper structure to designate which variant of GetUserViewOutput to return.
+        #[derive(Debug)]
+        enum UserStatus {
```
emmiegit: I used a custom enum `PageStatus` above to account for the different cases unique to pages, page exists, page doesn't exist, page exists but no permissions. I don't think users would have any cases other than it exists or not, so we could probably leave this as `Option<UserModel>` instead.

Though this made me think of a question, what parts of a deleted user would platform staff be able to see? Would it be through this route?

deepwell/src/services/view/service.rs
```diff
@@ -288,6 +289,90 @@ impl ViewService {
...
+                ctx,
+                user_ref,
+            )
+            .await?
+            {
+                // This user exists, return its data directly.
+                Some(user) => UserStatus::Found{ user },
+                // The user is missing, fetch the "missing user" data.
+                None => UserStatus::Missing,
+            },
```
emmiegit:
```suggestion
            },
            // For users visiting their own user info page
```

framerail/src/assets/README.md
```diff
@@ -0,0 +1,7 @@
+## Assets
+
+This directory is not meant to hold any actual asset files.
+
+For docker containers, this will be the mounting point for the global `assets/` directory, and anything that was previously in this directory will be obscured.
```
emmiegit:
```suggestion
For Docker containers, this will be the mount point for the global `assets/` directory, and anything that was previously in this directory will be obscured.
```

framerail/src/lib/defaults.ts
```diff
@@ -0,0 +1,23 @@
...
+    "footer-powered-by": {},
+    "terms": {},
+    "privacy": {},
+    "docs": {},
+    "security": {},
+
+    // Page actions
+    "save": {},
+    "cancel": {}
+  },
```
emmiegit, line 10: What is this section for?

framerail/src/lib/server/deepwell/page.ts
```diff
@@ -63,7 +67,23 @@ export async function pageMove(
...
+  revisionNumber: Optional<Number>
+): Promise<object> {
+  return client.request("page_revision_get", {
+    site_id: siteId,
+    page_id: pageId,
+    revision_number: revisionNumber ?? defaults.page.history.revisionNumber,
+    details: {
+      compiled_html: true,
+      wikitext: true
+    }
```
emmiegit, line 89: Are these always needed? They're toggleable here since they can increase the response size by a fair bit.

framerail/src/lib/server/deepwell/translate.ts
```diff
@@ -0,0 +1,16 @@
+import { client } from "$lib/server/deepwell/index.ts"
+
+export interface PageRoute {
```
emmiegit: Why is this defined here?

framerail/src/lib/server/deepwell/page.ts
```diff
@@ -63,7 +67,23 @@ export async function pageMove(
...
-    user_id: 1, // TODO: identify user session and pass the user to the API request
+    user_id: userId,
     revision_comments: revisionComments
   })
 }
+
+export async function pageRevision(
+  siteId: number,
+  pageId: Optional<number>,
+  revisionNumber: Optional<Number>
```
emmiegit:
```suggestion
  revisionNumber: Optional<number>
```

framerail/src/lib/server/load/login.ts
```diff
@@ -0,0 +1,43 @@
...
+  const url = new URL(request.url)
+  const domain = url.hostname
+  const sessionToken = cookies.get("wikijump_token")
+  const language = request.headers.get("Accept-Language")
+  let locales = parse(language).map((lang) =>
+    lang.region ? `${lang.code}-${lang.region}` : lang.code
+  )
+
+  let viewData: Record<string, any> = {
+    isLoggedIn: !!sessionToken
```
emmiegit: I believe `Boolean(sessionToken)` is equivalent, and is also more readable

framerail/src/lib/server/load/login.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  if (!locales.includes(defaults.fallbackLocale)) locales.push(defaults.fallbackLocale)
```
emmiegit, line 17: Hmm, so the idea is we should always request English as a translation so it's missing? I guess this opens a discussion on how we want to handle missing translations, so they can be identified and filled with minimal disruption to end users.

We can do something on the backend side which can use a virtual "missing locale", or expose a special warning or message, or something.

framerail/src/lib/server/load/login.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  let translateKeys: Record<string, Record<string, string | number> | {}> = {
```
emmiegit:
```suggestion
  const translateKeys: Record<string, Record<string, string | number> | {}> = {
```

framerail/src/lib/server/load/logout.ts
```diff
@@ -0,0 +1,38 @@
...
+    lang.region ? `${lang.code}-${lang.region}` : lang.code
+  )
+
+  let viewData: Record<string, any> = {
+    isLoggedIn: !!sessionToken
+  }
+
+  if (!locales.includes(defaults.fallbackLocale)) locales.push(defaults.fallbackLocale)
+
+  let translateKeys: Record<string, Record<string, string | number> | {}> = {
```
emmiegit: Probably worth making `TranslationKeys` its own type alias. Is there a server `types.ts` file we can use?

framerail/src/lib/server/load/page.ts
```diff
@@ -16,16 +19,20 @@ export async function loadPage(
...
-
-  // TODO set up svelte i18n, see WJ-1175
-  //
-  // TODO also set up deepwell fluent so that fallback
-  //      languages are used, i.e. if I do en-GB it falls back to
-  //      en generic
-  const locale = "en"
+  let locales = parse(language).map((lang) =>
+    lang.region ? `${lang.code}-${lang.region}` : lang.code
+  )
```
emmiegit: I looked at the `accept-language-parser` library, and I think there are a few things here:
* We don't preserve the `script` field from the response. Ideally we could just paste the whole locale string in pass that in as-is and let deepwell parse it for us, but absent that we can combine `lang.code`, `lang.script`, `lang.region` with dashes (excluding empty items). We could also look for a different parser library which doesn't parse out language codes for us and just gives us the order and quality.
* Which brings me to the quality field, which we presently ignore. While the first item is always considered to have quality `1.0`, later items are not necessarily in order. We should sort by quality to ensure the actual user priority listing is being honored.

Looking on crates I see [`accept-language`](https://docs.rs/accept-language/3.1.0/accept_language/fn.parse.html) which doesn't parse locales, just gives strings, and the parse function notes that it's equivalent to `window.navigator.languages` on supported browsers. So maybe we can pass in `window.navigator.languages || request.headers.get('Accept-Language')` and let the deepwell parse out the list of locales from that?

framerail/src/lib/server/load/user.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  if (errorStatus === null && username && viewData.user.slug !== username) {
+    throw redirect(308, `/-/user/${viewData.user.slug}`)
+  }
+
+  if (errorStatus !== null) {
+    translateKeys = Object.assign(translateKeys, {
+      "user-not-exist": {}
+    })
```
emmiegit: I think spread notation is preferred now? `{ ...translateKeys, 'new-key-1': {}, 'new-key-2': {} }`

framerail/src/routes/[slug]/[...extra]/+server.ts
```diff
@@ -18,55 +25,88 @@ export async function POST(event) {
...
-      title,
-      altTitle,
-      tags
+  try {
+    if (extra.includes("edit")) {
+      /** Edit or create page. */
+      let comments = data.get("comments")?.toString() ?? ""
+      let wikitext = data.get("wikitext")?.toString()
+      let title = data.get("title")?.toString()
+      let altTitle = data.get("alt-title")?.toString()
```
Comment (emmiegit) [inline framerail/src/routes/[slug]/[...extra]/+server.ts:34]: Does this also need `?? ""`?

framerail/src/routes/[x+2d]/login/+server.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  let nameOrEmail = data.get("name-or-email")?.toString()
+  let password = data.get("password")?.toString()
+
+  try {
+    let res = await authLogin(nameOrEmail, password, ipAddress, userAgent)
+
+    if (res.session_token) {
+      event.cookies.set("wikijump_token", res.session_token, {
+        path: "/"
```
Comment (emmiegit) [inline framerail/src/routes/[x+2d]/login/+server.ts]: We should set some other flags here:
* `httpOnly: true` - client-side JS has no need to read the cookie itself, just have it sent with any requests. greatly mitigates the risk of XSS vulnerabilities
* `sameSite: 'lax'` - based on [this description](https://owasp.org/www-community/SameSite), `Lax` seems like the proper choice
* `secure: true` - only sent via HTTPS. Wikijump will only support HTTPS so this won't cause issues, and resists ability to sniff and steal session cookies
* `maxAge: (..)` - I guess I should modify the RPC call for session creation to return the expiration date so it can be passed in here. for now add a TODO?

framerail/src/routes/[x+2d]/logout/+server.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export async function DELETE(event) {
+  let sessionToken = event.cookies.get("wikijump_token")
+
+  try {
+    let res = await authLogout(sessionToken)
+
+    event.cookies.set("wikijump_token", "", {
+      path: "/",
+      maxAge: 0
```
Comment (emmiegit) [inline framerail/src/routes/[x+2d]/logout/+server.ts]: I think you can just do `event.cookies.delete()` instead

framerail/src/routes/[x+2d]/user/[slug]/page.svelte
```diff
@@ -0,0 +1,64 @@
+<script lang="ts">
+  import { page } from "$app/stores"
+</script>
+
+<h1>UNTRANSLATED: Loaded user profile</h1>
+
+<textarea class="debug">{JSON.stringify($page, null, 2)}</textarea>
+
+<div class="user-info" data-id={$page.data.user.user_id}>
```
Comment (emmiegit) [inline framerail/src/routes/[x+2d]/user/[slug]/page.svelte:9]: `data-id` is a good addition :+1:

framerail/src/routes/[x+2d]/user/[slug]/page.svelte
```diff
@@ -0,0 +1,64 @@
...
+      {$page.data.user.name}
+    </h2>
+  {/if}
+
+  {#if $page.data.user.real_name}
+    <div class="user-attribute real-name">
+      {$page.data.internationalization?.["user-profile-info.name"]}
+      {$page.data.user.real_name}
+    </div>
+  {/if}
```
Comment (emmiegit) [inline framerail/src/routes/[x+2d]/user/[slug]/page.svelte:23]: Not a blocker, but just an idea. We could have an "advanced mode" setting like in Discord's UI to expose IDs in the UI.

framerail/src/routes/[x+2d]/user/[slug]/page.svelte
```diff
@@ -0,0 +1,64 @@
...
+  {#if $page.data.user.name}
+    <h2 class="user-attribute name">
+      {$page.data.user.name}
+    </h2>
+  {/if}
+
+  {#if $page.data.user.real_name}
+    <div class="user-attribute real-name">
+      {$page.data.internationalization?.["user-profile-info.name"]}
+      {$page.data.user.real_name}
```
Comment (emmiegit) [inline framerail/src/routes/[x+2d]/user/[slug]/page.svelte]: General note/nitpick, I think it would be helpful for CSS themers to wrap loose text in spans

emmiegit (commented):
Left some comments, thanks for all this work! Exciting to see localization support progress further

framerail/src/lib/defaults.ts
Zokhoi: These texts are needed for the edit box actions in both the loaded page and error page

deepwell/src/endpoints/locale.rs
Zokhoi: This was left over from when the fallback was not implemented and I just made my own endpoint that for looped the locales. However in the current fallback version, the key not found error obscures all the other existing keys that are requested in the same call.

framerail/src/routes/[x+2d]/logout/+server.ts
Comment (Zokhoi) [reply framerail/src/routes/[x+2d]/logout/+server.ts]: I think I tried that, but for whatever reason it didn't actually delete the cookie, so I resorted to maxAge 0 instead.

framerail/src/routes/[slug]/[...extra]/+server.ts
Comment (Zokhoi) [reply framerail/src/routes/[slug]/[...extra]/+server.ts:34]: I didn't look at how deepwell implements page edit, but I am thinking more along the lines of "if the request didn't specify a new value for the field, leave it undefined and we can choose to carry it on from the previous revision", and comment is the thing that should not be carried over so it has a empty string default.

Comment (emmiegit) [reply framerail/src/routes/[slug]/[...extra]/+server.ts:34]: Oh no, you're absolutely right, this should be left as-is.

The implementation is as you says, aside from the revision comments (which always must be populated) any body field can be omitted to say "leave it as the previous value". So if it's not specified it shouldn't be sent,

framerail/src/routes/[x+2d]/logout/+server.ts
Comment (emmiegit) [reply framerail/src/routes/[x+2d]/logout/+server.ts]: Weird. Maybe leave a comment noting that.

framerail/src/lib/server/deepwell/page.ts
Zokhoi: This is called when the user tries to view a revision's html or source by clicking on the button in revision list, maybe I can set it to toggle only the respective option on.

Zokhoi:
Alright I think I addressed most of these, the remaining open comments are mostly discussions to be had and such

framerail/src/lib/server/deepwell/page.ts
```diff
@@ -63,7 +67,25 @@ export async function pageMove(
...
     revision_comments: revisionComments
   })
 }
+
+export async function pageRevision(
+  siteId: number,
+  pageId: Optional<number>,
+  revisionNumber: Optional<number>,
+  compiledHtml: boolean,
+  wikitext: boolean
```
emmiegit: If these are optional, then isn't `compiledHtml?: boolean` and `wikitext?: boolean` necessary?

framerail/src/lib/server/deepwell/translate.ts
```diff
@@ -0,0 +1,9 @@
+import { client } from "$lib/server/deepwell/index.ts"
+import type { TranslateKeys } from "$lib/types"
+
+export async function translate(locales: string[], keys: TranslateKeys): Promise<object> {
```
emmiegit: Could add return type, `TranslatedKeys`, which I believe is `Record<string, string>` (or `Record<string, Optional<string>>`, do we want to change to returning null on no translation found?)

framerail/src/lib/server/load/login.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export async function loadLoginPage(request, cookies) {
+  // Set up parameters
+  const url = new URL(request.url)
+  const domain = url.hostname
+  const sessionToken = cookies.get("wikijump_token")
+  const language = request.headers.get("Accept-Language")
+  let locales = parse(language)
+    .sort((a, b) => a.quality - b.quality)
+    .map((lang) => (lang.region ? `${lang.code}-${lang.region}` : lang.code))
```
emmiegit: I appreciate the sorting, though this version of locale reconstruction still discards the `script` value.

Basically the implementation is:
```typescript
function rebuildLocale(lang: object): string {
    const parts = [lang.code]
    if (lang.script) parts.push(lang.script)
    if (lang.region) parts.push(lang.region)
    return parts.join('-')
}

[previous code]
.map(rebuildLocale)
```

framerail/src/lib/server/load/logout.ts
```diff
@@ -0,0 +1,39 @@
+import defaults from "$lib/defaults"
+import { translate } from "$lib/server/deepwell/translate"
+import type { TranslateKeys } from "$lib/types"
+import { parse } from "accept-language-parser"
+
+export async function loadLogoutPage(request, cookies) {
+  // Set up parameters
+  const url = new URL(request.url)
+  const domain = url.hostname
+  const sessionToken = cookies.get("wikijump_token")
```
emmiegit, line 10: Maybe worth having a `$lib/constants` for constant strings like `wikijump_token`.

framerail/src/lib/server/load/logout.ts
```diff
@@ -0,0 +1,39 @@
...
+  const language = request.headers.get("Accept-Language")
+  let locales = parse(language)
+    .sort((a, b) => a.quality - b.quality)
+    .map((lang) => (lang.region ? `${lang.code}-${lang.region}` : lang.code))
```
emmiegit: Ditto. (Helper function would make fixing this easy)

framerail/src/lib/server/load/user.ts
```diff
@@ -0,0 +1,71 @@
...
+import { parse } from "accept-language-parser"
+
+export async function loadUser(username?: string, request, cookies) {
+  const url = new URL(request.url)
+  const domain = url.hostname
+  const sessionToken = cookies.get("wikijump_token")
+  const language = request.headers.get("Accept-Language")
+  let locales = parse(language)
+    .sort((a, b) => a.quality - b.quality)
+    .map((lang) => (lang.region ? `${lang.code}-${lang.region}` : lang.code))
```
emmiegit: Ditto. Maybe this whole expression should be a helper function, like `parseAcceptLanguageHeader()` or something.

framerail/src/lib/server/load/login.ts
```diff
@@ -0,0 +1,44 @@
...
+import type { TranslateKeys } from "$lib/types"
+import { parse } from "accept-language-parser"
+
+export async function loadLoginPage(request, cookies) {
+  // Set up parameters
+  const url = new URL(request.url)
+  const domain = url.hostname
+  const sessionToken = cookies.get("wikijump_token")
+  const language = request.headers.get("Accept-Language")
+  let locales = parse(language)
```
emmiegit: Something I forgot to mention, but if the user has account-level preferences for locales, that would override whatever their local computer says (maybe they're logged in on a machine they don't typically use). Not sure where this step would go, since it would need to be fetched from the database.

framerail/src/routes/[x+2d]/logout/+server.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  try {
+    let res = await authLogout(sessionToken)
+
+    event.cookies.delete("wikijump_token", {
+      path: "/",
+      httpOnly: true,
+      secure: true,
+      sameSite: "lax"
+    })
```
Comment (emmiegit) [inline framerail/src/routes/[x+2d]/logout/+server.ts:14]: Does this work? You mentioned it having issues.

emmiegit (approved):
Looks good, just a couple comments. The only blocker is the local reconstruction part.

framerail/src/routes/[x+2d]/logout/+server.ts
Comment (Zokhoi) [reply framerail/src/routes/[x+2d]/logout/+server.ts:14]: It works now, last time I tried I was debugging cookie issues and I think that may have affected something.

framerail/src/lib/server/deepwell/translate.ts
Zokhoi: I would prefer returning null, since right now the no translation found error obscures all other strings for the keys that do exist, and on a page with a lot of translated strings it becomes a whole bunch of undefineds.

framerail/src/lib/server/load/login.ts
Zokhoi: I looked at the source code of `accept-language-parser`, seems that they didn't mention in the readme but they do parse script, and they already apply sorting in descending quality value. I will still keep the sorting in the code though.

framerail/src/lib/server/load/login.ts
Zokhoi: I would think there should be a store containing the user details that is only populated once for things like user profile in the top right corner, and we can pull the user locale preference from the store.

PR #1680, opened 2023-11-03
Avoid false positive lint failures (3) 

emmiegit:
Follow-up to #1678.

I thought ftml would be fine since it wasn't reporting issues, and didn't have the same kinds of imports we have in our other crates, but I was wrong. This should finally quiet our recent slate of unhelpful build failures.

Example failed build: https://github.com/scpwiki/wikijump/actions/runs/6713086734/job/18243975202


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1681, opened 2023-11-03
Clarify use of ID 0

emmiegit:
Due to the NULL / 0 conflation some languages make too easy, to avoid this, we do not permit 0 to be allocated as a valid ID for objects. This is clarified in the `docs/limits.md` document, as well as making some other improvements to that file.


Zokhoi (approved):

emmiegit:
thanks @Zokhoi

PR #1683, opened 2023-11-04
Update minio container health check system

emmiegit:
As noted in https://github.com/minio/minio/issues/18373, `curl` is no longer shipped in the `minio` container, which means our existing health check system is now broken. Using a recommendation in the linked issue, we can update the health check, and place that into a separate script for future update convenience.


Zokhoi (approved):

emmiegit:
thanks @Zokhoi

PR #1688, opened 2023-11-06
Remove nightly from Rust lint checks

emmiegit:
The regression with some builds, as well as issues we've had such as the aggressive unused imports lint resulted from _nightly_ clippy. This also resulted in nightly builds, which is why we've had strange build regressions. Since clippy has been landed in stable for some time, we can just switch to that.

We could also merge the rustfmt and clippy steps into the main build group, but I'm not doing that for two reasons:
1. Having a separate group is helpful for seeing at a glance what kind of build failures you have. If there's only one, you can't tell if the failure is "code isn't formatted" vs "it won't compile".
2. There is the question of whether format checks should be before or after compilation. If they're before, then something like a missing comma can mean your build doesn't even try to compile, if it's after, then you might have to wait for the entire compilation to finish before seeing said comma is missing and you have to push a fix.

### DIFF
--- a/.github/workflows/deepwell.yaml
+++ b/.github/workflows/deepwell.yaml
@@ -112,6 +112,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
+          toolchain: stable
           override: true
           components: rustfmt, clippy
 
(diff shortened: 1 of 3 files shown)


Zokhoi (approved):

emmiegit:
thanks @Zokhoi

PR #1691, opened 2023-11-07
Return null for frontend when translation missing

emmiegit:
For the frontend, we want to return `null` for any message keys whose translations could not be found, rather than returning a request-level error. We do this by adding a new method which captures that variety of error and returns `Option<_>` accordingly.


Zokhoi (approved):

emmiegit:
thanks @Zokhoi

PR #1692, opened 2023-11-09
[WJ-1200] Add "site users"

emmiegit:
As described in [WJ-1200](https://scuttle.atlassian.net/browse/WJ-1200), the "site user" is a special type of user which is created only when a site is, and represents it in totality. All such users have the name `site:[slug]`, for instance `site:scp-sandbox-3`. This allows us to do things with this virtual user such as sending staff messages from a standard account, forwarding messages received to different assigned staff members, and having one standard user to represent the site itself for matters such as "page whose attribution is staff".

This implements this new user type, adds logic to create it when a site is created, and edits it when a site is edited. Because colons (`:`) are not normally allowed in user slugs, there is a special exception for this user type.

[WJ-1200]: https://scuttle.atlassian.net/browse/WJ-1200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1702, opened 2023-11-13
Use user locale preferences in view, allow multiple locales

emmiegit:
One design goal is to allow multi-lingual users to specify multiple locales (in a specific order) that they want to see used if a previous locale is not available, so this change both uses the locales specified by the user in the localization order before using the browser `Accept-Language` setting, and changes the database so that the `user.locales` column is `TEXT[]` instead of `TEXT`.

This also requires that messages pass in a locale they want the message to be rendered as, since with multiple locales per user, we cannot assume any particular one in message drafting.

This does not address the case where the user is not logged in _and_ there is no `Accept-Language` header, which will need to be fixed in another PR.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1703, opened 2023-11-14
[WJ-1204] Rename "interaction" to "relation"

emmiegit:
After thinking about it for some time, `relation` is a better, more descriptive name than `interaction`, since it denotes an ongoing relationship between database objects rather than a one-off event. This PR goes through and changes all instances of "interaction" in DEEPWELL to "relation", and adds a document explaining the basic idea of the concept for common reference.

**Reviewer note:** This diff is a bit unhelpful, since in places it shows files as being deleted/created rather than moved. Keep this in mind when you see added files adding lots of "new" code.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1704, opened 2023-11-14
Delete legacy PHP files for items now implemented in DEEPWELL

emmiegit:
We keep items in `/legacy` to reflect things that are still TODO in Wikijump, and since these items have corresponding backend code concepts, we can delete them.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1705, opened 2023-11-14
Fix locale for guest with no Accept-Language header

Zokhoi:
This addresses the case for no user locale and no Accept-Language header, which happen for example with docker curl polling the web container for health status.

### DIFF
--- a/framerail/src/lib/server/load/page.ts
+++ b/framerail/src/lib/server/load/page.ts
@@ -25,7 +25,12 @@ export async function loadPage(
 
   // Request data from backend
   // Includes fallback locale in case there is no Accept-Language header
-  const response = await pageView(domain, [...locales, defaults.fallbackLocale], route, sessionToken)
+  const response = await pageView(
+    domain,
+    [...locales, defaults.fallbackLocale],
+    route,
+    sessionToken
+  )
 
   if (response.data?.site?.locale && !locales.includes(response.data.site.locale)) {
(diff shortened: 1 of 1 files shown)


Yossipossi1 (approved):

Zokhoi:
Thanks @Yossipossi1

framerail/src/lib/server/load/page.ts
```diff
@@ -21,11 +21,17 @@ export async function loadPage(
...
-  // Request data from backend
-  const response = await pageView(domain, locales, route, sessionToken)
-
   // TODO insert user preference at the beginning of the list
 
+  // Request data from backend
+  // Includes fallback locale in case there is no Accept-Language header
+  const response = await pageView(
+    domain,
+    [...locales, defaults.fallbackLocale],
```
emmiegit, line 30: This works, but I can't help but feel like it's strange to always have a fallback to English there even if it was never requested by the user. Would `if empty then [default_locale]` make sense to you?

Zokhoi: This is done in a similar fashion to fallbackLocale always being present in the later translation request when the entire locale list is constructed by (user preference ->) browser header -> site settings -> fallback locale. Here we only have the browser header and fallback locale.

emmiegit: Hmm, I guess that makes sense.

PR #1706, opened 2023-11-14
Remove example user locale

Zokhoi:
Follow up for #1702: quick fix for the example user in seeder having a locale while being a "system" user, which fails the Deepwell validation.


emmiegit (approved):
My bad 😅

PR #1710, opened 2023-11-17
Replace async-std with tokio

emmiegit:
This was one part of the RPC transition that was left behind, so I'm doing it here. With this, we don't directly depend on `async-std` anymore and instead use `tokio` for this and our main dependencies.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1745, opened 2023-12-18
Use newly-hosted rust-otp crate

emmiegit:
The otp crate we use is functional but unfortunately is not hosted on crates.io. See issue https://github.com/TimDumol/rust-otp/issues/4. However now there is [WesleyBatista/rust-otp](https://github.com/WesleyBatista/rust-otp) which is hosted on crates.io as [rust-otp](https://crates.io/crates/rust-otp). We should switch to this version so we can actually pin our dependencies.

### DIFF
--- a/deepwell/Cargo.lock
+++ b/deepwell/Cargo.lock
@@ -899,7 +899,6 @@ dependencies = [
  "log",
  "notify",
  "once_cell",
- "otp",
  "paste",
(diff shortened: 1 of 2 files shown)


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1755, opened 2023-12-23
Fix ECR builds for dev services

emmiegit:
As part of the project of fixing dev deploys, we need to update the ECR build process that happens on push to the `develop` branch.


emmiegit:
First deployment! The task is failing after it starts due to runtime things but I think this is enough progress to make it mergeable.

Zokhoi (approved):

emmiegit:
thanks @Zokhoi

PR #1774, opened 2024-01-14
Use lz4 as default PostgreSQL compression algorithm

emmiegit:
By default, Postgres compresses large `EXTENDED` storage (such as `TEXT`) so the `COMPRESSION` directive on the column is unnecessary.

We also set the default compression algorithm to `lz4`, which is faster and has better compression than the default `pglz`. It was first added in Postgres 14.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1778, opened 2024-01-15
Fix wrongly-failing workflows

emmiegit:
Annoyingly, at some point in the past, imports began to be considered "unused" even if publicly exposed. Because we error on warnings, this fails builds, but we don't want that.

We've patched this before, but it's appearing again. [Example here](https://github.com/scpwiki/wikijump/actions/runs/7525167707/job/20481089470)

This adds `-A unused` to _all_ the workflow files, which should silence this issue.

### DIFF
--- a/.github/workflows/ftml.yaml
+++ b/.github/workflows/ftml.yaml
@@ -20,7 +20,7 @@ jobs:
     name: Library
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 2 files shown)


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1783, opened 2024-01-16
Update rust to 1.75.0

emmiegit:
This updates the listed versions of rustc to `1.75.0`.

Note, however, that we are keeping `async-trait` for now because it still has limitations that we want to avoid dealing with for now. Later when these restrictions are lifted through further improvements to stable, we will be able to remove the dependency and use `async fn` normally in `trait`s,


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1

PR #1835, opened 2024-03-10
Fix 4-dot ellipsis typography (WJ-1099)

MrShwhale:
This fix should make it so that only ellipsis not bordered by dots are replaced. Also expands the typography RegexReplace, though this is not used besides ellipsis. Adds a lot of ellipsis tests.

### DIFF
--- a/ftml/src/preproc/typography.rs
+++ b/ftml/src/preproc/typography.rs
@@ -118,7 +118,7 @@ impl Replacer {
                             .expect("Regular expression lacks a full match");
                         let mtch = capture.name("repl").unwrap_or(full_match);
 
-                        offset = mtch.start() + repl_len;
+                        offset = mtch.start() + replacement.len();
 
                         mtch.range()
                     };
--- a/ftml/src/preproc/typography.rs
+++ b/ftml/src/preproc/typography.rs
@@ -59,16 +59,17 @@ static LOW_DOUBLE_QUOTES: Lazy<Replacer> = Lazy::new(|| Replacer::RegexSurround 
 });
 
 // … - HORIZONTAL ELLIPSIS
-static ELLIPSIS: Lazy<Replacer> = Lazy::new(|| Replacer::RegexReplace {
-    regex: Regex::new(r"(?:\.\.\.|\. \. \.)").unwrap(),
+static HORIZONTAL_ELLIPSIS: Lazy<Replacer> = Lazy::new(|| Replacer::RegexReplace {
+    regex: Regex::new(r"(?:^|[^\.])(?<repl>(\.\.|\. \. )\.)(?:[^\.]|$)").unwrap(),
     replacement: "\u{2026}",
 });
 
 /// Helper struct to easily perform string replacements.
 #[derive(Debug)]
 pub enum Replacer {
-    /// Replaces any text matching the regular expression with the static string.
-    /// The entire match is used, any capture groups are ignored.
+    /// Replaces any text matching the "repl" group,
+    /// (or the entire regular expression if "repl" is happening)
+    /// with the static string.
     RegexReplace {
         regex: Regex,
         replacement: &'static str,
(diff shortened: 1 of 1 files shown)


ftml/src/preproc/typography.rs
```diff
@@ -107,13 +108,19 @@ impl Replacer {
...
+                let mut offset = 0;
+                let repl_len = replacement.len();
+
+                while let Some(capture) = regex.captures_at(text, offset) {
                     let range = {
-                        let mtch = capture
+                        let full_match = capture
                             .get(0)
                             .expect("Regular expression lacks a full match");
+                        let mtch = capture.name("repl").or(Some(full_match)).unwrap();
```
emmiegit: You don't need to add an extra layer of `Some(_)`:
```suggestion
                        let mtch = capture.name("repl").unwrap_or(full_match);
```

ftml/src/preproc/typography.rs
```diff
@@ -107,13 +108,19 @@ impl Replacer {
...
 
-                        mtch.start()..mtch.end()
+                        offset = mtch.start() + repl_len;
```
emmiegit: Since `repl_len` is only used once, we can get rid of the separate variable and just use:
```suggestion
                        offset = mtch.start() + replacement.len();
```

ftml/src/preproc/typography.rs
```diff
@@ -107,13 +108,19 @@ impl Replacer {
...
+
```
emmiegit, line 119:
```suggestion
```

ftml/src/preproc/typography.rs
```diff
@@ -189,14 +196,41 @@ const TEST_CASES: [(&str, &str); 3] = [
...
+        ". . .why would you think that?",
+        "…why would you think that?",
+    ),
+    ("how could you. . .", "how could you…"),
+    ("Whales . . . they are cool", "Whales … they are cool"),
+    (
+        ". . . why would you think that?",
+        "… why would you think that?",
+    ),
+    ("how could you . . .", "how could you …"),
```
emmiegit, line 217: I think this test structure is a bit strange. It looks like there are three classes of test, prefix, suffix, and infix (i.e. whether the `...`s are at the start, end, or in the middle), and that you test each of `...` and `. . .`. I would suggest that they have a consistent order, for instance:
```rust
"Whales... they are cool",
"Whales. . . they are cool",
"...why would you think that?",
". . .why would you think that?",
"how could you ...",
"how could you . . .",
```

ftml/src/preproc/typography.rs
```diff
@@ -189,14 +196,41 @@ const TEST_CASES: [(&str, &str); 3] = [
...
+    (".... ..", ".... .."),
+    (". . .. ....", ". . .. ...."),
+    ("... . . . . . .", "… … …"),
+    ("..........", ".........."),
+    ("... ... ...", "… … …"),
+    ("... . . . ...", "… … …"),
```
emmiegit, line 228: Then for each of these dot tests, you could include a brief prefix (or code comment) to denote its purpose. FOr instance:
```rust
"four dots .... ..",
"only two . . .. ....",
// or
"..........", // continuous dots
"... ... ...", // groups of three
```

ftml/src/preproc/typography.rs
MrShwhale: They have now been reordered, is it better now?

ftml/src/preproc/typography.rs
MrShwhale: I chose to leave comments, are they descriptive enough?

emmiegit (approved):
LGTM, thanks for the contribution!

emmiegit:
Merged into ftml repository at https://github.com/scpwiki/ftml/commit/25b5a97a7d062ae66d27700815f2be087657314e

PR #1894, opened 2024-04-23
[WJ-994] Fix deepwell build in Docker

emmiegit:
There were some glibc issues when trying to push the existing deepwell build, instead of bothering with the glibc/musl nonsense we can just use musl by building in the Rust Alpine image instead.

I have pushed the modified image to dev. This PR also adds changes to have it run there, like the environment-specific configuration files.

While working on this, I also added dev Dokku deploys for the "push" GitHub workflow files.


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1!

PR #1904, opened 2024-04-30
[WJ-994] Set up single image for Dokku dev deploy

emmiegit:
As mentioned in previous WJ-994 PRs, I have a separate VPS which holds Dokku on it, which will serve as our dev deployment for the time. When someone has the time and experience to provide us with something more modular / similar to our intended production environment, I'll be happy to help them migrate us to that.

In the mean time, I'm fixing the issues that occurred with the previous deployment, which involved separate deepwell and framerail containers. I have one single container here, which has separate processes/workers inside for them. This way we avoid the DNS-related issues that come from Dokku assumptions and can easily deploy the one app for everything.


Zokhoi (approved):

emmiegit:
thanks @Zokhoi

PR #1906, opened 2024-05-02
Update stylelint rules

Zokhoi:
The rule `property-case` was deprecated in stylelint v15 and removed in v16.

The equivalent lint should be covered by prettier if I'm not mistaken. If we find out later on that wasn't the case, we can introduce [`stylelint-stylistic`](https://github.com/stylelint-stylistic/stylelint-stylistic) that brings back the deprecated rules.


emmiegit (approved):
Thanks for the fix!

PR #1918, opened 2024-05-12
[WJ-1196] [WJ-1031] Add support for S3 presign URLs to upload blobs

emmiegit:
This PR changes the upload process to use S3 presign URLs. See the [Jira issue](https://scuttle.atlassian.net/browse/WJ-1196) to understand why we are changing the upload system.

This PR has a number of corresponding changes:
* Creates a `blob_pending` table to track in-progress uploads. This table will use [cuid](https://crates.io/crates/cuid2)s to generate external IDs, like other non-public data scopes such as messages.
* Adds `BlobService` methods for creating, cancelling, and finishing pending blob uploads. The last is not in the JSONRPC API but is instead called by the relevant services instead.
* Methods which previously wanted file data to upload now want a pending blob upload ID. This should correspond to an _uploaded_ blob, or it will yield an error to the user. Pending blobs can be reused for a limited time, in case of reuploading the same file in multiple places or fixing a bad request.
* Adds instructions to the DEEPWELL README explaining how to locally call requests and how to upload files to the generated presign URLs. Also adds convenience scripts for these uses.
* Adds an `echo` API request, for testing purposes.
* Fixes a bug in `ProvidedValue` that prevents it from being used to serialize data.
* DEEPWELL version is bumped.


Zokhoi (approved):

emmiegit:
thanks @Zokhoi

PR #2014, opened 2024-08-03
[WJ-1243] Add layout concept to DEEPWELL

emmiegit:
This adds layout support, added in ftml, to DEEPWELL.

Configuration is set at three levels. Later in the list means higher priority.
* **Platform configuration:** This specifies the default layout used platform-wide, if no other layouts are specified.
* **Site setting:** This is an optional field in the `site` table, setting a default layout for this site.
* **Page setting:** This is an optional field in the `page` table (not versioned), setting the layout override for this page.

I also added an API methods for updating the page layout override, and added the layout field in the method for updating site data. I ran the page layout method locally and verified the database was updated as expected.

### DIFF
--- a/deepwell/.gitattributes
+++ b/deepwell/.gitattributes
@@ -1,1 +1,2 @@
 src/models              binary linguist-vendored linguist-generated
+.sqlx                   binary linguist-vendored linguist-generated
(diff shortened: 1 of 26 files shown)


Zokhoi (approved):

emmiegit:
thanks @Zokhoi

PR #2030, opened 2024-08-14
Add typescript safety by declaring global types for Framerail.

Moonded:
I got annoyed by the lint errors because the types were not declared globally, so here they are.

I tried using the typings from deepwell, but as someone who has barely any knowledge in rust, I hope that I actually understood them right and not made some awful typings.


framerail/src/app.d.ts
```diff
@@ -1,9 +1,103 @@
...
+    /** Data about the site. */
+    site: {
+      site_id: string
+      [anySite: any]: unknown
+    }
+    /** Data about the page itself. */
+    page: {
+      page_id: string
+      page_created_at: string
+      page_updated_at?: string
```
emmiegit: From what I understand, `field?` means the field itself may not be included. I think that the correct type is `string | null` since it's nullable, not optional, in the response you get from the API service.

framerail/src/app.d.ts
```diff
@@ -1,9 +1,103 @@
...
+      page_deleted_at?: string
+      page_revision_count: number
+      site_id: mumber | string
```
emmiegit: Typo
```suggestion
      site_id: number | string
```

Also, in what case could this be a string?

Moonded: Not sure why i thought this could be string aswell.  fixed in [f5a15c0](https://github.com/scpwiki/wikijump/pull/2030/commits/f5a15c00e2d754c4ce21ec438dae2bae44ba284b)

framerail/src/app.d.ts
Moonded: Not sure if i understand what you mean with `field?` .

edit: got what you mean

framerail/src/app.d.ts
```diff
@@ -1,9 +1,103 @@
...
 // and what to do when importing types
+
 declare namespace App {
   // interface Locals {}
-  // interface PageData {}
-  // interface Error {}
+  interface PageData {
+    /** Data about the site. */
+    site: {
+      site_id: string
```
emmiegit, line 10: I thought we needed commas after each field, but if it builds then I guess not

emmiegit (approved):

framerail/src/app.d.ts
```diff
@@ -1,9 +1,103 @@
...
+      offset: number | null
+      data: string
+      /** @deprecated Use `no_render` instead. */
+      noRender: boolean
+      [anyOptions: any]: unknown
+    }
+    /** Rendered Wikitext */
+    wikitext: string
+    /** Internalization as defined in translation keys for the page. */
+    internationalization: any
```
Zokhoi: This could be `internationalization?: Record<string, string | null>` for better typing before the actual keys can be autogenerated.

framerail/src/app.d.ts
```diff
@@ -1,9 +1,103 @@
...
+  }
+
+  interface Error {
+    /** Error message */
+    message: string
+    /**
+     * Error internationalization as defined in the translation keys for
+     * the page. Look at /lib/types.ts for the keys type definitions.
+     */
+    internationalization: any
```
Zokhoi: This could be `internationalization?: Record<string, string | null>` for better typing before the actual keys can be autogenerated.

framerail/src/app.d.ts
Moonded: changed, see: [2118b0e](https://github.com/scpwiki/wikijump/pull/2030/commits/2118b0e9f517ec4bda784db031e42de89544d0bc)

framerail/src/app.d.ts
Moonded: changed, see: [2118b0e](https://github.com/scpwiki/wikijump/pull/2030/commits/2118b0e9f517ec4bda784db031e42de89544d0bc)

Zokhoi (approved):

Moonded:
I meant to create a new PR, but git is just git.

PR #2032, opened 2024-08-18
[WJ-1270] Temporary fix for seeder issue

emmiegit:
After having spent a day or so on it, the permanent fix for this will take more time, so for now I'll just do this as a temporary workaround so seeding works again. I also cherry-pick commits from the WJ-1270 permanent fix branch which is needed for the migration step.

This has two changes:
1. Fix for cyclic dependencies in the seeding and page creation process.
2. Use of temporary SeaORM implementations where the pages (and site) is only created in the SeaORM transaction and not yet in the database proper.

### DIFF
--- a/deepwell/Cargo.lock
+++ b/deepwell/Cargo.lock
@@ -690,7 +690,7 @@ dependencies = [
 
 [[package]]
 name = "deepwell"
-version = "2024.8.2"
+version = "2024.8.17"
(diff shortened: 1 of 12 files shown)


Yossipossi1 (approved):

emmiegit:
thanks @Yossipossi1!

PR #2066, opened 2024-09-08
[WJ-1272] Set up DigitalOcean App Platform as new dev

emmiegit:
I have set up the new dev environment, which instead of being on a (somewhat messy) Dokku instance on a single droplet, is instead hosted on [DigitalOcean App Platform](https://www.digitalocean.com/products/app-platform). The DNS for `wikijump.dev` is not yet set up, but you can see it here: https://wikijump-qdpi9.ondigitalocean.app. (The error is because it sees this as an incorrect custom domain, because it's DO's and not set in the database.)

This will automatically deploy on push to `develop`, and more recent deploys will not fail if a deploy is already pending making it suitable for rapid development.

https://scuttle.atlassian.net/browse/WJ-1272


deepwell/seeder/filters.json
```diff
@@ -162,7 +169,7 @@
     },
     {
         "site": null,
-        "regex": "^admin(ministrator)?",
+        "regex": "^admin(ministrators?)?",
```
Zokhoi: Should be `^admin(istrators?)?` ?

It matches anything starting with `admin` though, so the latter part isn't necessary.

emmiegit: True, we can just get rid of the rest.

Zokhoi (approved):

PR #2082, opened 2024-09-10
Add redirect for fallback route

Zokhoi:
The fallback comes up for routes like `//page`.

### DIFF
--- a/framerail/src/routes/[...fallback]/+page.svelte
+++ b/framerail/src/routes/[...fallback]/+page.svelte
@@ -3,12 +3,10 @@
 <script lang="ts">
   import { page } from "$app/stores"
   import Redirect from "$lib/redirect.svelte"
-  let redirectPage =
-    `/${ 
-    $page.params.fallback
-      .split("/")
-      .filter((v) => v.length)
-      .join("/")}`
+  let redirectPage = `/${$page.params.fallback
+    .split("/")
+    .filter((v) => v.length)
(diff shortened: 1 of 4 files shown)


framerail/src/routes/[...fallback]/+page.svelte
```diff
@@ -1 +1,12 @@
-<h1>UNTRANSLATED:General fallback route</h1>
+<!-- //[...fallback] should not be a valid page, therefore redirects to corresponding page -->
+
+<script lang="ts">
+  import { page } from "$app/stores"
+  import Redirect from "$lib/redirect.svelte"
+  let redirectPage = `/${$page.params.fallback
+    .split("/")
+    .filter((v) => v.length)
+    .join("/")}`
```
Comment (emmiegit) [inline framerail/src/routes/[...fallback]/+page.svelte:9]: So this will change something weird like `/scp/-/1000` into `/scp-1000`? I guess that's reasonable as a fallback route handler.

emmiegit (approved):

emmiegit:
Sensible redirection rules.

framerail/src/routes/[...fallback]/+page.svelte
Comment (Zokhoi) [reply framerail/src/routes/[...fallback]/+page.svelte:9]: No, it only filters out segments that are empty, for example `//page//edit` will become `/page/edit`.

Comment (emmiegit) [reply framerail/src/routes/[...fallback]/+page.svelte:9]: Oh you're right, I misread.

PR #2084, opened 2024-09-15
[WJ-1278] Fix custom domains

emmiegit:
The issue was that I was using the wrong JOIN condition. It seems that sea-orm-cli did not generate the relation I expected for `SiteDomain`, looking for a matching `custom_domain` column (i.e. the site's default domain). I manually added a standard relation for `site_domain -> site` and use that instead.

**Screenshot:**

Visiting `https://scpwiki.localhost` produces the following:

image

Which is the correct output, if it didn't recognize the custom domain it would produce the domain error page.


emmiegit:
Created https://github.com/SeaQL/sea-orm/issues/2358 for the `sea-orm-cli` entity generation issue.

Zokhoi (approved):

emmiegit:
thanks @Zokhoi

PR #2103, opened 2024-09-24
Frontend things [4]

Zokhoi:
Add some more UI functionality, closing in on [completion of basic page operations UI / WJ-1264](https://scuttle.atlassian.net/browse/WJ-1264).

Changes:
* View page rating (with test scorer)
* Revision rollback
* Rerender page
* Setting page layout
* Setting page parents

### DIFF
--- a/deepwell/src/endpoints/page.rs
+++ b/deepwell/src/endpoints/page.rs
@@ -22,9 +22,9 @@ use super::prelude::*;
 use crate::models::page::Model as PageModel;
 use crate::services::page::{
     CreatePage, CreatePageOutput, DeletePage, DeletePageOutput, EditPage, EditPageOutput,
-    GetPageAnyDetails, GetPageDirect, GetPageOutput, GetPageReferenceDetails,
-    GetPageScoreOutput, MovePage, MovePageOutput, RestorePage, RestorePageOutput,
-    RollbackPage, SetPageLayout,
+    GetPageAnyDetails, GetPageDirect, GetPageOutput, GetPageReference,
+    GetPageReferenceDetails, GetPageScoreOutput, MovePage, MovePageOutput, RestorePage,
+    RestorePageOutput, RollbackPage, SetPageLayout,
 };
 use crate::services::{Result, TextService};
 use crate::web::{PageDetails, Reference};
@@ -77,10 +77,9 @@ pub async fn page_get_score(
     ctx: &ServiceContext<'_>,
     params: Params<'static>,
 ) -> Result<Option<GetPageScoreOutput>> {
-    let GetPageReferenceDetails {
+    let GetPageReference {
         site_id,
         page: reference,
-        details: _,
     } = params.parse()?;
 
     info!("Getting score for page {reference:?} in site ID {site_id}");
--- a/deepwell/src/endpoints/page.rs
+++ b/deepwell/src/endpoints/page.rs
@@ -76,20 +76,16 @@ pub async fn page_get_direct(
 pub async fn page_get_score(
     ctx: &ServiceContext<'_>,
     params: Params<'static>,
-) -> Result<Option<GetPageScoreOutput>> {
+) -> Result<GetPageScoreOutput> {
     let GetPageReference {
         site_id,
         page: reference,
     } = params.parse()?;
 
     info!("Getting score for page {reference:?} in site ID {site_id}");
-    match PageService::get_id(ctx, site_id, reference).await {
-        Ok(page_id) => Ok(Some(GetPageScoreOutput {
-            page_id,
-            rating: ScoreService::score(ctx, page_id).await?,
-        })),
-        Err(_) => Ok(None),
-    }
+    let page_id = PageService::get_id(ctx, site_id, reference).await?;
+    let rating = ScoreService::score(ctx, page_id).await?;
+    Ok(GetPageScoreOutput { page_id, rating })
 }
 
 pub async fn page_edit(
--- a/deepwell/src/endpoints/page.rs
+++ b/deepwell/src/endpoints/page.rs
@@ -22,8 +22,9 @@ use super::prelude::*;
 use crate::models::page::Model as PageModel;
 use crate::services::page::{
     CreatePage, CreatePageOutput, DeletePage, DeletePageOutput, EditPage, EditPageOutput,
-    GetPageAnyDetails, GetPageDirect, GetPageOutput, GetPageReferenceDetails, MovePage,
-    MovePageOutput, RestorePage, RestorePageOutput, RollbackPage, SetPageLayout,
+    GetPageAnyDetails, GetPageDirect, GetPageOutput, GetPageReferenceDetails,
+    GetPageScoreOutput, MovePage, MovePageOutput, RestorePage, RestorePageOutput,
+    RollbackPage, SetPageLayout,
 };
 use crate::services::{Result, TextService};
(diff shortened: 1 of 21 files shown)


deepwell/src/endpoints/page.rs
```diff
@@ -72,6 +73,26 @@ pub async fn page_get_direct(
...
 }
 
+pub async fn page_get_score(
+    ctx: &ServiceContext<'_>,
+    params: Params<'static>,
+) -> Result<Option<GetPageScoreOutput>> {
+    let GetPageReferenceDetails {
+        site_id,
+        page: reference,
+        details: _,
```
emmiegit: Is there not a non-details version of the struct we can use here?

deepwell/src/endpoints/page.rs
```diff
@@ -72,6 +73,26 @@ pub async fn page_get_direct(
...
+    } = params.parse()?;
+
+    info!("Getting score for page {reference:?} in site ID {site_id}");
+    match PageService::get_id(ctx, site_id, reference).await {
+        Ok(page_id) => Ok(Some(GetPageScoreOutput {
+            page_id,
+            rating: ScoreService::score(ctx, page_id).await?,
+        })),
+        Err(_) => Ok(None),
```
emmiegit: Why are we ignoring errors? Shouldn't it be
```rs
let page_id = PageService::get_id(ctx, site_id, reference).await?;
let rating = ScoreService::score(ctx, page_id).await?;
Ok(GetPageScoreOutput {
    page_id,
    rating,
})
```

deepwell/src/services/view/service.rs
```diff
@@ -174,6 +172,15 @@ impl ViewService {
                 if Self::can_access_page(ctx, user_permissions).await? {
                     debug!("User has page access, return text data");
 
+                    if options.rerender {
```
emmiegit: So this is like `/some-page/rerender/true`? I don't think random viewers should be able to do this, only those with edit permission should be able to

deepwell/src/endpoints/page.rs
Zokhoi: Doesn't look like it. I'll make one then

deepwell/src/services/view/service.rs
Zokhoi: The viewing permission is stubbed, I'll add another stub for edit permissions

deepwell/src/endpoints/parent.rs
```diff
@@ -85,3 +88,90 @@ pub async fn parent_remove(
...
+    let pages: Vec<String> = PageService::get_pages(ctx, site_id, parents.as_slice())
+        .await?
+        .iter()
+        .map(|p| p.slug.clone())
+        .collect();
+
+    Ok(pages)
+}
+
+pub async fn parent_modify(
```
emmiegit: I think `update` is a better name, like get, create, update, delete as the basic ops.

deepwell/src/endpoints/parent.rs
```diff
@@ -85,3 +88,90 @@ pub async fn parent_remove(
...
+        .await?
+        .iter()
+        .map(|p| Reference::from(p.parent_page_id))
+        .collect();
+
+    let pages: Vec<String> = PageService::get_pages(ctx, site_id, parents.as_slice())
+        .await?
+        .iter()
+        .map(|p| p.slug.clone())
+        .collect();
```
emmiegit, line 114: A clone is only needed here because `.iter()` borrows, but nothing else takes ownership so we can just use that. Replace `.iter()` with `.into_iter()` and you can just do `|p| p.slug`

deepwell/src/endpoints/parent.rs
```diff
@@ -85,3 +88,90 @@ pub async fn parent_remove(
...
+
+    Ok(pages)
+}
+
+pub async fn parent_modify(
+    ctx: &ServiceContext<'_>,
+    params: Params<'static>,
+) -> Result<ModifyParentOutput> {
+    let input: ParentModifyDescription = params.parse()?;
```
emmiegit: Structs should be consistent, say `UpdateParents` and `UpdateParentsOutput`

deepwell/src/endpoints/parent.rs
```diff
@@ -85,3 +88,90 @@ pub async fn parent_remove(
...
+            for parent in parents {
+                if let Ok(Some(model)) = ParentService::create(
+                    ctx,
+                    ParentDescription {
+                        site_id: input.site_id,
+                        parent: parent.clone(),
+                        child: input.child.clone(),
+                    },
+                )
+                .await
```
emmiegit: This ignores any failures. Shouldn't the whole call fail if one of the constituent parts fails?

deepwell/src/endpoints/parent.rs
```diff
@@ -85,3 +88,90 @@ pub async fn parent_remove(
...
+    let input: ParentModifyDescription = params.parse()?;
+
+    info!(
+        "Modifying multiple parental relationship for child {:?} in site ID {}",
+        input.child, input.site_id,
+    );
+
+    let creation = match input.add {
+        Some(parents) => {
+            let mut creation = Vec::new();
```
emmiegit: Nitpick: We know the length of the end vector so we can preallocate:
```suggestion
            let mut creation = Vec::with_capacity(parents.len());
```

deepwell/src/services/page/structs.rs
```diff
@@ -107,6 +113,12 @@ pub struct GetPageOutput {
     pub layout: Layout,
 }
 
+#[derive(Serialize, Debug, Clone)]
+pub struct GetPageScoreOutput {
+    pub page_id: i64,
+    pub rating: ScoreValue,
```
emmiegit: Shouldn't the name be `score`? For consistency

deepwell/src/endpoints/parent.rs
Zokhoi: Would `try_join_all` revert all the parental relation creation or deletion if one of them failed? Otherwise the update would be half done, and there are no guarantees which ones would be added or deleted even if they could succeed.

Right now all that can succeed will succeed, so if the user specified some pages that exist and some that don't, it will add all the pages that do exist as parents.

emmiegit: In deepwell, all methods are wrapped in a transaction (it's what's inside the `ctx`). So if something fails then all DB changes will be reverted. Whether you use `try_join!()` (though since this is a dynamic list of futures it would need to be a [`JoinSet`](https://docs.rs/tokio/latest/tokio/task/join_set/struct.JoinSet.html) I guess) or run them one after another is not relevant.

The issue I was bringing attention to here is, because no error is raised (it is silenced), the transaction will go through and commit. If it gets passed up the chain then the transaction will get rolled back.

deepwell/src/endpoints/parent.rs
```diff
@@ -85,3 +89,93 @@ pub async fn parent_remove(
...
+                        parent: parent.to_owned(),
+                        child: input.child.clone(),
+                    },
+                )
+            });
+            Some(
+                try_join_all(removal)
+                    .await?
+                    .iter()
+                    .map(|p| p.was_deleted)
```
emmiegit, line 170: I guess I'm still not convinced that a list of the end structures (particularly the booleans) is all that useful, but I guess it's fine to keep it here for now.

emmiegit (approved):

PR #2127, opened 2024-10-07
Avoid domain dot removal code

emmiegit:
In the configuration struct, we have both `main_domain` (e.g. `.wikijump.com`) and `main_domain_no_dot` (e.g. `wikijump.com`). Since we already have one field with what we want (i.e. no dot prefix), then we don't need to do this manipulation and can just return the borrowed config field.


Zokhoi (approved):

emmiegit:
Thanks @Zokhoi

PR #2133, opened 2024-10-10
[WJ-1032] Add file_rollback

emmiegit:
This adds rollback capability to files, which like for page rollbacks, changes a file to have the same state it had earlier. (This cannot be used to move a file to a different page). While working on this, I discovered a bug in file moves where it wasn't properly updating the name.

Example output after local tests:
```
 revision_id | revision_type |          created_at           | revision_number | file_id | page_id | site_id | user_i>
-------------+---------------+-------------------------------+-----------------+---------+---------+---------+------->
           1 | create        | 2024-10-10 08:52:01.028316+00 |               0 |       1 |       1 |       1 |       >
           2 | regular       | 2024-10-10 08:52:13.901203+00 |               1 |       1 |       1 |       1 |       >
           3 | rollback      | 2024-10-10 08:52:21.541878+00 |               2 |       1 |       1 |       1 |       >
(3 rows)
```

### DIFF
--- a/deepwell/src/models/sea_orm_active_enums.rs
+++ b/deepwell/src/models/sea_orm_active_enums.rs
@@ -20,14 +20,18 @@ pub enum AliasType {
 #[sea_orm(rs_type = "String", db_type = "Enum", enum_name = "file_revision_type")]
 #[serde(rename_all = "kebab-case")]
 pub enum FileRevisionType {
-    #[sea_orm(string_value = "regular")]
-    Regular,
     #[sea_orm(string_value = "create")]
     Create,
     #[sea_orm(string_value = "delete")]
     Delete,
+    #[sea_orm(string_value = "regular")]
+    Regular,
+    #[sea_orm(string_value = "rollback")]
+    Rollback,
     #[sea_orm(string_value = "undelete")]
     Undelete,
+    #[sea_orm(string_value = "undo")]
+    Undo,
 }
 #[derive(
     Debug, Clone, PartialEq, Eq, EnumIter, DeriveActiveEnum, Copy, Serialize, Deserialize,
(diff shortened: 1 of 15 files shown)


deepwell/src/models/sea_orm_active_enums.rs
```diff
@@ -60,8 +64,12 @@ pub enum PageRevisionType {
     Move,
     #[sea_orm(string_value = "regular")]
     Regular,
+    #[sea_orm(string_value = "rollback")]
+    Rollback,
     #[sea_orm(string_value = "undelete")]
     Undelete,
+    #[sea_orm(string_value = "undo")]
+    Undo,
```
Zokhoi, line 72: What does `PageRevisionType::Undo` mean? Is it like the opposite of a rollback?

emmiegit: Undos are a [planned operation](https://scuttle.atlassian.net/browse/WJ-990). I am avoiding the term "revert" because it means two different things in git vs wikidot, and instead have two unambiguous terms:

* Rollback (what wikidot calls "revert", what git calls "reset") is where you set the current state of a page or file to some revision in the past.
* Undo (what git calls "revert") is where it applies the opposite of the given revision. Since this involves applying a diff, it can fail.

An example of the difference  between the two. Imagine we have a page with this history:

<details>
<summary>Revision 10</summary>

```md
**Item #:** SCP-XX

**Object Class:** Euclid

**Special Containment Procedures:** SCP-XX is to be kept in a cardboard box.
```

</details>

<details>
<summary>Revision 11</summary>

```md
**Item #:** SCP-XX

**Object Class:** HAHA EPIC TROLL

**Special Containment Procedures:** SCP-XX is to be kept in a cardboard box.
```

</details>

</details>

<details>
<summary>Revision 12</summary>

```md
**Item #:** SCP-XX

**Object Class:** HAHA EPIC TROLL

**Special Containment Procedures:** SCP-XX is to be kept in a cardboard or paper box.
```

</details>

So here, revision 11 is vandalism and we'd like to remove it. But the page has been edited since, so that means we'd lose any subsequent (legitimate) changes with a simple rollback. But using undo, similar to `git revert`, we can remove just the specific change from 10 -> 11 and leave the 12 revision in-place.

So basically it would produce the 11..10 diff (not that this is the _inverse_ of the 10..11 diff that was applied in revision 11):
```diff
3c3
< **Object Class:** HAHA EPIC TROLL
---
> **Object Class:** Euclid
```

And then we could attempt to apply it.

(If there are conflicts, then like `git revert`, we should error out rather than attempting to ram our way through. Also, we should have some way of allowing a user to verify that the undo change is what they want before doing it.)

Zokhoi (approved):

emmiegit:
thanks @Zokhoi

PR #2186, opened 2024-11-23
Various fixes for frontend

Zokhoi:
Fixes frontend for some of the changes introduced in deepwell:

* page operations passing last revision id #2117
* user avatar upload using presign url #1918

Also adds stub file for [platform information page (WJ-1279)](https://scuttle.atlassian.net/jira/software/c/projects/WJ/issues/WJ-1279)


framerail/src/routes/[slug]/[...extra]/page.svelte
```diff
@@ -257,6 +260,7 @@
     fdata.set("site-id", $page.data.site.site_id)
     fdata.set("page-id", $page.data.page.page_id)
     fdata.set("revision-number", revisionNumber)
+    fdata.set("last-revision-id", $page.data.page_revision.revision_id)
```
Comment (emmiegit) [inline framerail/src/routes/[slug]/[...extra]/page.svelte:263]: This sets it as a string? (Assuming since you use `parseInt()` above)

framerail/src/routes/about/+page.svelte
```diff
@@ -0,0 +1 @@
+<h1>UNTRANSLATED:TODO: info page</h1>
```
emmiegit, line 1: Not relevant for this PR obviously but the JSONRPC method for server info is `info`.

emmiegit (approved):

framerail/src/routes/[slug]/[...extra]/page.svelte
Comment (Zokhoi) [reply framerail/src/routes/[slug]/[...extra]/page.svelte:263]: All data sent by forms are either `string` or `Blob`, so to get `number`, `boolean` or other types we need to parse it from string.

Comment (emmiegit) [reply framerail/src/routes/[slug]/[...extra]/page.svelte:263]: Ah, I see!

PR #2193, opened 2024-11-28
[WJ-1032] [WJ-1287] Add hard deletion blob operation, add files to seeder

emmiegit:
This PR took a bit, and as part of developing it, I ended up implementing [WJ-1287](https://scuttle.atlassian.net/browse/WJ-1287) as well because it made it easier to debug the issues with [WJ-1032](https://scuttle.atlassian.net/browse/WJ-1032), where the hard deletion process was stalling.

I also add some general improvements to blob management, particularly with error handling. More information is now sent to API consumers for some classes of error.

## Blob Hard Deletion

Blobs throughout Wikijump are maintained in a content-addressable system which eliminates duplicates and makes it easy to reuse files and file revisions and similar data objects throughout the code. However, there was a stubbed operation which here in this PR, we finally implement.

Normally, a deletion just adds a new tombstone revision which marks the file as deleted. This enables easy restoration of the file by adding a resurrection revision afterwards.

But in some rare circumstances (e.g. court order, presence of illegal/abuse material, etc.) the platform will want the ability to purge blob data, meaning it is actually deleted within S3. In such a case, all uses of the blob throughout the platform should be deleted.

This is implemented in a pair of operations, one of which is just the read-only version of the other. First, you submit an S3 hash to expunge, and it reports what data it would've deleted if the operation went through. The other returns the same but also makes the changes.

An example output follows:

```json
{
    "total_revisions": 6,
    "total_files": 5,
    "total_files_deleted": 3,
    "total_pages": 2,
    "total_sites": 2,
    "total_users": 0,
    "sample_revision_ids": [2, 4, 5, 7, 8, 9],
    "sample_file_ids": [2, 4, 5, 6, 7],
    "sample_page_ids": [1, 29],
    "sample_site_ids": [1, 2],
    "sample_user_ids": []
}
```

This involves:
1. Users with this blob set as their avatar have their profile picture set to be blank.
2. If a blob is used as the current version of a file, it is deleted.
3. For all revisions (now non-current), the hash is removed (replaced with the empty hash), and the hash column is also hidden.
4. The S3 hash is blacklisted, preventing any new blobs from ever being uploaded with this hash in the future.
5. Then the S3 object itself is deleted from the bucket.

This makes it so that the file data is fully, comprehensively removed from the system. It also permits the calling platform staff member to see what results the operation will have, and for the UI to display information about it.

After the blob is blacklisted, trying to upload it again later is not permitted:
```json
{
    "code": 4026,
    "message": "Uploaded blob content is blacklisted",
    "data": null
}
```

With the change made in the following section, you can run the `blob_hard_delete_list` (preview set of changes) and `blob_hard_delete_confirm` (apply set of changes) on a freshly-seeded database to see it produce output.

## File Seeding

This also adds seeding of files. Files can be added in `seeder/files.json`, which are then added during the seeding process, including editing and deleting files (so we get a revision history with different blobs). Files to be uploaded are sourced from the same seeder directory, please keep them small to avoid wasting repository space. (The added images are either from Wikidot or are public domain.)

Here is what the `file` table looks like after seeding:
```sql
> SELECT * FROM file;
 file_id |          created_at           |          updated_at           |          deleted_at           | from_wikidot |        name        | page_id | site_id 
---------+-------------------------------+-------------------------------+-------------------------------+--------------+--------------------+---------+---------
       1 | 2024-11-28 07:42:45.392146+00 |                               |                               | f            | default_avatar.png |       1 |       1
       2 | 2024-11-28 07:42:45.392146+00 |                               |                               | f            | lock.svg           |       1 |       1
       3 | 2024-11-28 07:42:45.392146+00 |                               |                               | f            | null               |       1 |       1
       4 | 2024-11-28 07:42:45.392146+00 |                               |                               | f            | restricted.svg     |      29 |       2
       5 | 2024-11-28 07:42:45.392146+00 | 2024-11-28 07:42:50.94121+00  |                               | f            | restricted2.svg    |      29 |       2
       6 | 2024-11-28 07:42:45.392146+00 |                               | 2024-11-28 07:42:51.054437+00 | f            | restricted3.svg    |      29 |       2
       7 | 2024-11-28 07:42:45.392146+00 | 2024-11-28 07:42:51.109073+00 | 2024-11-28 07:42:51.172743+00 | f            | restricted4.svg    |      29 |       2
       8 | 2024-11-28 07:42:45.392146+00 |                               |                               | f            | seed_pages.JSON    |      32 |       2
       9 | 2024-11-28 07:42:45.392146+00 |                               |                               | f            | seed_users.JSON    |      32 |       2
      10 | 2024-11-28 07:42:45.392146+00 |                               |                               | f            | seed_sites.JSON    |      32 |       2
      11 | 2024-11-28 07:42:45.392146+00 | 2024-11-28 07:42:51.727185+00 |                               | f            | SCP.txt            |      12 |       4
(11 rows)
```

[WJ-1287]: https://scuttle.atlassian.net/browse/WJ-1287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WJ-1032]: https://scuttle.atlassian.net/browse/WJ-1032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Zokhoi (approved):

emmiegit:
thanks Zokhoi

PR #2214, opened 2024-12-15
File actions for frontend

Zokhoi:
Adds frontend UI for performing page file actions:
* List
* Upload
* Delete
* Edit
* Move
* Restore
* Revision history
* Rollback

### DIFF
--- a/framerail/src/routes/[slug]/[...extra]/page.svelte
+++ b/framerail/src/routes/[slug]/[...extra]/page.svelte
@@ -531,6 +531,33 @@
     }
   }
 
+  async function rollbackFileRevision(revisionNumber: number, comments?: string) {
+    let fdata = new FormData()
+    fdata.set("site-id", $page.data.site.site_id)
+    fdata.set("page-id", $page.data.page.page_id)
+    fdata.set("file-id", fileEditId)
+    fdata.set("revision-number", revisionNumber)
+    fdata.set("last-revision-id", fileMap.get(fileEditId)?.revision_id)
+    if (comments !== undefined) fdata.set("comments", comments)
+    let res = await fetch(`/${$page.data.page.slug}/file-rollback`, {
+      method: "POST",
+      body: fdata
+    }).then((res) => res.json())
+    if (res?.message) {
+      showErrorPopup.set({
+        state: true,
+        message: res.message,
+        data: res.data
+      })
+    } else {
+      getFileList()
+      showFileHistory = false
+      fileRevisionMap = new Map()
+      handleFileHistory(fileEditId)
+      invalidateAll()
+    }
+  }
+
   onMount(() => {
     if ($page.data?.options.history) handleHistory()
   })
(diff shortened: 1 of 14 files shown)


framerail/src/routes/[slug]/[...extra]/page.svelte
```diff
@@ -789,6 +1006,428 @@
...
+          </div>
+          <div class="file-attribute mime">
+            {$page.data.internationalization?.["wiki-page-file.mime"]}
+          </div>
+          <div class="file-attribute size">
+            {$page.data.internationalization?.["wiki-page-file.size"]}
+          </div>
+          <div class="file-attribute action" />
+        </div>
+        {#each [...fileMap].sort((a, b) => b[0] - a[0]) as [_, file] (file.file_id)}
```
Comment (emmiegit) [inline framerail/src/routes/[slug]/[...extra]/page.svelte:1058]: I assume the 0 index here is the filename or something?

framerail/src/routes/[slug]/[...extra]/page.svelte
```diff
@@ -789,6 +1006,428 @@
...
+              {revisionItem.user_id}
+            </div>
+            <div class="revision-attribute page">
+              {revisionItem.page_id}
+            </div>
+            <div class="revision-attribute name">
+              {revisionItem.name}
+            </div>
+            <div class="revision-attribute mime">
+              {revisionItem.mime_hint.split(";")[0]}
```
Comment (emmiegit) [inline framerail/src/routes/[slug]/[...extra]/page.svelte]: I don't think MIMEs will always have a semicolon, and also, why can't we just show the whole MIME type to users?

emmiegit (approved):

framerail/src/routes/[slug]/[...extra]/page.svelte
Comment (Zokhoi) [reply framerail/src/routes/[slug]/[...extra]/page.svelte]: I personally think the charset is kind of an unnecessary info, but we can go with showing the whole thing

Comment (emmiegit) [reply framerail/src/routes/[slug]/[...extra]/page.svelte]: My thinking is that people interested in MIME types are likely to be technical, and so then providing the whole string is more useful than not.

Comment (emmiegit) [reply framerail/src/routes/[slug]/[...extra]/page.svelte]: I am pretty sure there is a way to get a textual, more human-readable file type from a MIME. However I don't know if our current magic library provides it, nor what its internationalization status is.

framerail/src/routes/[slug]/[...extra]/page.svelte
Comment (Zokhoi) [reply framerail/src/routes/[slug]/[...extra]/page.svelte:1058]: Using the spread operator on a map gives an iterator of `[key, value]`, so in this case `[0]` refers to the key, which is the file id, so this each block sorts by file id in descending order.

framerail/src/routes/[slug]/[...extra]/page.svelte
Comment (emmiegit) [reply framerail/src/routes/[slug]/[...extra]/page.svelte]: So I just checked, the current crate we use, `filemagic`, doesn't do this. So adding a "human-readable file description" will need to be an extension added later if that's what we do.

Comment (Zokhoi) [reply framerail/src/routes/[slug]/[...extra]/page.svelte]: Wikidot provides the file type as a pretty rudimentary human readable string, for example it shows .png as "PNG image data" but .js as "No description"

PR #2231, opened 2024-12-28
[WJ-1292] Wiki page refactor

Zokhoi:
The main wiki page component `routes/[slug]/[...extra]/page.svelte` is getting pretty big as we add components to it, so this PR refactors the page and separates the components into their own files.


framerail/src/routes/[slug]/[...extra]/EditorPane.svelte
```diff
@@ -0,0 +1,103 @@
+<script lang="ts">
+  import { page } from "$app/stores"
+  import { goto } from "$app/navigation"
+  import { useErrorPopup } from "$lib/stores"
+  let showErrorPopup = useErrorPopup()
+
+  function cancelEdit() {
+    let options: string[] = []
+    if ($page.data.options.no_render) options.push("norender")
```
Comment (emmiegit) [inline framerail/src/routes/[slug]/[...extra]/EditorPane.svelte:9]: Maybe it makes sense for us to start storing the entire "extra" part of the URL as-is so it can be easily re-attached in cases like this?

framerail/src/routes/[slug]/[...extra]/VotePane.svelte
```diff
@@ -0,0 +1,152 @@
...
+      >
+        0
+      </button>
+      <button
+        class="vote-subbutton clickable"
+        type="button"
+        on:click|stopPropagation={() => castVote(-1)}
+      >
+        -1
+      </button>
```
Comment (emmiegit) [inline framerail/src/routes/[slug]/[...extra]/VotePane.svelte:133]: Note: This is for the standard vote scheme but we will need to fit page vote settings into here at some point (e.g. disabled voting and 5-star).

emmiegit (approved):

framerail/src/routes/[slug]/[...extra]/EditorPane.svelte
Comment (Zokhoi) [reply framerail/src/routes/[slug]/[...extra]/EditorPane.svelte:9]: Some extra parts are only relevant to editing a page (parent, title, tags) while some others are only relevant to viewing (comments, history, offset), and some options are relevant in both scenarios (no_render, no_redirect). So I'm not sure if reattaching the entire extra part will be useful.

framerail/src/routes/[slug]/[...extra]/VotePane.svelte
Comment (Zokhoi) [reply framerail/src/routes/[slug]/[...extra]/VotePane.svelte:133]: Created [WJ-1294 - UI for alternate voting mechanisms](https://scuttle.atlassian.net/browse/WJ-1294)

framerail/src/routes/[slug]/[...extra]/EditorPane.svelte
Comment (emmiegit) [reply framerail/src/routes/[slug]/[...extra]/EditorPane.svelte:9]: That is a good point, I forgot about that. I think this arrangement makes sense then.

PR #2238, opened 2025-01-03
Update ftml wasm build

Zokhoi:
Updated wasm build to match ftml 1.28.1.

Also updated some helper functions due to changes in #1297 and [ftml#17](https://github.com/scpwiki/ftml/pull/17).


emmiegit (approved):
Thank you!

PR #146, opened 2021-03-25
Update rust to 1.51.0

emmiegit:
See the [Rust version announcement post](https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html).

PR #2332, opened 2025-04-14
Update esbuild in framerail

emmiegit:
This resolves [security advisory #146](https://github.com/scpwiki/wikijump/security/dependabot/146) by upgrading `esbuild` to a fixed version. There are several parent depenencies, but upgrading `vite` we can also upgrade `esbuild`.

This bumps a vite to a new major version, which brings in several changes, see [migrating from v5](https://vite.dev/guide/migration.html) on the vite website. Further changes to adjust for the vite upgrade are needed, as as-is framerail does not render properly.

### DIFF
--- a/framerail/package.json
+++ b/framerail/package.json
@@ -51,6 +51,6 @@
     "svelte-check": "^4.1.6",
     "svelte-preprocess": "^6.0.3",
     "tslib": "^2.8.1",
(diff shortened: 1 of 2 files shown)


hoah2333:
Closing this as it's already implemented in #2713.

PR #2337, opened 2025-04-21
Update mininum name length: add separate check for chars

emmiegit:
This adds a separate check for chars (codepoints) in addition to the one for bytes. This way, you can set a byte requirement that restricts short ASCII names but are more permissive for other scripts, like names using CJK characters.

https://scuttle.atlassian.net/browse/WJ-1122

### DIFF
--- a/deepwell/src/services/alias/service.rs
+++ b/deepwell/src/services/alias/service.rs
@@ -126,7 +126,7 @@ impl AliasService {
                     error!(
                         "User's name is not long enough ({} < {} chars)",
                         slug.len(),
-                        ctx.config().minimum_name_bytes,
+                        ctx.config().minimum_name_chars,
                     );
                     return Err(Error::UserNameTooShort);
                 }
--- a/deepwell/src/services/alias/service.rs
+++ b/deepwell/src/services/alias/service.rs
@@ -85,7 +85,6 @@ impl AliasService {
                     error!(
                         "No target site with ID {target_id} exists, cannot create alias",
                     );
-
                     return Err(Error::SiteNotFound);
                 }
 
@@ -94,7 +93,6 @@ impl AliasService {
                     error!(
                         "Site with conflicting slug '{slug}' already exists, cannot create alias",
                     );
-
                     return Err(Error::SiteExists);
                 }
             }
@@ -103,7 +101,6 @@ impl AliasService {
                     error!(
                         "No target user with ID {target_id} exists, cannot create alias",
                     );
-
                     return Err(Error::UserNotFound);
                 }
 
--- a/deepwell/src/services/alias/service.rs
+++ b/deepwell/src/services/alias/service.rs
@@ -124,7 +124,7 @@ impl AliasService {
 
                 if slug.chars().count() < config.minimum_name_chars {
                     error!(
-                        "User's name is not long enough ({} < {} bytes)",
+                        "User's name is not long enough ({} < {} chars)",
                         slug.len(),
                         ctx.config().minimum_name_bytes,
                     );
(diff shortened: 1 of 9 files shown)


deepwell/src/services/alias/service.rs
```diff
@@ -112,17 +109,25 @@ impl AliasService {
...
                         ctx.config().minimum_name_bytes,
                     );
+                    return Err(Error::UserNameTooShort);
+                }
 
+                if slug.chars().count() < config.minimum_name_chars {
+                    error!(
+                        "User's name is not long enough ({} < {} bytes)",
+                        slug.len(),
+                        ctx.config().minimum_name_bytes,
```
Zokhoi, line 129: Should this error print the number of chars instead?

deepwell/src/services/user/service.rs
```diff
@@ -65,14 +65,23 @@ impl UserService {
             return Err(Error::UserSlugEmpty);
         }
 
-        // Check if username contains the minimum amount of required bytes.
-        if name.len() < ctx.config().minimum_name_bytes {
+        // Check if username contains the minimum amount of required bytes and chars.
+        let config = ctx.config();
+        if name.len() < config.minimum_name_bytes {
             error!(
                 "User's name is not long enough ({} < {})",
```
Zokhoi: `({} < {} bytes)` to be consistent with other places?

deepwell/src/services/user/service.rs
```diff
@@ -65,14 +65,23 @@ impl UserService {
...
                 ctx.config().minimum_name_bytes,
             );
+            return Err(Error::UserNameTooShort);
+        }
 
+        if name.chars().count() < config.minimum_name_chars {
+            error!(
+                "User's name is not long enough ({} < {} bytes)",
+                slug.len(),
+                ctx.config().minimum_name_bytes,
```
Zokhoi, line 83: Same as previous for printing the number of chars instead

deepwell/src/services/alias/service.rs
```diff
@@ -112,17 +109,25 @@ impl AliasService {
...
+                    );
```
Zokhoi: Should this error print the number of chars instead?

deepwell/src/services/user/service.rs
```diff
@@ -65,14 +65,23 @@ impl UserService {
...
+            );
```
Zokhoi: Same as previous for printing the number of chars instead

deepwell/src/services/user/service.rs
emmiegit: Good catches for all of these. Fixing.

deepwell/src/services/alias/service.rs
```diff
@@ -112,17 +109,25 @@ impl AliasService {
...
                         ctx.config().minimum_name_bytes,
                     );
+                    return Err(Error::UserNameTooShort);
+                }
 
+                if slug.chars().count() < config.minimum_name_chars {
+                    error!(
+                        "User's name is not long enough ({} < {} chars)",
+                        slug.len(),
+                        ctx.config().minimum_name_bytes,
```
Zokhoi: Also change this to show min chars instead of min bytes?

deepwell/src/services/user/service.rs
```diff
@@ -65,14 +65,23 @@ impl UserService {
...
                 ctx.config().minimum_name_bytes,
             );
+            return Err(Error::UserNameTooShort);
+        }
 
+        if name.chars().count() < config.minimum_name_chars {
+            error!(
+                "User's name is not long enough ({} < {} chars)",
+                slug.len(),
+                ctx.config().minimum_name_bytes,
```
Zokhoi: Ditto

deepwell/src/services/alias/service.rs
emmiegit: I guess I'm not on my game today. Fixed.

Zokhoi (approved):

emmiegit:
Thanks!

PR #2355, opened 2025-05-04
Fix incorrect page view enum cases

emmiegit:
As described in the first commit, this broke when the enum variant names in deepwell were modified but the corresponding code in framerail was not. This PR fixes that and adds an error case for if none of the variants match.

### DIFF
--- a/framerail/src/routes/+error.svelte
+++ b/framerail/src/routes/+error.svelte
@@ -221,7 +221,7 @@ as soon as we can figure out prettier support for it.
   UNTRANSLATED:No such site
   {@html $page.error.html}
 {:else}
-  UNTRANSLATED:Fallback error, something really went wrong
+  UNTRANSLATED:Fallback error, invalid view type from backend
 {/if}
 
 <style global lang="scss">
--- a/framerail/src/routes/+error.svelte
+++ b/framerail/src/routes/+error.svelte
@@ -218,7 +218,7 @@ as soon as we can figure out prettier support for it.
   UNTRANSLATED:Lacks permissions for page
   {@html $page.error.compiled_html}
 {:else}
-  UNTRANSLATED:Fallback error, invalid view type from backend
+  UNTRANSLATED:Fatal error: Unable to display view
 {/if}
 
 <style global lang="scss">
--- a/framerail/src/routes/+error.svelte
+++ b/framerail/src/routes/+error.svelte
@@ -84,7 +84,7 @@
 Use svelte-switch-case package with {#switch data.view}
 as soon as we can figure out prettier support for it.
 -->
-{#if $page.error.view === "page_missing"}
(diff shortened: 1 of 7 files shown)


Zokhoi:
Also change `framerail/src/routes/+error.svelte`? Lines 87, 217, 220

framerail/src/routes/+error.svelte
```diff
@@ -214,14 +214,11 @@ as soon as we can figure out prettier support for it.
...
-{:else if $page.error.view === "page_permissions"}
+{:else if $page.error.view === "permissions"}
   UNTRANSLATED:Lacks permissions for page
   {@html $page.error.compiled_html}
-{:else if $page.error.view === "site_missing"}
-  UNTRANSLATED:No such site
-  {@html $page.error.html}
 {:else}
-  UNTRANSLATED:Fallback error, something really went wrong
+  UNTRANSLATED:Fallback error, invalid view type from backend
```
Zokhoi: This case is the generic fallback, I think we can keep the message as is, since the error might not actually be invalid view type. It could also be things like unable to connect to backend.

framerail/src/routes/[x+2d]/admin/+error.svelte
```diff
@@ -17,7 +17,7 @@ as soon as we can figure out prettier support for it.
   UNTRANSLATED:No such site
   {@html $page.error.html}
 {:else}
-  UNTRANSLATED:Fallback error, something really went wrong
+  UNTRANSLATED:Fallback error, invalid view type from backend
```
Comment (Zokhoi) [inline framerail/src/routes/[x+2d]/admin/+error.svelte]: ditto

framerail/src/routes/[x+2d]/user/+error.svelte
```diff
@@ -17,7 +17,7 @@ as soon as we can figure out prettier support for it.
   UNTRANSLATED:No such site
   {@html $page.error.html}
 {:else}
-  UNTRANSLATED:Fallback error, something really went wrong
+  UNTRANSLATED:Fallback error, invalid view type from backend
```
Comment (Zokhoi) [inline framerail/src/routes/[x+2d]/user/+error.svelte]: ditto

framerail/src/routes/+error.svelte
emmiegit: That is true, though I feel like "went really wrong" is kind of not the best message. Like you mentioned, there are a number of issues that can cause this.

Maybe `Fatal error: Unable to display view`?

Zokhoi: That could do, I guess when we localize it in the future we can reword it to be more verbose

Zokhoi (approved):

framerail/src/routes/+error.svelte
emmiegit: Absolutely, yeah

PR #2573, opened 2025-10-31
Fix in-place string substitution functions

emmiegit:
These were implemented quickly and only used for a few simple cases. Because no unit tests were put in place, some of these had some bugs. This PR fixes these issues and adds test cases to catch regressions.

Due to some of the complexity required to fix `char_replace_in_place()`, I scrapped it and am just using `regex_replace_in_place()` instead. This was meant to be a quick function definition, if that wasn't working then it's not worth it.

### DIFF
--- a/deepwell/src/services/special_page/service.rs
+++ b/deepwell/src/services/special_page/service.rs
@@ -182,7 +182,7 @@ impl SpecialPageService {
         //
         // See https://fluent-compiler.readthedocs.io/en/latest/usage.html#:~:text=You%20will%20notice%20the%20extra%20characters%20\u2068%20and%20\u2069%20in%20the%20output.
         static CONTROL_CHAR_REGEX: LazyLock<Regex> =
-            LazyLock::new(|| Regex::new("[\u{2068}|\u{2069}]").unwrap());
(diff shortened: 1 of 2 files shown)


deepwell/src/services/special_page/service.rs
```diff
@@ -179,7 +181,10 @@ impl SpecialPageService {
         // a desired localization property of Fluent.
         //
         // See https://fluent-compiler.readthedocs.io/en/latest/usage.html#:~:text=You%20will%20notice%20the%20extra%20characters%20\u2068%20and%20\u2069%20in%20the%20output.
-        char_replace_in_place(&mut wikitext, &['\u{2068}', '\u{2069}'], "");
+        static CONTROL_CHAR_REGEX: LazyLock<Regex> =
+            LazyLock::new(|| Regex::new("[\u{2068}|\u{2069}]").unwrap());
```
Zokhoi: I think this square bracket regex also replaces the pipe character?

emmiegit: Oh you're right. A dash would've been fine or not having a char group but this isn't okay. Damn, good catch.

Zokhoi (approved):

PR #2586, opened 2025-11-14
[WJ-1308] Add frontend support for legacy layout

Zokhoi:
This adds legacy layout with switcher where applicable, subject to site and page settings.


framerail/src/routes/+layout.svelte
```diff
@@ -12,60 +14,137 @@
...
+  function setLayout() {
+    if ($page.route.id.startsWith("/[x+2d]/")) {
+      // this is a special page, use Wikijump layout
+      $pageLayout = Layout.WIKIJUMP
+    } else {
+      $pageLayout =
+        $page.data?.page?.layout ??
+        $page.data?.site?.layout ??
+        $page.error?.site?.layout ??
+        Layout.WIKIJUMP
```
emmiegit, line 28: Oh hmm. So individual `layout` fields are exposed, but the computed "layout for this page" (as calculated in deepwell methods) is not. I guess we don't really need them then?

emmiegit (approved):
Very nice!

framerail/src/routes/+layout.svelte
Zokhoi: You mean `SettingsService::get_layout`? Wasn't that used for page rendering and passing to ftml?

I see there are layout settings at the category level, so if we are to unify the fallback chain we can just call `get_layout` in page viewer and expose it

emmiegit: Makes sense. For now this logic is fine though.

PR #2601, opened 2025-12-03
Add page delete pane

Zokhoi:
This adds a pane for page options > delete, giving the operator the options of moving the page to `deleted` category or deleting it entirely. Currently no second confirmation is done for deleting the page.

### DIFF
--- a/locales/fluent/base/zh_Hans.ftl
+++ b/locales/fluent/base/zh_Hans.ftl
@@ -30,6 +30,7 @@ cancel = 取消
 change = 修改
 clear = 清空
 close = 关闭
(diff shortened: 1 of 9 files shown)


locales/fluent/base/zh_Hans.ftl
```diff
@@ -30,6 +30,7 @@ cancel = 取消
 change = 修改
 clear = 清空
 close = 关闭
+confirm = 确认
```
emmiegit, line 33: Huh, cool simplification of 確認

emmiegit (approved):
Thanks for adding!

PR #2602, opened 2025-12-07
[WJ-1285] Don't allow rollback to expose hidden fields

Syxless:
[Fixes WJ-1285.](https://scuttle.atlassian.net/jira/software/c/projects/WJ/issues/?jql=project%20%3D%20%22WJ%22%20ORDER%20BY%20created%20DESC&selectedIssue=WJ-1285)

Rollbacks could rehydrate fields that moderators previously hid. This patch makes rollback respect revision.hidden for both pages and files.

Page rollback:
- Skip wikitext restore when 'wikitext' is hidden.
- Only restore title, alt_title, tags if they are not hidden.

File rollback:
- Only restore name, s3_hash, mime, size when they are not hidden.

Field names reflect the existing hidden contract used in page/file endpoints.

Prevents rollback from exposing suppressed content. No tests requested.

### DIFF
--- a/deepwell/src/services/file/service.rs
+++ b/deepwell/src/services/file/service.rs
@@ -508,9 +508,6 @@ impl FileService {
             FileRevisionService::get_latest(ctx, site_id, page_id, file_id),
         )?;
 
-        // TODO Handle hidden fields, see https://scuttle.atlassian.net/browse/WJ-1285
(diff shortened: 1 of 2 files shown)


emmiegit (approved):
Looks good to me!

PR #2607, opened 2025-12-08
WJ-1290: convert page attribution to relation

Syxless:
Convert page attribution into the generic relation system.

- Remove the dedicated page_attribution table from the base schema.
- Add page-attribution relation type (page -> user) with metadata {attribution_type, attribution_date}.
- Enforce uniqueness via relation index on (relation_type, page, user, attribution_type, attribution_date).
- Drop SeaORM page_attribution model/references; wire new relation types/metadata.

Tests: run.

### DIFF
--- a/deepwell/migrations/20220906103252_deepwell.sql
+++ b/deepwell/migrations/20220906103252_deepwell.sql
@@ -202,7 +202,6 @@ CREATE TABLE relation (
     deleted_by BIGINT REFERENCES "user"(user_id),
     deleted_at TIMESTAMP WITH TIME ZONE,
 
-    UNIQUE (relation_type, dest_type, dest_id, from_type, from_id, overwritten_at, deleted_at),
     CHECK ((overwritten_by IS NULL) = (overwritten_at IS NULL)),  -- ensure overwritten field consistency
     CHECK ((deleted_by IS NULL) = (deleted_at IS NULL)),          -- ensure deleted field consistency
     CHECK (
@@ -210,6 +209,24 @@ CREATE TABLE relation (
         ((overwritten_by IS NULL) != (deleted_at IS NULL))        -- or they are overwritten XOR deleted
     )
 );
+
+CREATE UNIQUE INDEX relation_unique_general_active
+    ON relation (relation_type, dest_type, dest_id, from_type, from_id, overwritten_at, deleted_at)
+    WHERE relation_type <> 'page-attribution';
+
+CREATE UNIQUE INDEX relation_unique_page_attribution_active
+    ON relation (
+        relation_type,
+        dest_type,
+        dest_id,
+        from_type,
+        from_id,
+        (metadata ->> 'attribution_type'),
+        (metadata ->> 'attribution_date'),
+        coalesce(overwritten_at, 'infinity'),
+        coalesce(deleted_at, 'infinity')
+    )
+    WHERE relation_type = 'page-attribution';
 
 --
 -- Session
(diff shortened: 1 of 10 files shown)


deepwell/migrations/20220906103252_deepwell.sql
```diff
@@ -211,6 +210,24 @@ CREATE TABLE relation (
...
+        dest_type,
+        dest_id,
+        from_type,
+        from_id,
+        (metadata ->> 'attribution_type'),
+        (metadata ->> 'attribution_date'),
+        coalesce(overwritten_at, 'infinity'),
+        coalesce(deleted_at, 'infinity')
+    )
+    WHERE relation_type = 'page-attribution';
```
emmiegit, line 229: Interesting, thinking on this you are right that a different unique constraint is needed. I'm wondering though, presently no other relations have different uniqueness constraints, do you think that would change or is something we should be thinking about?

In terms of future migrations, I guess we drop and re-create `relation_unique_general_active` where it has a `relation_type NOT IN (...)` with the new type(s). So that should be fine.

deepwell/src/services/relation/page_attribution.rs
```diff
@@ -0,0 +1,134 @@
...
+    Rewrite,
+    Translator,
+    Maintainer,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+pub struct PageAttributionMetadata {
+    pub attribution_type: PageAttributionKind,
+    pub attribution_date: Date,
+}
```
emmiegit, line 37: Question: Any reason either of these structs couldn't be `Copy`?

deepwell/src/services/relation/page_attribution.rs
```diff
@@ -0,0 +1,134 @@
...
+    pub page_id: i64,
+    pub user_id: i64,
+    pub metadata: PageAttributionMetadata,
+    pub created_by: i64,
+}
+
+impl RelationService {
+    /// Creates a page attribution relation without overwriting other attributions
+    /// for the same page / user combination. This mirrors the legacy uniqueness of
+    /// (page_id, user_id, attribution_type, attribution_date).
```
emmiegit: When you say "legacy uniqueness", what do you mean? That the old SQL table used this unique constraint?

deepwell/src/services/relation/page_attribution.rs
```diff
@@ -0,0 +1,134 @@
...
+                    .add(relation::Column::Metadata.eq(metadata_json.clone()))
+                    .add(relation::Column::OverwrittenAt.is_null())
+                    .add(relation::Column::DeletedAt.is_null()),
+            )
+            .one(txn)
+            .await?
+            .is_some();
+
+        if already_exists {
+            return Ok(());
```
emmiegit, line 88: Maybe add a log line here and below noting the "already exists, doing nothing" and "checked for existence, going to insert" cases.

deepwell/migrations/20220906103252_deepwell.sql
Syxless: Only page-attribution needs a different uniqueness rule, so it has its own index. If future relations need special uniqueness, we can extend the `NOT IN` or create new indexes.

deepwell/src/services/relation/page_attribution.rs
Syxless: They can be `Copy`; I kept `Clone` for consistency but can switch if preferred.

deepwell/src/services/relation/page_attribution.rs
Syxless: Refers to the old SQL table's unique constraint `(page_id, user_id, attribution_type, attribution_date)`.

deepwell/src/services/relation/page_attribution.rs
emmiegit: Cool, could you add `Copy` in addition to `Clone`?

deepwell/src/services/relation/page_attribution.rs
emmiegit: I see. I think it could be rephrased to avoid misunderstanding - by using legacy like this it could sound like the unique constraint is legacy.

Maybe something like:

> This mirrors the unique constraint of (page_id, user_id, attribution_type, attribution_date) found when this was implemented as a separate SQL table.

deepwell/src/services/relation/page_attribution.rs
Syxless: Done.

deepwell/src/services/relation/page_attribution.rs
Syxless: Done.

emmiegit (approved):

PR #2608, opened 2025-12-10
WJ-1206: add page attribution service and RPC endpoints

Syxless:
Feature [WJ-1206](https://scuttle.atlassian.net/browse/WJ-1206).

Adds a page attribution relation service and wires JSON-RPC endpoints to manage it.

Service/API:
- List active attributions for a page via `page_attribution_get_page`.
- Idempotent create/update via `page_attribution_update` with overwrite tracking and uniqueness on page/user/kind/date.
- Soft-delete via `page_attribution_delete` with actor attribution.

Behavior:
- Converts relation models to DTOs with metadata and timestamps.
- Filters out overwritten/deleted relations.
- Ensures object types match the page→user relation contract.

Tests: ran through docker.

[WJ-1206]: https://scuttle.atlassian.net/browse/WJ-1206?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deepwell/src/endpoints/page_attribution.rs
+++ b/deepwell/src/endpoints/page_attribution.rs
@@ -58,7 +58,7 @@ pub async fn page_attribution_update(
 pub async fn page_attribution_delete(
     ctx: &ServiceContext<'_>,
     params: Params<'static>,
-) -> Result<Vec<PageAttribution>> {
+) -> Result<()> {
     let input: ClearPageAttributions<'_> = params.parse()?;
 
     info!(
--- a/deepwell/src/endpoints/page_attribution.rs
+++ b/deepwell/src/endpoints/page_attribution.rs
@@ -20,17 +20,20 @@
 
 use super::prelude::*;
 use crate::services::relation::{
-    GetPageAttributions, PageAttribution, RelationService, RemovePageAttribution,
-    UpdatePageAttribution,
+    ClearPageAttributions, GetPageAttributions, PageAttribution, RelationService,
+    SetPageAttributions,
 };
 
 pub async fn page_attribution_get_page(
     ctx: &ServiceContext<'_>,
     params: Params<'static>,
 ) -> Result<Vec<PageAttribution>> {
-    let input: GetPageAttributions = params.parse()?;
+    let input: GetPageAttributions<'_> = params.parse()?;
 
-    info!("Getting page attributions for page {}", input.page_id);
+    info!(
+        "Getting page attributions for page {:?} on site {}",
+        input.page, input.site_id,
+    );
 
     RelationService::get_page_attributions(ctx, input).await
 }
@@ -38,31 +41,30 @@ pub async fn page_attribution_get_page(
 pub async fn page_attribution_update(
     ctx: &ServiceContext<'_>,
     params: Params<'static>,
-) -> Result<PageAttribution> {
-    let input: UpdatePageAttribution = params.parse()?;
+) -> Result<Vec<PageAttribution>> {
+    let input: SetPageAttributions<'_> = params.parse()?;
 
     info!(
-        "Updating page attribution for page {} user {} ({:?} @ {}) (relation {:?})",
-        input.page_id,
-        input.user_id,
-        input.metadata.attribution_type,
-        input.metadata.attribution_date,
-        input.relation_id,
+        "Setting {} page attributions for page {:?} on site {} (requested by {})",
+        input.attributions.len(),
+        input.page,
+        input.site_id,
+        input.updated_by,
     );
 
-    RelationService::update_page_attribution(ctx, input).await
+    RelationService::set_page_attributions(ctx, input).await
 }
 
 pub async fn page_attribution_delete(
     ctx: &ServiceContext<'_>,
     params: Params<'static>,
-) -> Result<PageAttribution> {
-    let input: RemovePageAttribution = params.parse()?;
+) -> Result<Vec<PageAttribution>> {
+    let input: ClearPageAttributions<'_> = params.parse()?;
 
     info!(
-        "Deleting page attribution relation {} (requested by {})",
-        input.relation_id, input.removed_by,
+        "Clearing page attributions for page {:?} on site {} (requested by {})",
+        input.page, input.site_id, input.removed_by,
     );
 
-    RelationService::remove_page_attribution(ctx, input).await
+    RelationService::clear_page_attributions(ctx, input).await
 }
--- a/deepwell/src/endpoints/page_attribution.rs
+++ b/deepwell/src/endpoints/page_attribution.rs
@@ -1,0 +1,68 @@
+/*
+ * endpoints/page_attribution.rs
+ *
+ * DEEPWELL - Wikijump API provider and database manager
+ * Copyright (C) 2019-2025 Wikijump Team
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+use super::prelude::*;
+use crate::services::relation::{
+    GetPageAttributions, PageAttribution, RelationService, RemovePageAttribution,
+    UpdatePageAttribution,
+};
+
+pub async fn page_attribution_get_page(
+    ctx: &ServiceContext<'_>,
+    params: Params<'static>,
+) -> Result<Vec<PageAttribution>> {
+    let input: GetPageAttributions = params.parse()?;
+
+    info!("Getting page attributions for page {}", input.page_id);
+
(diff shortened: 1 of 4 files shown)


deepwell/src/services/relation/page_attribution.rs
```diff
@@ -108,37 +128,232 @@ impl RelationService {
...
+        deleted_by: model.deleted_by,
+        deleted_at: model.deleted_at,
+        metadata,
+    })
+}
+
+impl PageAttribution {
+    fn try_from_model(model: RelationModel) -> Result<Self> {
+        convert_model(model)
+    }
```
emmiegit, line 315: Question: Any reason this isn't using [`TryFrom`](https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html)?

deepwell/src/services/relation/page_attribution.rs
```diff
@@ -108,37 +128,232 @@ impl RelationService {
...
         page_id: i64,
         user_id: i64,
         metadata: &PageAttributionMetadata,
     ) -> Result<bool> {
-        let txn = ctx.transaction();
         let metadata_json = serde_json::to_value(metadata)?;
-        let exists = Relation::find()
+        find_page_attribution(ctx, page_id, user_id, &metadata_json)
+            .await
+            .map(|relation| relation.is_some())
```
emmiegit, line 156: I think you can just do `.filter_map(|relation| relation)` here.

deepwell/src/services/relation/page_attribution.rs
```diff
@@ -108,37 +128,232 @@ impl RelationService {
...
-                    .add(relation::Column::FromId.eq(user_id))
-                    .add(relation::Column::Metadata.eq(metadata_json))
+                    .add(relation::Column::DestId.eq(page_id))
                     .add(relation::Column::OverwrittenAt.is_null())
                     .add(relation::Column::DeletedAt.is_null()),
             )
-            .one(txn)
-            .await?
-            .is_some();
+            .order_by_asc(relation::Column::CreatedAt)
```
emmiegit: I think that, for attributions, we should have a more apparent-to-end-users order that we return them in. I'm thinking `(date DESC, attribution_type, user_name ASC)`?

So for a set of attributions, you get an ordering like:
```
2025-01-02 | TRANSLATOR | Dan (translator of current version)
2025-01-01 | REWRITE | Alice (co-rewriter)
2025-01-01 | REWRITE | Bob (co-rewriter)
2025-01-01 | REWRITE | Charlie (co-rewriter)
2009-01-05 | TRANSLATOR | Edgar (translator of original version)
2008-02-02 | AUTHOR | Francis (original author)
```

Unless you have other thoughts. The idea being that the most relevant information (the current version of the page) is at the top, and prior history is later.

However, I don't know if there's a clean way of specifying this in SeaORM since it involves JSON fields? So maybe don't have an `.order_by` here and then sort in code?

Any thoughts on what you think makes the most sense here?

deepwell/src/services/relation/page_attribution.rs
```diff
@@ -108,37 +128,232 @@ impl RelationService {
             ..Default::default()
         };
 
-        model.insert(txn).await?;
-        Ok(())
+        let model = model.insert(ctx.transaction()).await?;
```
emmiegit: General style note, even if there's only one use of `txn`, I have a `let txn = ctx.transaction();` on a separate line for consistency with SQL execute calls like `.insert()`.

deepwell/src/endpoints/page_attribution.rs
```diff
@@ -0,0 +1,68 @@
...
+) -> Result<PageAttribution> {
+    let input: RemovePageAttribution = params.parse()?;
+
+    info!(
+        "Deleting page attribution relation {} (requested by {})",
+        input.relation_id, input.removed_by,
+    );
+
+    RelationService::remove_page_attribution(ctx, input).await
+}
```
emmiegit, line 70: I think in terms of the calling pattern for page attribution, one API call should update a set of attributions for a page. That is, it submits a list of entries associated with one page. For example: (data in pseudocode)

```yaml
page_id: 12345
updater_user_id: 1005

attributions:
   attrib_type: rewrite
   user_id: 2000
   date: 2020-01-06

   attrib_type: rewrite
   user_id: 2001
   date: 2020-01-06

   attrib_type: author
   user_id: 1000
   date: 2010-01-05

   attrib_type: maintainer
   user_id: 500
   date: null
```

Which then deletes any prior attribution entries for page 12345 and inserts the above. Then `page_attribution_delete()` is the same except it doesn't insert anything after.

deepwell/src/services/relation/page_attribution.rs
```diff
@@ -44,12 +61,31 @@ pub struct CreatePageAttribution {
...
+    pub relation_id: Option<i64>,
+    pub page_id: i64,
+    pub user_id: i64,
+    pub metadata: PageAttributionMetadata,
+    pub created_by: i64,
+}
+
+#[derive(Deserialize, Debug, Copy, Clone)]
+pub struct GetPageAttributions {
+    pub page_id: i64,
```
emmiegit: FYI we have a concept of the [`Reference`](https://github.com/scpwiki/wikijump/blob/develop/deepwell/src/types/reference.rs), which is the ID-or-slug that is used for things like pages and users. I think it makes sense to use for the fetch method since it's just the page.

Speaking of, this and the other structs should require a `site_id`. If it doesn't match the `page_id` in question then the request should fail with `ServiceError::PageNotFound`. (This is generally done by [`PageService::get()`](https://github.com/scpwiki/wikijump/blob/fae2bcf9033ec8e878c53c5bf7fe2ff0a332b695/deepwell/src/services/page/service.rs#L717) or [`PageService::get_id()`](https://github.com/scpwiki/wikijump/blob/fae2bcf9033ec8e878c53c5bf7fe2ff0a332b695/deepwell/src/services/page/service.rs#L782), not you directly. Since relations don't store site ID as its own thing, I suggest using `PageService::get()` so we're always checking the site ID.)

deepwell/src/services/relation/page_attribution.rs
Syxless: Added` impl TryFrom<RelationModel> for PageAttribution` and now use it in the service.

deepwell/src/services/relation/page_attribution.rs
Syxless: the existence check now uses the dedicated `find_page_attribution` helper and `.is_some()`, so no manual option mapping remains.

deepwell/src/services/relation/page_attribution.rs
Syxless: we now sort in code by `attribution_date` DESC, then `attribution_type`, then `user_id` after fetching active relations (since DB ordering on JSON fields isn’t clean).

deepwell/src/services/relation/page_attribution.rs
Syxless: creation now binds `let txn = ctx.transaction();` before insert for consistency.

deepwell/src/endpoints/page_attribution.rs
Syxless: `page_attribution_update` now accepts site_id + page Reference and a list of attributions, clears existing active ones for that page, inserts the new set (deduped), and returns the sorted list. `page_attribution_delete` clears all attributions for the page using the site/page Reference and removed_by.

deepwell/src/services/relation/page_attribution.rs
```diff
@@ -44,12 +64,38 @@ pub struct CreatePageAttribution {
     pub created_by: i64,
 }
 
+#[derive(Deserialize, Debug, Clone)]
+pub struct PageAttributionInput {
+    pub user_id: i64,
+    pub metadata: PageAttributionMetadata,
+}
```
emmiegit, line 71: Maybe better to rename to like `PageAttributionEntry`?

deepwell/src/services/relation/page_attribution.rs
```diff
@@ -108,37 +139,204 @@ impl RelationService {
...
+        };
+
+        Relation::update_many()
+            .set(delete_model)
+            .filter(page_attributions_condition(page.page_id))
+            .exec(txn)
+            .await?;
 
-        Ok(exists)
+        Ok(Vec::new())
```
emmiegit: If this always returns an empty vec, why not change the return type to `()`?

deepwell/src/services/relation/page_attribution.rs
```diff
@@ -108,37 +139,204 @@ impl RelationService {
...
     }
 }
+
+fn page_attribution_condition(
```
emmiegit, line 258: Maybe add a brief doc to each of these to help readers differentiate. Something like
```rs
/// Builds a `Condition` for querying one specific page attribution entry.
fn page_attribution_condition(...)

/// Builds a `Connection` for all page attribution entries for a page.
fn page_attributions_condition(...)
```

emmiegit (approved):
Looks good, just a few minor comments.

deepwell/src/services/relation/page_attribution.rs
Syxless: Renamed `PageAttributionInput` -> `PageAttributionEntry`

deepwell/src/services/relation/page_attribution.rs
Syxless: Done.

deepwell/src/services/relation/page_attribution.rs
Syxless: Done.

PR #2610, opened 2025-12-11
WJ-1341: include page attributions in page view

Syxless:
Fixes WJ-1341.

Add page attribution data to page views so Framerail can display authorship.

Changes:
- Extend `GetPageViewOutput::Found` with `attributions: Vec<PageAttribution>`.
- `ViewService::page` fetches attributions for the found page (site_id + page reference); Missing/Permissions cases unchanged.

Tests: ran.

### DIFF
--- a/deepwell/src/services/view/service.rs
+++ b/deepwell/src/services/view/service.rs
@@ -124,6 +124,7 @@ impl ViewService {
             Found {
                 page: PageModel,
                 page_revision: PageRevisionModel,
+                attributions: Vec<PageAttribution>,
             },
             Missing,
             Private,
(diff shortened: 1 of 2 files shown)


deepwell/src/services/view/structs.rs
```diff
@@ -64,6 +64,7 @@ pub enum GetPageViewOutput {
         options: PageOptions,
         page: PageModel,
         page_revision: PageRevisionModel,
+        attributions: Vec<crate::services::relation::PageAttribution>,
```
emmiegit: Can you add a `use` for this at the top of this file?

deepwell/src/services/view/service.rs
```diff
@@ -256,6 +259,20 @@ impl ViewService {
...
+                    GetPageAttributions {
+                        site_id: page.site_id,
+                        page: Reference::Id(page.page_id),
+                    },
+                )
+                .await?,
+            ),
+            _ => None,
+        };
+
```
emmiegit: I think this assignment of `attributions` is incorrect. It only has a value when `PageStatus::Found`, so either it should be in that branch of `match status`, or my preferred location, adding `attributions` as a field to `PageStatus::Found` itself.

So I think the right place to fetch `attributions` is [around in the `if` block starting at line 164](https://github.com/scpwiki/wikijump/blob/3a2477a93c097f237e2f67b1331cfceade7f242e/deepwell/src/services/view/service.rs#L164), and then `attributions` should be a field in `PageStatus::Found` which is then just moved in the `match status` block.

emmiegit (approved):

PR #2623, opened 2025-12-27
[WJ-1337] Add navigation page support

emmiegit:
This PR adds navigation pages as a setting for sites (per-category) and compiles / exposes that information to the page viewer response. These are stored in `page_revision` for regular page views, and can be retrieved for missing and error page views.

This also updates the outdater to rerender pages which use a particular nav page. Specifically, there are two rerender modes now, full (what you normally think of, basically processing changes from an edit), and nav-only (only updating the top bar and side bar, if they exist).

Fields in the JSON:
image


deepwell/src/services/outdate.rs
```diff
@@ -145,11 +144,13 @@ impl OutdateService {
...
         page_slug: &str,
-        depth: u32,
+        depth: RerenderDepth,
     ) -> Result<()> {
+        let config = ctx.config();
+
         // If a template page has been updated,
         // we need to recompile everything in that category.
-        if page_slug == "_template" {
+        if page_slug == config.special_page_template {
```
Zokhoi, line 153: Note: change this when special -> underscore rename PR is merged

Zokhoi (approved):

PR #2633, opened 2026-01-02
Adjust deepwell health check delays to account for bootstrap period

Zokhoi:
Make use of the docker options `--start-period` and `--start-interval` to prevent health check fails during startup period.

This solves the problem mentioned in #2472.

### DIFF
--- a/install/local/docker-compose.yaml
+++ b/install/local/docker-compose.yaml
@@ -71,9 +71,11 @@ services:
     restart: always
     healthcheck:
       test: ["CMD", "wikijump-health-check"]
-      interval: 600s
+      interval: 120s
       timeout: 2s
       retries: 3
+      start_period: 600s
+      start_interval: 30s
     depends_on:
       cache:
         condition: service_healthy
--- a/install/local/docker-compose.yaml
+++ b/install/local/docker-compose.yaml
@@ -75,7 +75,7 @@ services:
       timeout: 2s
       retries: 3
       start_period: 600s
-      start_interval: 30s
+      start_interval: 10s
     depends_on:
       cache:
         condition: service_healthy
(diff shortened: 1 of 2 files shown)


install/local/docker-compose.yaml
```diff
@@ -71,9 +71,11 @@ services:
     restart: always
     healthcheck:
       test: ["CMD", "wikijump-health-check"]
-      interval: 600s
+      interval: 120s
       timeout: 2s
       retries: 3
+      start_period: 600s
+      start_interval: 30s
```
emmiegit: The health check is just a curl request, I think a shorter interval (and thus faster "ready" status) makes sense here. Also, 10s is the health check interval for the other services.

```suggestion
      start_interval: 10s
```

install/local/docker-compose.yaml
```diff
@@ -71,9 +71,11 @@ services:
     restart: always
     healthcheck:
       test: ["CMD", "wikijump-health-check"]
-      interval: 600s
+      interval: 120s
```
emmiegit, line 74:
```suggestion
      interval: 30s
```

emmiegit (approved):
Hooray! 🎉

install/local/docker-compose.yaml
Zokhoi: The container is considered ready when a health check succeeds during the start period, so having a shorter interval here does not affect the readyness. The interval was at 120s before #2472, and framerail is also at 120s interval.

emmiegit: Keeping it the same as framerail seems fine. I was just thinking of reducing it to match some of the other services, though it's true that this is more similar to framerail than postgres or redis.

PR #2638, opened 2026-01-05
Set up integration tests for DEEPWELL

emmiegit:
This is a long-overdue change, it implements proper integration tests for DEEPWELL as an internal API service.

This uses `tests/`, the integration test directory for Rust, to create separate binaries which start an instance which connects to a local resources (e.g. S3, database, etc.) in order to perform tests using DEEPWELL endpoints as the point of entry. In order to do this, we change the crate to be library-first, with the primary contents in `lib.rs` and `main.rs` only calling the entrypoint function. This library version of DEEPWELL is what test runners use when running non-server instances for their integration tests.

Integration tests should have no effect on the local database, since any changes are reverted when the test ends (whether success or failure). However, any objects added to the S3 bucket or keys changed in Redis will remain.

I have also added a `mock_mailcheck` configuration field, enabled on local and in test environments, which gets consistent responses for testing and also doesn't rely on that third-party service (which among other issues, has a strict rate-limit, which could cause tests to spontaneously fail).

To start us off, this adds a few integration tests. There are some basic ones for the simple endpoints (health, misc, and translations), as well as basic tests for page and user operations.

I also bump the crate version.

jira: WJ-1377


deepwell/src/start.rs
```diff
@@ -0,0 +1,102 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+//! Entrypoint to the server-based execution mode of deepwell.
```
emmiegit, line 21: This file is basically just what `main.rs` used to be.

deepwell/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+//! Library version of the deepwell crate.
+//!
+//! This allows you to customize how the server is run, and work with
+//! the crate's types and systems.
```
emmiegit, line 24: We now export `deepwell` as a library and a binary, where the binary just runs `deepwell::start()`. This is necessary since integration tests work by importing the main crate as a dependency.

deepwell/src/services/email/service.rs
```diff
@@ -185,4 +187,89 @@ impl EmailService {
...
+            None => (email, "", EmailType::Invalid),
+            Some((user, domain)) => {
+                let email_type = match domain {
+                    "gmial.com" => EmailType::DidYouMean("gmail.com"),
+                    "yaho.com" => EmailType::DidYouMean("yahoo.com"),
+                    "disposable.com" => EmailType::Disposable,
+                    "forwarding.com" => EmailType::Relay,
+                    "private.me" => EmailType::Private,
+                    "spam.xxx" => EmailType::Spam,
+                    "invalid.com" => EmailType::Invalid,
```
emmiegit, line 242: Mocking MailCheck to avoid ratelimiting and enable consistent testing results.

deepwell/tests/locale.rs
```diff
@@ -0,0 +1,303 @@
...
+            "locales": ["zh", "zh_Hans", "ko"],
+            "messages": {
+                "alt-title": {},
+                "close": {},
+            },
+        }),
+    );
+    assert_eq!(output.len(), 2);
+    assert!(output.contains_key("alt-title"));
+    assert!(output.contains_key("close"));
```
emmiegit, line 244: Fundamentally, integration tests involve calling a series of deepwell endpoints by passing in the raw JSON (to ensure the deserialization is working as expected for framerail), and they receive Rust objects in response (to make assertions easier and not having to serialize-then-deserialize those every time).

deepwell/tests/page.rs
```diff
@@ -0,0 +1,329 @@
...
+
+use deepwell::constants::ADMIN_USER_ID;
+use deepwell::endpoints;
+use deepwell::error::prelude::*;
+use deepwell::models::sea_orm_active_enums::PageRevisionType;
+use deepwell::services::ServiceContext;
+use serde_json::json;
+
+#[tokio::test]
+async fn basic_edit() {
```
emmiegit, line 31: An example of a mutable integration test. It creates a page, checks the data is as set, modifies it, checks it again, etc.

This only uses the JSONRPC methods - we could use deepwell service functions to verify things, but we shouldn't be modifying data that way. The point is to pretend we are framerail / the end user, and this series of actions have taken place, which naturally must only happen through the JSONRPC API.

deepwell/src/config/mod.rs
```diff
@@ -37,8 +37,13 @@ pub struct SetupConfig {
...
 impl SetupConfig {
-    pub fn load() -> Self {
-        run_runtime_action();
+    pub async fn load() -> Self {
+        run_runtime_action().await;
+        Self::load2()
+    }
+
+    /// Internal version of `load()` that doesn't do runtime actions.
+    pub(crate) fn load2() -> Self {
```
loop-index: [nit] Can this be named something more descriptive like `load_configs`?

deepwell/tests/common/runner.rs
```diff
@@ -0,0 +1,123 @@
...
+                    "Call to method '{}' succeeded when it should have failed\n{:?}",
+                    stringify!($endpoint),
+                    result,
+                );
+            }
+        }
+    };
+}
+
+macro_rules! cleanup {
```
loop-index: While we can do our due diligence to make sure this is always called at the end of the test, I feel a bit unsafe in making this call manual. Maybe we can wrap the state+txn+ctx in a struct, and implement a Drop trait that calls the cleanup operations once ctx goes out of scope?

deepwell/tests/common/assert.rs
```diff
@@ -0,0 +1,131 @@
...
+        assert!(
+            extracted.is_some(),
+            "Cannot find error within trace matching '{}':\n{:?}",
+            stringify!($pattern),
+            exn_error,
+        );
+    }};
+}
+
+/// Asserts that there are no erroras within the `Exn<Error>` matching the given pattern.
```
loop-index:
```suggestion
/// Asserts that there are no errors within the `Exn<Error>` matching the given pattern.
```

deepwell/tests/common/runner.rs
```diff
@@ -0,0 +1,123 @@
...
+        // Explicitly drop all these bindings to prevent reuse later
+        mem::drop($ctx);
+
+        // We always rollback since we want the database state to be the same for each test
+        $txn.rollback()
+            .await
+            .expect("Unable to roll back transaction");
+
+        mem::drop($state);
+    }};
```
loop-index: Will there be additional rollback operations for S3/redis as well? It's fine to leave it out of this PR's scope, but maybe a TODO will suffice.

loop-index (commented):
Left some comments, generally looks like a robust framework for implementing tests. Thanks for doing this 🙌

deepwell/tests/common/runner.rs
emmiegit: This is something I struggled with for a while. I tried to make the system take in function or callback, but I could not get the ownership/borrowing right, so I let it as this.

However, you're right, I had not considered Drop-based cleanup. I think that's the best way to do this.

deepwell/tests/common/runner.rs
emmiegit: That is more difficult to implement; but you're right a TODO would be good.

loop-index:
LGTM, but I don't have approve perms.

loop-index (approved):

PR #2645, opened 2026-01-09
Rename "special error" -> "basic error"

emmiegit:
Companion to #2643. I've noticed I'm using "special XXX" in too many places, and this PR renames "special errors" to "basic errors" instead. Being a more unique term, it should make distinguishing concepts easier in the future.


locales/fluent/basic-error/zh_Hans.ftl
```diff
@@ -1,6 +1,6 @@
-### 特殊错误 HTML 模板
+### 根源错误 HTML 模板
```
Zokhoi:
```suggestion
### 基本错误 HTML 模板
```

emmiegit: Thanks!

Zokhoi (approved):

PR #2653, opened 2026-01-17
Rename "special page" → "blueprint page"

emmiegit:
As part of my general efforts using less ambiguous names like "special X", I'm renaming "special pages" to "blueprints". These are the pages like `_template`, `_404`, etc.

I also rename `deepwell.toml` → `config.toml` in `install/` for consistency and better tab completion.

Thanks @Zokhoi for the name suggestion.


Zokhoi:
I don't think "underscore page" conveys the purpose of these pages well. These are [template pages](https://www.wikidot.com/doc:templates), but as the word "template" is taken for the main template, I think calling them "blueprint", "format" or something along those lines might be better.

emmiegit:
Hmm, yeah I think you're right. "Underscore page" is descriptive inasmuch as you can apply a label to the concept when *you're already familiar with it from Wikidot*, but not helpful for conveying it to someone for the first time.

I think "blueprint page" is a much better term, I'll switch to that.

emmiegit:
Renamed underscore -> blueprint.

locales/fluent/wiki-page/zh_Hans.ftl
```diff
@@ -104,7 +104,7 @@ wiki-page-file-restore = 恢复
 ### 维基页面父页面
 
 wiki-page-parent = 父页面
-### 特殊页面
+### 藍圖页面
```
Zokhoi:
```suggestion
### 蓝图页面
```
`藍圖` is Traditional Chinese, though this is the comments and doesn't matter as much

Zokhoi (approved):

locales/fluent/wiki-page/en.ftl
```diff
@@ -104,7 +104,7 @@ wiki-page-file-restore = Restore
 ### Wiki page parents
 
 wiki-page-parent = Page parents
-### Special Page fallback strings
+### Underscore Page fallback strings
```
Zokhoi: Ooh this also needs to be updated

locales/fluent/wiki-page/zh_Hans.ftl
emmiegit: ...For some reason I looked at 页 and thought "traditional" even though it's obviously not

locales/fluent/wiki-page/en.ftl
emmiegit: Also fixed (rebase)

PR #2657, opened 2026-01-18
Display navigation bars in framerail

Zokhoi:
Display the navigation bars in framerail.

Reference #2623 for deepwell

### DIFF
--- a/framerail/package.json
+++ b/framerail/package.json
@@ -1,6 +1,6 @@
 {
   "name": "framerail",
-  "version": "2026.1.8",
+  "version": "2026.1.18",
   "description": "Framerail - Wikijump frontend service",
   "repository": "https://github.com/scpwiki/wikijump/tree/develop/framerail",
   "license": "AGPL-3.0-or-later",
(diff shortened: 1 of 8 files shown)


framerail/src/lib/server/load/logout.ts
```diff
@@ -27,7 +27,7 @@ export async function loadLogoutPage(request, cookies) {
     "logout.toast": {}
   }
 
-  const translated = await translate(locales, translateKeys)
+  const translated = await translate(locales, translateKeys, defaults.translateStripKeys)
```
emmiegit: Would it make sense to have `translate()`'s default for the last arg be `defaults.translateStripKeys` instead of `[]`? And then it's only overridden if a specific key needs to have its control characters stripped, rather than in all uses.

emmiegit (approved):

framerail/src/lib/server/load/logout.ts
Zokhoi: That will be simpler yeah

PR #2682, opened 2026-02-14
Update Caddyfile every 5 minutes

emmiegit:
Five minutes is a much more reasonable time to wait for domain settings to become live. I think this is a safe interval because:
* Caddyfile generation is not a particularly expensive process, it's essentially doing three queries and then generating a string.
* Caddy detects if the Caddyfile is different from the current one, and if so, takes no action.
* Even when Caddy does update its configuration, it does so very quickly and without disruption to ongoing operations.

### DIFF
--- a/install/common/caddy/crontab
+++ b/install/common/caddy/crontab
@@ -1,2 +1,2 @@
-# Hourly regeneration of Caddyfile from DEEPWELL
+# Regeneration of Caddyfile from DEEPWELL every 5 minutes
 */5 * * * * /usr/local/bin/wikijump-update-caddyfile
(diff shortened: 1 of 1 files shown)


install/common/caddy/crontab
```diff
@@ -1,2 +1,2 @@
 # Hourly regeneration of Caddyfile from DEEPWELL
```
Zokhoi: Also update the explanation?
```suggestion
# Regeneration of Caddyfile from DEEPWELL every 5 minutes
```

Zokhoi (approved):

install/common/caddy/crontab
emmiegit: Good catch!

PR #2701, opened 2026-02-22
Locale and other UX improvements

Zokhoi:
Improves user locale experience and other minor stuff.

* Respects locales set in user profile when displaying page options
* Reorganizes locale messages in fluent
* Message for user trying to view self profile when not logged in, instead of "user not exist"
* Fix non-existant locale message causing framerail to error out
* Fix user session cookie not expiring
* Add call trace to error popup


framerail/src/lib/popup/error.svelte
```diff
@@ -94,7 +94,7 @@
       </div>
       {#if $showErrorPopup.data}
         <div id="model-message-extra" class="modal-message-extra">
-          {$showErrorPopup.data}
+          {$showErrorPopup.data.call_trace}
```
emmiegit, line 97: I think this is useful to include, though the issue with deepwell traces is that we cannot really localize them. It's helpful for developers but is less actionable for end users.

(The point of the integer error codes + error arguments are to eventually enable proper localization of errors)

emmiegit (approved):
Very nice!

framerail/src/lib/popup/error.svelte
emmiegit: Not a blocker for the PR, just a note for the future.

Zokhoi: I have some reservations about showing the full trace to users, even if they were localized, but at this point in development it is very useful.

I'd think the user only need to see the most surface error (localized), and have the option to send the full error trace to the platform as an error report.

emmiegit: Ah, I misunderstood what this code meant. You're just including this here for now, to enable debugging, and weren't planning on exposing it in production. In that case, we're on the same page.

PR #2702, opened 2026-02-22
Always record call_trace

emmiegit:
As noted in [discussion on #2701](https://github.com/scpwiki/wikijump/pull/2701#discussion_r2837833130), the `call_trace` field will be retained internally only, and so it is safe to record the data even in the case of authentication issues. (This is because only developers will be reviewing the logs and it cannot be used by attackers to determine internal authentication state).


Zokhoi:
Hmm, I'm not familiar with the relevant code in deepwell, but I believe right now we don't have code to block showing `call_trace` in production, so maybe reserve this for later?

emmiegit:
That's fine, we aren't in any rush to merge this.

PR #2031, opened 2024-08-15
Locales types generator

Moonded:
Ok, so after fixing my git mistakes, here the PR for generating types based on fluent locale files.

edit: at this point im not sure if i actually fixed my git mistakes

PR #2715, opened 2026-02-24
Update fluent locale types

Zokhoi:
Update type generator and typings for locale strings from fluent.

Previously the translation string attributes were not generated, and all the locale strings were read-only. The strings will need to be modifiable, as some variables may be populated in framerail after translation.

Reference #2031

### DIFF
--- a/framerail/src/types/index.ts
+++ b/framerail/src/types/index.ts
@@ -1,4 +1,13 @@
 export interface Locales {
+  readonly "site-info": string
+  readonly "site-info.name": string
+  readonly "site-info.slug": string
+  readonly "site-info.tagline": string
+  readonly "site-info.description": string
+  readonly "site-info.default-page": string
+  readonly "site-info.locale": string
+  readonly "site-info.layout": string
+  readonly "admin-unauthorized": string
   readonly "-service-name": string
   readonly "message-loading": string
   readonly "goto-home": string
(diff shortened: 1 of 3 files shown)


framerail/src/types/index.ts
```diff
@@ -1,127 +1,398 @@
...
+  "wiki-page-file-revision-type.undo": string
+  "wiki-page-file-restore": string
+  "wiki-page-file-restore.new-page": string
+  "wiki-page-file-restore.new-name": string
+  "wiki-page-parent": string
+  "wiki-page-missing": string
+  "wiki-page-private": string
+  "wiki-page-banned": string
+  "wiki-page-no-render": string
 }
```
emmiegit, line 398: How do we generate this file?

locales/typed/index.ts
```diff
@@ -28,14 +31,13 @@ fs.readdirSync(input).forEach((folder) => {
...
+    combined[entry] = "string";
   });
 });
 
 const types = JSON.stringify(combined, null, 2)
-  .replace(/"readonly /g, 'readonly "')
   .replace(/"/g, "")
   .replace(/,/g, "")
-  .replace(/(?=\S*['-])([a-zA-Z0-9'-]+)/g, '"$1"');
+  .replace(/(?=\S*['-._])([a-zA-Z0-9'-._]+)/g, '"$1"');
```
emmiegit: Need to escape `.` or it will match all chars:
```suggestion
  .replace(/(?=\S*['-\._])([a-zA-Z0-9'-\._]+)/g, '"$1"');
```

framerail/src/types/index.ts
Zokhoi: In `locales/typed/` run `pnpm dev`

locales/typed/index.ts
Zokhoi: In square brackets, the dot does not carry special meaning and will be treated as the literal character, but there is no harm in escaping it.

However the dash does have special meaning and will need to be escaped to match the character.

emmiegit (approved):

PR #2734, opened 2026-03-05
Implement user registration page

hoah2333:
Note that I also cleaned up some deprecated configs in `svelte.config.js`.

### DIFF
--- a/framerail/src/routes/[x+2d]/register/+server.ts
+++ b/framerail/src/routes/[x+2d]/register/+server.ts
@@ -1,0 +1,38 @@
+import { userCreate } from "$lib/server/deepwell/user.js"
+import { UserType } from "$lib/types.js"
+
+export async function POST(event) {
+  const data = await event.request.formData()
+
+  const userAgent = event.request.headers.get("User-Agent")
+  const ipAddress = event.getClientAddress()
+
+  const username = data.get("username")?.toString()
+  const email = data.get("email")?.toString()
+  const password = data.get("password")?.toString()
+
+  try {
+    if (!username) throw new Error("Username is required")
+    if (!email) throw new Error("Email is required")
+    if (!password) throw new Error("Password is required")
+
+    const res = await userCreate(
+      UserType.Regular,
+      username,
+      email,
+      ["en"],
+      password,
+      ipAddress
+    )
+
+    return new Response(JSON.stringify(res))
+  } catch (error) {
+    return new Response(
+      JSON.stringify({
+        message: error.message,
+        code: error.code,
+        data: error.data
+      })
+    )
+  }
+}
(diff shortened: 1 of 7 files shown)


Zokhoi (approved):

framerail/src/routes/[x+2d]/register/+server.ts
```diff
@@ -0,0 +1,38 @@
...
+  try {
+    if (!username) throw new Error("Username is required")
+    if (!email) throw new Error("Email is required")
+    if (!password) throw new Error("Password is required")
+
+    const res = await userCreate(
+      UserType.Regular,
+      username,
+      email,
+      ["en"],
```
Comment (emmiegit) [inline framerail/src/routes/[x+2d]/register/+server.ts:23]: Shouldn't locales come from the input, rather than being English?

emmiegit (approved):

emmiegit (approved):

framerail/src/routes/[x+2d]/register/+server.ts
Comment (Zokhoi) [reply framerail/src/routes/[x+2d]/register/+server.ts:23]: We will need an api that returns the currently available locales for this option in the register form

Comment (emmiegit) [reply framerail/src/routes/[x+2d]/register/+server.ts:23]: Is this something to add on the frontend side, or a thing I need to add?

Comment (Zokhoi) [reply framerail/src/routes/[x+2d]/register/+server.ts:23]: Just took a look, currently the deepwell locale check is pretty lax, only requiring a properly formatted locale string
So I guess auto-generating it along with the locale types is fine, no need for an api after all

Comment (hoah2333) [reply framerail/src/routes/[x+2d]/register/+server.ts:23]: That might be better handled on the frontend. We could use a hardcoded list to reduce network overhead, though we’ll need a plan for when new languages are added.

Comment (emmiegit) [reply framerail/src/routes/[x+2d]/register/+server.ts:23]: We could have a script that runs in CI that compares the list of languages to the `locales` dir.

Though, how would this work with regional locales? Say a user from Luxembourg who prefers French; presently the locale logic in deepwell returns `fr` (generic) if Luxembourg French specifically isn't available.

Comment (Zokhoi) [reply framerail/src/routes/[x+2d]/register/+server.ts:23]: I'm not sure what you mean by this question?
The deepwell locale checker already accepts regions, so no problem there.
In the frontend, the user gets to choose from a (generated) list of available locales. The frontend might allow user to enter their custom locale string, but if that happens then it needs to enforce having at least one existing locale to fall back to.

Comment (emmiegit) [reply framerail/src/routes/[x+2d]/register/+server.ts:23]: Oh, I misunderstood. I thought the idea was to have a list of languages, and my concerns were:
* Ensuring this list is in sync with available languages (can be done with CI)
* If this is a list of *locales*, not languages, then I was concerned about regions and related metadata.

PR #2737, opened 2026-03-06
[feat] Role and Permission system

loop-index:
Implemented a simple RBAC-style role/permission system. Took some heavy inspiration from Discord's role system.
- Permissions are resource:action pairs (ie, page:edit) shared across platform (assuming all sites have the same resource types and actions)
- Roles are scoped to site. Site comes with several default roles on creation.
- User can have multiple roles across multiple sites.
- Permission check: Does [user] have any [role] with [permission] to [act] on [resource]
- Implicit roles: Roles that are granted conditionally by the system and not persisted. Example: Guest role granted when user visit a site without membership, or Author role granted when user act on a page that they authored. Idea is to allow customization of permissions for these roles per-site.

Planned improvements:
- Role management UI
- Temporary role grants with expiration. Might be useful for temp bans, temp edit grant for 001 proposals, etc.
- Role hierarchy, where lower-ranking roles cannot affect/grant higher-ranking roles
- Caching to improve role/permission lookup
- Per page category permissions like Wikidot. Have not thought too much about this - one idea is to expand the permission schema to allow for subtypes (ie page-component:edit), but then permissions would need to be scoped per-site.

As a proof-of-concept, I've added a permission check for page edits. Currently only the admin account is allowed to edit pages.


deepwell/src/models/site.rs
```diff
@@ -9,11 +9,8 @@ use serde::{Deserialize, Serialize};
 pub struct Model {
     #[sea_orm(primary_key)]
     pub site_id: i64,
-    #[serde(with = "time::serde::rfc3339")]
     pub created_at: TimeDateTimeWithTimeZone,
-    #[serde(with = "time::serde::rfc3339::option")]
     pub updated_at: Option<TimeDateTimeWithTimeZone>,
-    #[serde(with = "time::serde::rfc3339::option")]
     pub deleted_at: Option<TimeDateTimeWithTimeZone>,
```
Zokhoi, line 17: These attributes are required for interfacing with framerail (and are manually added if I recall correctly), please add them back

framerail/src/routes/[slug]/[...extra]/page.svelte
```diff
@@ -20,7 +20,27 @@
...
+
+    const res = await fetch(`/${page.data.page.slug}/check-edit-permission`, {
+      method: "POST",
+      body: fdata
+    }).then((r) => r.json())
+
+    if (!res.can_edit) {
+      errorPopupState.current = {
+        state: true,
+        message: "You don't have permission to edit this page",
```
Comment (Zokhoi) [inline framerail/src/routes/[slug]/[...extra]/page.svelte]: Please attach `Untranslated` for hardcoded messages

deepwell/src/models/user_role.rs
```diff
@@ -0,0 +1,53 @@
...
+#[sea_orm(table_name = "user_role")]
+pub struct Model {
+    #[sea_orm(primary_key, auto_increment = false)]
+    pub user_id: i64,
+    #[sea_orm(primary_key, auto_increment = false)]
+    pub role_id: i64,
+    pub assigned_at: TimeDateTimeWithTimeZone,
+    pub assigned_by: i64,
+    pub expires_at: Option<TimeDateTimeWithTimeZone>,
+    pub deleted_at: Option<TimeDateTimeWithTimeZone>,
```
Zokhoi, line 20: Please add serde attributes for time

deepwell/src/models/role.rs
```diff
@@ -0,0 +1,67 @@
...
+    #[sea_orm(column_type = "Text")]
+    pub name: String,
+    #[sea_orm(column_type = "Text", nullable)]
+    pub description: Option<String>,
+    pub from_wikidot: bool,
+    pub implicit: bool,
+    pub is_system: bool,
+    pub level: i32,
+    pub created_at: TimeDateTimeWithTimeZone,
+    pub deleted_at: Option<TimeDateTimeWithTimeZone>,
```
Zokhoi, line 24: Please add serde attributes for time

deepwell/src/models/site.rs
loop-index: Pretty sure these are removed by the codegen CLI. I've cleaned up other files but must have missed this one. Will push a fix when I have the time

emmiegit: They were, unfortunately there are some things the codegen CLI can't do properly. I've made issues for those things, but until they're implemented we unfortunately just have to make some manual tweaks.

emmiegit:
Just a small note, temporary bans are implemented - this is done by adding a metadata property to the ban noting its expiration date.

deepwell/migrations/20220906103252_deepwell.sql
```diff
@@ -938,6 +934,74 @@ CREATE INDEX forum_post_parent_idx ON forum_post (parent_post_id);
...
+
+-- Lookup table for permissions. This is shared across all sites.
+CREATE TABLE permission (
+    permission_id BIGSERIAL PRIMARY KEY,
+    description TEXT NOT NULL,
+    resource_type TEXT NOT NULL,
+    action TEXT NOT NULL,
+
+    UNIQUE (resource_type, action)
+);
```
emmiegit: So the idea here is that permissions like `page:read` and `page:edit` will be in this table? I suppose that _actions_ would be implemented in code (e.g. `edit`) vs the whole permission?

I suppose the reason for that would be that some permissions may be more finely-tuned, like for page categories?

Also if it is platform-wide, keep in mind that permission descriptions need to be localizable, which wouldn't really work with a single `description` column. (This would be fine if they were per-site, since then it would site admins setting the description, which would then be localized/customized to their site's audience.)

Additionally, if we want to have the concepts of platform permissions and custom site permissions, we can add a nullable `site_id` column (`NULL` = platform, otherwise = for that site). Though I'm undecided about this given the localization issue.

deepwell/migrations/20220906103252_deepwell.sql
```diff
@@ -938,6 +934,74 @@ CREATE INDEX forum_post_parent_idx ON forum_post (parent_post_id);
...
+
+-- Roles in a site.
+CREATE TABLE role (
+    role_id BIGSERIAL PRIMARY KEY,
+
+    -- A role entry denotes a unique role in a site. A user maybe an admin in one site but not in another.
```
emmiegit: Wording suggesetion:
```suggestion
    -- Denotes a unique role in a site. A user may be an admin in one site but a regular site member in another.
```

deepwell/migrations/20220906103252_deepwell.sql
```diff
@@ -938,6 +934,74 @@ CREATE INDEX forum_post_parent_idx ON forum_post (parent_post_id);
...
+    site_id BIGINT NOT NULL REFERENCES site(site_id),
+    name TEXT NOT NULL,
+    description TEXT,
```
emmiegit:
```suggestion
    description TEXT NOT NULL,
```

This doesn't preclude an empty description if they don't fill it out, and avoids confusion between `''` and `NULL` as values.

deepwell/migrations/20220906103252_deepwell.sql
```diff
@@ -938,6 +934,74 @@ CREATE INDEX forum_post_parent_idx ON forum_post (parent_post_id);
...
+    from_wikidot BOOLEAN NOT NULL DEFAULT false,
+
+    -- Implicit roles are invisible to the user, granted by the system under specific conditions
+    -- i.e. Guest role granted when a non-member visits the site,
+    -- or Author role granted when a user interacts with a page they authored.
+    -- Implicit roles are visible to admins where they can select the role's permissions.
+    -- Implicit roles cannot be manually assigned..
```
emmiegit: Typo:
```suggestion
    -- Implicit roles cannot be manually assigned.
```

deepwell/migrations/20220906103252_deepwell.sql
```diff
@@ -938,6 +934,74 @@ CREATE INDEX forum_post_parent_idx ON forum_post (parent_post_id);
...
+    implicit BOOLEAN NOT NULL DEFAULT false,
+
+    -- System roles cannot be deleted (i.e. Admin)
+    is_system BOOLEAN NOT NULL DEFAULT false,
```
emmiegit: What system roles would you imagine existing? A site can _start out_ with a set of roles, but I would like sites to be able to modify role configurations and permissions flexibly. SCP-EN for instance would not have a "moderator" role akin to current Wikidot, but instead granting permissions to edit and lock pages depending on what team(s) they were on.

In my view, the only system/undeletable roles should reflect the most basic relations, which I think are just "guest" and "site member".

deepwell/migrations/20220906103252_deepwell.sql
```diff
@@ -938,6 +934,74 @@ CREATE INDEX forum_post_parent_idx ON forum_post (parent_post_id);
...
+
+    -- Rudimentary role hierarchy.
+    -- Roles with higher level are granted more permissions than roles with lower levels.
+    -- Users with role management permissions can only grant roles with lower levels and affect users of lower levels.
+    level INTEGER NOT NULL CHECK (level >= 0),
+
+    created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
```
emmiegit, line 985: As is conventional for other objects:
```suggestion
    created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
    updated_at TIMESTAMP WITH TIME ZONE,
```

deepwell/migrations/20220906103252_deepwell.sql
emmiegit: Thinking about this a bit more, I'm wondering about the relation here between "implicit" and "system" roles. My mind keeps going towards the notion of "virtual" roles which are like your description of implicit roles (assigned automatically by the system) but also having the "system" property of being undeletable. For instance, you could not remove the "site member" virtual role directly, as it instead reflects that the user is presently a site member - the way to get rid of it would be to revoke the user's membership.

deepwell/migrations/20220906103252_deepwell.sql
```diff
@@ -938,6 +934,74 @@ CREATE INDEX forum_post_parent_idx ON forum_post (parent_post_id);
...
+    -- Implicit roles cannot be manually assigned..
+    implicit BOOLEAN NOT NULL DEFAULT false,
+
+    -- System roles cannot be deleted (i.e. Admin)
+    is_system BOOLEAN NOT NULL DEFAULT false,
+
+    -- Rudimentary role hierarchy.
+    -- Roles with higher level are granted more permissions than roles with lower levels.
+    -- Users with role management permissions can only grant roles with lower levels and affect users of lower levels.
+    level INTEGER NOT NULL CHECK (level >= 0),
```
emmiegit, line 983: This system would certainly work, though I am wondering if there are hierarchical systems that would be better than simple numerical precedence, as this can cause some issues in Discord (which you have noted is an understandable inspiration for this work).

For instance, it could be a tree. There is one role that is the "root" that can manage any other role (or maybe multiple roots?), its children which can manage any roles under them, their children, etc. This way there isn't a strict total ranking, allowing some roles to be "equal" to each other because they manage different domains.

This solution also wouldn't be too complicated to represent in the database, just have a `parent_role_id` column that foreign keys to self.

deepwell/migrations/20220906103252_deepwell.sql
```diff
@@ -938,6 +934,74 @@ CREATE INDEX forum_post_parent_idx ON forum_post (parent_post_id);
...
+CREATE TABLE role_permission (
+    role_id BIGINT NOT NULL REFERENCES role(role_id),
+    permission_id BIGINT NOT NULL REFERENCES permission(permission_id),
+    PRIMARY KEY (role_id, permission_id)
+);
+
+-- User role assignments (many-to-many)
+CREATE TABLE user_role (
+    user_id BIGINT NOT NULL REFERENCES "user"(user_id),
+    role_id BIGINT NOT NULL REFERENCES role(role_id),
```
emmiegit, line 1009: So this is indirectly site-siloed since roles are per-site, though having a denormalized `site_id` column could be useful for querying all roles a user has (which I imagine would be a common operation)?

I guess I'm not sure how complex this query would be:
```sql
SELECT * FROM user_role
WHERE user_id = $1
AND role_id IN (
    SELECT role_id FROM role
    WHERE site_id = $2
)
```

deepwell/seeder/roles.json
```diff
@@ -0,0 +1,56 @@
...
+    {
+        "name": "guest",
+        "description": "Implicit role for all visitors (including non-members)",
+        "implicit": true,
+        "is-system": true,
+        "level": 1,
+        "permissions": [
+            "page:view"
+        ]
+    }
```
emmiegit, line 115: I think this needs a few more implicit roles:

* Clarify `guest` as specifically user ⊄ `site`. (To avoid rules applied on guests affecting site members, principle of least surprise and all)
* Add `user` which is all registered users (i.e. `guest` ∪ `site_member`).
* Add `unregistered` which is all logged-out users.
* Then finally, `everyone` which is all users (i.e. `user` ∪ `unregistered`).

(Names subject to change, for instance should `guest` be what I call "unregistered" here? Or just non-site-members? We should decide on clear terminology for all of these cases.)

deepwell/seeder/users.json
```diff
@@ -104,5 +104,22 @@
...
+        "password": "guestuser1",
+        "locales": ["en"],
+        "real_name": "Guest",
+        "gender": null,
+        "birthday": null,
+        "location": "Everywhere and nowhere",
+        "biography": "Wikijump guest user",
+        "user_page": null,
+        "aliases": [
+        ]
```
emmiegit, line 123: What is this user for?

deepwell/src/endpoints/page.rs
```diff
@@ -207,6 +208,26 @@ pub async fn page_edit(
...
+    params: Params<'static>,
+) -> Result<PageEditPermissionOutput> {
+    let input: PageEditPermission = parse!(params, Page);
+    info!(
+        "Checking edit permission for page {:?} in site ID {}",
+        input.page, input.site_id,
+    );
+
+    let can_edit =
+        PageService::check_user_permission(&ctx, input.site_id, input.user_id, "edit")
```
emmiegit: General code note, I am strongly in favor of using explicit enum types rather than string enum values. This eliminates the possibility of string typos or other inconsistencies silently causing bugs (since an issue with an enum type would cause a compilation failure).

deepwell/src/services/permission/service.rs
```diff
@@ -0,0 +1,322 @@
...
+
+        let make_error = || {
+            Error::new(
+                format!("failed to get permissions for role ID {}", role_id),
+                ErrorType::Role,
+            )
+        };
+
+        let role_permissions = RolePermission::find()
+            .filter(role_permission::Column::RoleId.eq(role_id))
```
emmiegit, line 149: We should `.order_by()` something consistent, say scope then action?

deepwell/src/services/permission/service.rs
```diff
@@ -0,0 +1,322 @@
...
+                Condition::all()
+                    .add(user_role::Column::UserId.eq(user_id))
+                    .add(role::Column::SiteId.eq(site_id)),
+            )
+            .all(txn)
+            .await
+            .or_raise(make_error)?
+            .into_iter()
+            .map(|ur| ur.role_id)
+            .collect();
```
emmiegit: It's possible for us to select only the role ID in the query, but we can worry about that later.

deepwell/src/services/permission/service.rs
```diff
@@ -0,0 +1,322 @@
...
+
+        let make_error =
+            || Error::new("failed to fetch permission", ErrorType::Permission);
+
+        let permission = match reference {
+            Reference::Id(id) => permission::Entity::find_by_id(id)
+                .one(txn)
+                .await
+                .or_raise(make_error)?,
+            _ => None,
```
emmiegit: Why are you accepting a `Reference` if there isn't an accepted meaning for the string variant of the enum? Wouldn't it be better to accept a `permission_id` only instead?

deepwell/src/services/role/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+#[allow(dead_code)]
+pub const ADMIN_ROLE_NAME: &str = "admin";
+#[allow(dead_code)]
+pub const MODERATOR_ROLE_NAME: &str = "moderator";
+#[allow(dead_code)]
+pub const MEMBER_ROLE_NAME: &str = "member";
+#[allow(dead_code)]
+pub const GUEST_ROLE_NAME: &str = "guest";
```
emmiegit: If these are for the default roles (mirroring Wikidot), then I think these values should be in seeder data, not code.

deepwell/src/services/role/service.rs
```diff
@@ -0,0 +1,372 @@
...
+            PermissionService::remove_permission_from_role(
+                ctx,
+                role.role_id,
+                *permission_id,
+            )
+            .await
+            .or_raise(make_error)?;
+        }
+
+        let new_permissions = if let Maybe::Set(permission_ids) = &permission_ids {
```
emmiegit: Nitpick: I think `match permission_ids` would look cleaner here.

deepwell/src/services/role/service.rs
```diff
@@ -0,0 +1,372 @@
...
+        AuditService::log(
+            &ctx,
+            ip_address,
+            AuditEvent::RoleDelete {
+                role_id: role.role_id,
+                deleting_user_id,
+            },
+        )
+        .await
+        .or_raise(make_error)?;
```
emmiegit, line 273: Question: When a role is deleted, is it removed from all users who have it?

If we don't remove it from users, then how do we ensure that deleted roles aren't coming up when we query all roles on a user?

deepwell/src/services/role/service.rs
```diff
@@ -0,0 +1,372 @@
...
+        let txn = ctx.transaction();
+
+        let make_error = || Error::new("failed to get role", ErrorType::Role);
+
+        let role = match reference {
+            Reference::Id(id) => {
+                Role::find_by_id(id).one(txn).await.or_raise(make_error)?
+            }
+            _ => None,
+        };
```
emmiegit, line 307: Same question re `Reference`.

deepwell/src/services/role/structs.rs
```diff
@@ -0,0 +1,53 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+use crate::types::Maybe;
+use sea_orm::prelude::TimeDateTimeWithTimeZone;
```
emmiegit: In regular code, we have been using `use time::OffsetDateTime;` (it's a lot shorter of a type name).

deepwell/migrations/20220906103252_deepwell.sql
loop-index: Fair concern - putting page category aside, I think one way to solve the localization problem is using the [resource:action] pair as the key to look up in the locale file. Then we don't have to bake the description inside the entry. That would still not work with the custom site permissions though.

> I suppose that actions would be implemented in code (e.g. edit) vs the whole permission?

I think for most actions, just calling a lookup to see if user has permission for a resource would suffice. Checks needing additional logic can have a wrapper over the lookup function to allow for more granular checking.

deepwell/migrations/20220906103252_deepwell.sql
loop-index: Hmm, by 'undeletable' for system role I meant the role cannot be removed from the _site_, not from the _user_.  Revoking an user's membership would be removing the Member UserRole relation for that user and that site, not removing the Member role from that site entirely.

With that in mind, I was thinking a few basic roles like Admin, Member, Guest(virtual) and Author(virtual). Wikidot roles like Moderator I think can be added in for backwards compatibility, but not as a system role so we can migrate away from that role if we'd like.

(Also yes, "virtual" role is a better name than implicit)

emmiegit: > Wikidot roles like Moderator I think can be added in for backwards compatibility, but not as a system role so we can migrate away from that role if we'd like.

Agreed.

I think this logic actually also applies to "author", since that is a socially-mediated definition rather than a technical one. For instance on SCP-EN, being qualified for an author page involves having a number of articles above a certain score threshold, and some page types don't qualify. This logic shouldn't be in our backend but reflected in per-site configuration.

deepwell/migrations/20220906103252_deepwell.sql
loop-index: Yep, I also thought about several options, hierarchy tree (basically an org chart) included. My only concern would be implementation complexity in doing lookups up/down the tree, because we are essentially doing a new query for each up/down level we traverse.

Also would there be any possible cases where we might need more than one org root? That might require a dummy root node.

deepwell/migrations/20220906103252_deepwell.sql
emmiegit: To explore this a bit more, what exactly do you have in mind for custom site permissions? Any illustrative examples?

deepwell/migrations/20220906103252_deepwell.sql
emmiegit: I agree that lookups up and down the tree are non-ideal, but considering tree depths aren't expected to be extreme I am leaning towards saying the cost is acceptable. (We can also set up caching in redis etc for permissions too)

I cannot think of real reasons why more than one root would make sense, I mostly brought it up for completeness. Maybe a separation-of-powers thing in a site? But that could still have a dummy root note in effective implementation.

deepwell/seeder/roles.json
loop-index: In practice, what are usecases where we need to have an unique set of permissions for `user` and `everyone`? Just curious.

The way I think about it, guest encompasses all visitors without membership, whether registered or not, so it can be its own role.

deepwell/seeder/users.json
loop-index: For testing out guest permissions. I guess I can remove it and keep on local if the practice is not to add test users to the data.

deepwell/src/services/permission/service.rs
loop-index: I was thinking that the string variant could be the resource:action key. I should consolidate the permission lookup to accept either id or key.

deepwell/seeder/roles.json
emmiegit: I'm good with that. `guest` as a virtual role for all users without membership, whether registered or not.

I also finally remembered the Wikidot term for non-registered/logged out users: `anonymous`.

So then the scheme would be:
* `registered` or `user` (?) - All users with accounts.
* `anonymous` - All users who are logged out or lack an account.
* `everyone` - All users. ([Top type](https://en.wikipedia.org/wiki/Any_type))
* `guest` - All `anonymous` users, and all `registered` users who are NOT a site member.

----

Anyways, while I do not have particular _permission_ schemes in mind to distinguish between non-site members and anonymous/unregistered users, there are certainly cases where being able to distinguish between those at a platform level would be helpful. For instance, on a system join page, being able to say:
* Anonymous users: First make an account, then apply for membership.
* Non-member users: Apply for membership.
* Member users: You're already a member!
Seems useful.

I'll also note that Wikidot's permissions _do_ enable distinguishing between anonymous users and registered non-site members, even if, in practice, the two are treated as the same.

deepwell/src/services/role/mod.rs
loop-index: They are in seeder data; I guess I should have made this an enum instead, the idea was to avoid using strings in lookup params..

deepwell/src/services/role/service.rs
loop-index: It should remove the role from all users with that role, yes. I think a foreign key delete constraint should work.

deepwell/migrations/20220906103252_deepwell.sql
loop-index: Only use case that I can think of right now is per PageCategory permissions - since each site may have a different list of page categories, the permissions will also have to be scoped per site like how Wikidot currently does it.

deepwell/migrations/20220906103252_deepwell.sql
loop-index: My intention for "Author" role was more inline with the Wikidot "page creator" role actually, and it is only applied when an user visits a page they created/credited with. This is used for page author-only locks for example, where the system needs to determine if the current user is an author of this page.

> being qualified for an author page involves having a number of articles above a certain score threshold,

For the SCP wiki notion of "author" then yes, that should be a site-custom role.

emmiegit: So less of a "role" and that, for the scope of a *particular page*, if someone is an author or maintainer of that page, different permissions could apply?

deepwell/src/services/role/service.rs
emmiegit: I don't think the delete constraint would cascade to a `deleted_at` column, right? (Since we are preferring soft deletion here to avoid loss of historical data)

deepwell/migrations/20220906103252_deepwell.sql
loop-index: Yep! Wikidot allows for customizing permissions for page creators, so this is what I'm trying to replicate under the same abstraction

deepwell/src/services/role/service.rs
loop-index: Ah, you're right. Then some additional logic to remove this role will be needed.

deepwell/migrations/20220906103252_deepwell.sql
emmiegit: Ah I see! I'll note that instead of "page creator" being special, I think we should have page attribution metadata be first-class. So if Alice and Bob are co-authors on a page, they're both equally considered authors in terms of page permissions, even if it was technically Bob who made the page etc. (Ignoring cases like Alice being banned or something, which would supersede that)

deepwell/src/services/permission/service.rs
emmiegit: Hmm. I think a code enum with fixed constant strings for all `resource:action`s would probably be better, given our other discussions.

deepwell/migrations/20220906103252_deepwell.sql
```diff
@@ -952,18 +952,18 @@ CREATE TABLE permission (
...
-    -- Implicit roles are invisible to the user, granted by the system under specific conditions
+    -- Virtual roles are invisible to the user, granted by the system under specific conditions
     -- i.e. Guest role granted when a non-member visits the site,
     -- or Author role granted when a user interacts with a page they authored.
-    -- Implicit roles are visible to admins where they can select the role's permissions.
-    -- Implicit roles cannot be manually assigned..
-    implicit BOOLEAN NOT NULL DEFAULT false,
+    -- Virtual roles are visible to admins where they can select the role's permissions.
+    -- Virtual roles cannot be manually assigned.
+    is_virtual BOOLEAN NOT NULL DEFAULT false,
```
emmiegit, line 978: Design question:

Would it be better for virtual roles to be in the database (and automatically added by jobs or code logic)? This would mean we can just query roles and get all of them.

Or should virtual roles be added automatically in the future `RoleService` when querying for a user? This way they're always up-to-date but it requires that callers go through this code path to ensure they're getting both all the real roles and the virtual ones.

I'm leaning more towards the latter but I'm curious as to your thoughts.

loop-index: Are you talking about a `Role` (a set of permissions) or a `UserRole` relation ([user] has [role] relationship)? 
- Roles should be saved to the database, because if we want to control the permissions a virtual role can have, we need a persistent reference to the role.
- UserRole relations are saved for `[user] has [real role]` but not for `[user] has [virtual role]`. Virtual roles are evaluated on query, based on a passed-in context (current site? current page? has user id?)

In my latest commit I've actually implemented that logic in `RoleService`. When querying for a user's roles, real roles are fetched first from the DB, then virtual roles are appended based on the input context.

loop-index:
Made some updates to the system:
- A permission cache to quickly look up permissions; currently restricted to read perms for caution. Falls back to DB lookup
- Per category permission scoping. Category is abstracted; just need a resource_type variant and a numerical ID. Added a resolver trait to implement different category resolvers for different resource types.
- Removed the Permissions table; previously it was used as a lookup table of sorts, but with enum variants in code it's kinda obsolete. Perms is now flattened into the RolePermission mapping; this helps with category scoping since RolePermissions are scoped per Role, and thus per Site.

Sorry for the messy commit history, git hard

deepwell/migrations/20220906103252_deepwell.sql
```diff
@@ -956,6 +956,70 @@ CREATE INDEX forum_post_parent_idx ON forum_post (parent_post_id);
...
+    role_id BIGSERIAL PRIMARY KEY,
+
+    -- Denotes a unique role in a site. A user may be an admin in one site but a regular site member in another.
+    site_id BIGINT NOT NULL REFERENCES site(site_id),
+    name TEXT NOT NULL,
+    description TEXT NOT NULL,
+    from_wikidot BOOLEAN NOT NULL DEFAULT false,
+
+    -- Virtual roles are invisible to the user, granted by the system under specific conditions
+    -- i.e. Guest role granted when a non-member visits the site,
```
emmiegit:
```suggestion
    -- e.g. Guest role granted when a non-member visits the site,
```

* `i.e.` - id est, substitute "that is" or "more specifically" and see if the sentence works
* `e.g.` - exempli gratia, substitute "for example" when mentually checking

deepwell/migrations/20220906103252_deepwell.sql
```diff
@@ -956,6 +956,70 @@ CREATE INDEX forum_post_parent_idx ON forum_post (parent_post_id);
...
+    -- or Author role granted when a user interacts with a page they authored.
+    -- Virtual roles are visible to admins where they can select the role's permissions.
+    -- Virtual roles cannot be manually assigned.
+    is_virtual BOOLEAN NOT NULL DEFAULT false,
+
+    -- System roles cannot be deleted (i.e. Admin)
+    is_system BOOLEAN NOT NULL DEFAULT false,
```
emmiegit: With our follow-up discussions, I'm clear on the `is_virtual` flag, but I'm not clear on `is_system`. It seems that the only non-virtual system roles `root` and `page-author`?

Basically, how should I conceive of what "system" is?

deepwell/migrations/20220906103252_deepwell.sql
```diff
@@ -956,6 +956,70 @@ CREATE INDEX forum_post_parent_idx ON forum_post (parent_post_id);
...
+    from_wikidot BOOLEAN NOT NULL DEFAULT false,
+
+    -- Virtual roles are invisible to the user, granted by the system under specific conditions
+    -- i.e. Guest role granted when a non-member visits the site,
+    -- or Author role granted when a user interacts with a page they authored.
+    -- Virtual roles are visible to admins where they can select the role's permissions.
+    -- Virtual roles cannot be manually assigned.
+    is_virtual BOOLEAN NOT NULL DEFAULT false,
+
+    -- System roles cannot be deleted (i.e. Admin)
```
emmiegit:
```suggestion
    -- System roles cannot be deleted (e.g. Admin)
```

deepwell/migrations/20220906103252_deepwell.sql
```diff
@@ -956,6 +956,70 @@ CREATE INDEX forum_post_parent_idx ON forum_post (parent_post_id);
...
+    created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
+    updated_at TIMESTAMP WITH TIME ZONE,
+    deleted_at TIMESTAMP WITH TIME ZONE,
+
+    UNIQUE (site_id, name)
+);
+
+-- Role permissions (1-to-many)
+CREATE TABLE role_permission (
+    id BIGSERIAL PRIMARY KEY,
```
emmiegit: We don't really have any bare `id` columns, maybe `permission_id` here? Or `role_permission_id` though that's long.

deepwell/migrations/20220906103252_deepwell.sql
```diff
@@ -956,6 +956,70 @@ CREATE INDEX forum_post_parent_idx ON forum_post (parent_post_id);
...
+    role_id BIGINT NOT NULL REFERENCES role(role_id),
+    -- Denormalized to avoid a join when filtering permissions by site.
+    site_id BIGINT NOT NULL REFERENCES site(site_id),
+    resource_type TEXT NOT NULL,
+    -- Polymorphic reference to a resource category. For example, if the resource_type is "forum_category", then this references forum_category_id.
+    -- NULL means the permission applies to all resources of the given type
+    resource_category_id BIGINT,
+    action TEXT NOT NULL,
```
emmiegit, line 1002: Should we add a database enum type for either of `resource_type` or `action`?

deepwell/migrations/20220906103252_deepwell.sql
```diff
@@ -956,6 +956,70 @@ CREATE INDEX forum_post_parent_idx ON forum_post (parent_post_id);
...
+    UNIQUE (site_id, role_id, resource_type, resource_category_id, action)
+);
+
+-- User role assignments (many-to-many)
+CREATE TABLE user_role (
+    user_id BIGINT NOT NULL REFERENCES "user"(user_id),
+    role_id BIGINT NOT NULL REFERENCES role(role_id),
+    -- Denormalized FK to avoid a join.
```
emmiegit: Just for additional clarity on why this is denormalization:
```suggestion
    -- Denormalized FK to avoid a join. (Remember, roles are tied to a specific site)
```

deepwell/seeder/permissions.json
```diff
@@ -0,0 +1,42 @@
...
+    {
+        "description": "Can create new pages",
+        "resource-type": "page",
+        "action": "create"
+    },
+    {
+        "description": "Can delete pages",
+        "resource-type": "page",
+        "action": "delete"
+    },
```
emmiegit, line 21: Should `page:restore` (i.e. undeletion) be separate from `page:delete`, or be implied in this permission?

deepwell/seeder/permissions.json
```diff
@@ -0,0 +1,42 @@
...
+    {
+        "description": "Can view role assignments",
+        "resource-type": "role",
+        "action": "view"
+    },
+    {
+        "description": "Can assign roles to users",
+        "resource-type": "role",
+        "action": "assign"
+    },
```
emmiegit: So `role:assign` only can add a role to a user? So if I mess up and add a role X to Bob, but I have only `role:assign` and not `role:remove` I cannot fix this?

Mostly asking because my intuition is that `role:assign` means add _and_ remove - if it only means add then we should rename it to `role:add`?

deepwell/src/database/seeder/mod.rs
```diff
@@ -485,6 +491,109 @@ pub async fn seed(state: &ServerState) -> Result<()> {
...
+                    level: role_template.level,
+                },
+                SEED_IP_ADDRESS,
+            )
+            .await
+            .or_raise(make_error)?;
+
+            // Assign permissions to role
+            for perm_spec in &role_template.permissions {
+                let parts = perm_spec.split(':').collect::<Vec<_>>();
```
emmiegit: Not a big deal for the seeder (since it's off the hot path), but for parsing permission strings generally, we can collect into an [`ArrayVec`](https://docs.rs/arrayvec/latest/arrayvec/struct.ArrayVec.html) of 3 to avoid a heap allocation each time.

deepwell/src/endpoints/page.rs
```diff
@@ -202,11 +205,52 @@ pub async fn page_edit(
...
+    info!(
+        "Checking edit permission for page {:?} in site ID {}",
+        input.page, input.site_id,
+    );
+
+    let can_edit = PageService::check_user_permission(
+        ctx,
+        input.site_id,
+        input.user_id,
+        input.page,
```
emmiegit: General note, one reason why we use a lot of structs for requests is to prevent confusion with argument order. Since different IDs are the same type, but may represent different types of object (e.g. site vs user), mixing up the order can cause silent bugs where the code compiles but queries never produce the right results.

deepwell/src/services/permission/cache.rs
```diff
@@ -0,0 +1,188 @@
...
+
+#[allow(dead_code)]
+impl PermissionCache {
+    /// Build Redis cache key for a permission hash.
+    fn key(site_id: i64, resource_type: &str, action: &str) -> String {
+        format!(
+            "perm:{}:{}:{}",
+            site_id,
+            resource_type.to_lowercase(),
+            action.to_lowercase()
```
emmiegit: Would these values ever not be lowercase? If we're getting them from an enum

If we're concerned about them not being lowercase despite this, we could add an assertion?

deepwell/src/services/permission/cache.rs
```diff
@@ -0,0 +1,188 @@
...
+        )
+    }
+
+    /// Build the field key within a permission hash for a specific category.
+    fn category_field(resource_category_id: Option<i64>) -> String {
+        resource_category_id
+            .map(|id| id.to_string())
+            .unwrap_or_else(|| DEFAULT_CATEGORY_KEY.to_string())
+    }
```
emmiegit: Since you just need to borrow a `&str` from this, I think this is a good use case for `Cow`:

```suggestion
    fn category_field(resource_category_id: Option<i64>) -> Cow<'static, str> {
        resource_category_id
            .map(|id| Cow::Owned(id.to_string()))
            .unwrap_or_else(|| Cow::Borrowed(DEFAULT_CATEGORY_KEY))
    }
```

deepwell/src/services/permission/cache.rs
```diff
@@ -0,0 +1,188 @@
...
+        let mut redis = ctx.redis();
+
+        let resource_str = resource_type.to_string();
+        let action_str = action.to_string();
+
+        let key = Self::key(site_id, &resource_str, &action_str);
+        let field = Self::category_field(resource_category_id);
+
+        let role_ids_str: Option<String> =
+            redis.hget(&key, &field).await.map_err(|e| {
```
emmiegit: Is there a reason why `.or_raise()` can't be used here? (And if it can't, shouldn't we add the `e` error information to the error message, since that is now the lowest-level error?)

deepwell/src/services/permission/cache.rs
```diff
@@ -0,0 +1,188 @@
...
+                    "Failed to read permission cache key '{}' field '{}': {}",
+                    key, field, e
+                );
+                Error::new("Permission cache read error", ErrorType::Permission)
+            })?;
+
+        // Fall back to default category if specific category is not found.
+        let role_ids_str = if role_ids_str.is_none()
+            && resource_category_id.is_some()
+            && resource_category_id != Some(-1)
```
emmiegit: `-1` being `_default`?

deepwell/src/services/permission/cache.rs
```diff
@@ -0,0 +1,188 @@
...
+                let parsed = s
+                    .split(',')
+                    .filter_map(|part| part.trim().parse::<i64>().ok())
+                    .collect();
+                Ok(Some(parsed))
+            }
+        }
+    }
+
+    pub async fn build_tree(ctx: &ServiceContext<'_>, site_id: i64) -> Result<()> {
```
emmiegit: Nitpick: Maybe rename `build_tree_cache()` or `build_redis_tree()` to emphasize that it's creating a tree structure for storage in the cache (as opposed to returning an object as the return value)?

deepwell/src/services/permission/service.rs
```diff
@@ -0,0 +1,350 @@
...
+            Some(reference) => {
+                resolve_category_reference(ctx, site_id, resource_type, reference).await?
+            }
+            None => None,
+        };
+
+        let make_error = || {
+            Error::new(
+                format!(
+                    "failed to remove permission {}:{}:{} from role ID {}",
```
emmiegit:
```suggestion
                    "failed to remove permission '{}:{}:{}' from role ID {}",
```

deepwell/src/services/permission/service.rs
```diff
@@ -0,0 +1,350 @@
...
+        let category_condition = match resource_category_id {
+            Some(id) => role_permission::Column::ResourceCategoryId.eq(id),
+            None => role_permission::Column::ResourceCategoryId.is_null(),
+        };
+
+        Ok(RolePermission::find()
+            .filter(role_permission::Column::SiteId.eq(site_id))
+            .filter(role_permission::Column::ResourceType.eq(&resource_str))
+            .filter(category_condition)
+            .filter(role_permission::Column::Action.eq(&action_str))
```
emmiegit, line 181: I did not know you could do this. I've always done
```rust
.filter(
    Condition::all()
        .add(...)
        .add(...)
)
```

deepwell/src/services/permission/service.rs
```diff
@@ -0,0 +1,350 @@
...
+            // If no scoped permission, check for _default permission
+            Self::query_roles_with_permission(ctx, site_id, resource_type, None, action)
+                .await
+                .or_raise(make_error)?
+        } else {
+            specific_roles
+        };
+
+        info!(
+            "Queried database for permission {}:{}:{}, found {} roles",
```
emmiegit:
```suggestion
            "Queried database for permission '{}:{}:{}', found {} roles",
```

deepwell/src/services/permission/service.rs
```diff
@@ -0,0 +1,350 @@
...
+
+                if let Some(cached) = cached_roles {
+                    info!("Permission cache hit, {} roles found", cached.len());
+                    cached
+                } else {
+                    info!("Permission cache miss, querying database directly");
+
+                    // If cache miss, rebuild tree async
+                    #[allow(unused_must_use)]
+                    PermissionCache::build_tree(ctx, site_id);
```
emmiegit: The correct way to note that you're explicitly not using a result is through:

```rust
                    let _ = PermissionCache::build_tree(ctx, site_id);
```

However, in this case, the `must_use` comes from the fact that you have a future `Result` you're not handling. So the actual change here is:


```suggestion
                    PermissionCache::build_tree(ctx, site_id).await.or_raise(make_error)?;
```

deepwell/src/services/permission/structs.rs
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+use crate::types::{Action, Reference, Resource};
+
+#[derive(Debug, Clone)]
+pub struct PermissionInput {
+    pub resource_type: Resource,
+    pub resource_category: Option<Reference<'static>>,
+    pub action: Action,
+}
```
emmiegit: Any reason it can't be
```suggestion
pub struct PermissionInput<'a> {
    pub resource_type: Resource,
    pub resource_category: Option<Reference<'a>>,
    pub action: Action,
}
```

deepwell/src/services/role/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+use strum_macros::{Display, EnumString};
```
emmiegit:
```suggestion
 */

use strum_macros::{Display, EnumString};
```

deepwell/src/services/role/service.rs
```diff
@@ -0,0 +1,557 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+use super::prelude::*;
```
emmiegit, line 21:
```suggestion
 */

use super::prelude::*;
```

deepwell/src/services/role/service.rs
```diff
@@ -0,0 +1,557 @@
...
+
+        let model = role::ActiveModel {
+            site_id: Set(site_id),
+            name: Set(name.clone()),
+            description: Set(description.clone().unwrap_or_default()),
+            is_virtual: Set(is_virtual),
+            is_system: Set(is_system),
+            level: Set(level),
+            created_at: Set(now),
+            updated_at: Set(Some(now)),
```
emmiegit: If it's created but not yet changed after that, `updated_at` should be `NULL`
```suggestion
```

deepwell/src/services/role/service.rs
```diff
@@ -0,0 +1,557 @@
...
+                    .map(|permission| role_permission::ActiveModel {
+                        role_id: Set(role.role_id),
+                        site_id: Set(role.site_id),
+                        resource_type: Set(permission.resource.to_string()),
+                        resource_category_id: Set(permission
+                            .resource_category
+                            .as_ref()
+                            .and_then(|r| match r {
+                                Reference::Id(id) => Some(*id),
+                                Reference::Slug(_) => None,
```
emmiegit: I don't understand this logic, why is `Reference::Slug(_)` mapped to `None`?

deepwell/src/services/role/service.rs
```diff
@@ -0,0 +1,557 @@
...
+                ctx,
+                GetSiteMember {
+                    site_id: input.site_id,
+                    user_id: input.user_id.unwrap(),
+                },
+            )
+            .await
+            .or_raise(make_error)?;
+
+            // TODO: Add invitation acceptance logic here
```
emmiegit: What do you mean by "invitation acceptance logic"?

deepwell/src/services/role/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+#[allow(dead_code)]
+pub enum SystemRole {
+    Root,
+    Member,
+    Guest,
+    Registered,
+    Anonymous,
+    Everyone,
+    PageAuthor,
+}
```
emmiegit, line 40: You can add a `strum_macros::IntoStaticStr` derive and a `name() -> &'static str` method. Since `&'static str`s are `Copy` and cheap to pass around, that seems useful since there are cases where we are passing around the string version of these enums.

deepwell/src/services/role/service.rs
```diff
@@ -0,0 +1,557 @@
...
+
+        info!(
+            "Applying these virtual roles for user ID {:?} in site ID {}: {:?}",
+            input.user_id, input.site_id, applied_virtual_roles
+        );
+
+        // Build the final list of applicable roles
+        let applicable_virtual_roles = applied_virtual_roles
+            .into_iter()
+            .filter_map(|name| virtual_role_name_map.get(&name).cloned())
```
emmiegit: Then down here, we could just push `SystemRole::Everyone` or whatever to the vector without needing to clone the strings, and do `.filter_map(|role| virtual_role_name_map.get(role.name()).cloned())`

deepwell/src/types/permissions.rs
```diff
@@ -0,0 +1,83 @@
...
+                // Try to parse as ID first, fall back to treating as slug
+                let reference = resource_category
+                    .parse::<i64>()
+                    .map(Reference::Id)
+                    .unwrap_or_else(|_| {
+                        Reference::Slug(Cow::Owned(resource_category.to_string()))
+                    });
+                (resource, Some(reference), action)
+            }
+            _ => return Err("invalid permission format"),
```
emmiegit: Looks like this is the only error type? It's usually better to define a light or unit struct as an error rather than a bare `&'static str`. The typing is more clear (since you can't perform any string operations on it), and it's easier to find the origin of this error type.

Something like
```rust
pub struct PermissionParseError;
```

deepwell/migrations/20220906103252_deepwell.sql
loop-index: Original intent was that system roles cannot be deleted (admin, mods) but since we narrowed it down to only make root user an undeletable role, I can remove this flag and use the SiteUser relation or something.

deepwell/migrations/20220906103252_deepwell.sql
loop-index: I did look into this before and the general consensus seems to be that Postgres enums are a pain to work with unless we have a solid list of all the variants that we want. But that is also mainly due to schema migration setups; since we are rebuilding the db anyway, that might not be that much of a hassle. lmk what you think.

deepwell/seeder/permissions.json
loop-index: Right now I'm just putting in an arbitrary list of permissions. I figure we will add more as we integrate more operations with permissions.

deepwell/seeder/permissions.json
loop-index: Fair enough, should have a single role for both add and remove. I'll keep it as `assign` and remove the `role:remove`.

deepwell/src/services/permission/cache.rs
loop-index: Right, this is unnecessary.

deepwell/src/services/permission/service.rs
loop-index: Right, I've also just found out recently that Rust asyncs are not run until awaited.

deepwell/migrations/20220906103252_deepwell.sql
loop-index: on further examination, seems like SiteUser is the representative account of a site and not the root user/site creator. I'll keep this for now and think of a way to represent this role in a way that is permanent and transferable (like Master Admin on wikidot) such that there's always at least one user per site with full admin permissions.

deepwell/src/services/role/service.rs
loop-index: Forgot to fill this in, this was placeholder before I implemented the category resolver.

deepwell/src/services/role/service.rs
loop-index: I think I was sleep-deprived and misunderstood the SiteApplication relation somehow. It made sense in my mind back then, but I can't remember the reasoning now. removed.

deepwell/src/services/permission/cache.rs
loop-index: I planned to use it as a convention but then forgot to drop it when no longer needed. With hmget() this is now removed.

deepwell/migrations/20220906103252_deepwell.sql
emmiegit: Correct, the "site user" is a special user type not controlled by any human which represents the site as a whole. So like, it's the user to message you on ban, etc.

As for `root`, I think we should avoid making it an equivalent of master admin. That one user being truly unevictable is a lot of trust and has caused issues on some branches if that trust turns out to be misplaced. I would prefer if it is possible for sites to construct their roles in such a way that all members of an admin-equivalent role have equal access, and then some kind of majority/multi approval system enables actions affecting the group as a whole.

Or put in more programmer terms, if `root` isn't a person but a role you can assume by doing "sudo" (whatever that means for a site's staff).

Implementing mechanisms for approved actions like this is out of scope for this PR, but I just wanted to give a sense of what I'm thinking to avoid social issues we have due to Wikidot design decisions.

deepwell/migrations/20220906103252_deepwell.sql
emmiegit: I'll start by noting that the recreating the database thing is a temporary measure to avoid having a long series of migrations as we stabilized the early schema. However after we reach the stage where we launch a prod instance, we will be using migrations iteratively as expected from then on.

So you raise a good point about enums, and something we should think about for the code. Overall, I have been using Postgres enums in our other tables, but you are correct that adding or removing types is more cumbersome. It *is* possible through `ALTER TYPE` etc., and native types do have the advantage that it enforces against typos or other incorrect value issues. We also have some `CHECK` constraints using these enums as well.

I've considered both angles, and I'm not sure. I would be fine with in-code enums if we set up a consistent trait system for it and checking to ensure that invalid values can't find their way into the table. Overall I think, whatever we decide, we should be consistent in the code.

What do your thoughts?

Syxless: another option worth considering is using reference tables instead of either enums or strings. so like a `permission_resource` table and a `permission_action` table, then foreign keys from role_permission into those

adding new resources/actions is just an INSERT rather than an ALTER TYPE, so no migration pain, but you still get db-level validation via foreign keys. and it fits the existing pattern in deepwell where everything uses integer ids and relations

postgres enums are fine when the variant list is truly static, but permissions are going to grow as we add features, so i think either reference tables or rust-side validation with text columns makes more sense here

loop-index: I actually started out this PR with that idea lol (DB side ref table). It definitely works but causes additional joinage and complexity when querying. I deprecated it in favor of flattening it onto role_permission and code enums.

I think code-side enums with some sort of validation makes the most sense. Making sure that the read/write entry points to the table only uses valid enum variants should be good. Maybe we can write some sort of trait that validates and generates query conditions instead of passing values directly to queries?

Syxless: this sounds good.

emmiegit: This makes sense to me as well.

deepwell/src/services/permission/service.rs
emmiegit: FYI in case you need to kick something off and have it run in the background / not block on its result, you can use [`tokio::spawn`](https://docs.rs/tokio/latest/tokio/task/fn.spawn.html). (Doesn't apply in this case but just in case it's helpful)

deepwell/src/services/permission/service.rs
```diff
@@ -0,0 +1,354 @@
...
+                if let Some(cached) = cached_roles {
+                    info!("Permission cache hit, {} roles found", cached.len());
+                    cached
+                } else {
+                    info!("Permission cache miss, querying database directly");
+
+                    // If cache miss, rebuild tree async
+                    let _ = PermissionCache::build_permission_cache(ctx, site_id)
+                        .await
+                        .or_raise(make_error);
```
emmiegit: This is still not correct. This produces a `Result` that you are not `?`ing on, so if this errors you're just eating the error. I don't see an indication that this is what you want.

deepwell/src/services/role/service.rs
```diff
@@ -0,0 +1,562 @@
...
+            .or_raise(make_error)?;
+            attributions
+                .iter()
+                .any(|attr| attr.user_id == input.user_id.unwrap())
+        } else {
+            false
+        };
+
+        // Collect virtual roles to apply based on flags
+        let mut applied_virtual_roles = Vec::new();
```
emmiegit: Since this will be at minimum 2 items and at most 4: (optimization hint for allocation)
```suggestion
        let mut applied_virtual_roles = Vec::with_capacity(4);
```

deepwell/src/services/permission/service.rs
loop-index: missed it, will fix

emmiegit (approved):

PR #2760, opened 2026-03-14
Add missing resources to Komodo dev TOML files

emmiegit:
There are a few objects which are not presently stored in our TOML definition files, so we should add them here and then set the resource sync to delete any missing items (declaring it total).


emmiegit:
Going to merge after #2759 just to be safe

Zokhoi (approved):

PR #2765, opened 2026-03-15
Refactor forms with Superforms

hoah2333:
The current `<form>` submission hacks are somewhat brittle and lack proper type safety. To address this, I've refactored all submission logic to [Superforms](https://superforms.rocks/), including a revamp of non-form submission flows.

To achieve full type-safety, I replaced global `$app/state` usage with `$props`. This allows SvelteKit to automatically pass types from the server, eliminating the need to manually define everything in `app.d.ts`.

By the way, I've manually mapped almost all the endpoints we're using for now. We should probably look into [ts-rs](https://github.com/Aleph-Alpha/ts-rs) or a similar tool to automate the type generation in the future.


framerail/src/routes/[slug]/[...extra]/FilePane.svelte
```diff
@@ -303,7 +374,7 @@
             {new Date(file.file_created_at).toLocaleString()}
           </div>
           <div class="file-attribute updated-at">
-            {file.file_updated_at ? new Date(file.file_updated_at).toLocaleString() : ""}
+            {new Date(file.file_updated_at ?? file.file_created_at).toLocaleString()}
```
Comment (Zokhoi) [inline framerail/src/routes/[slug]/[...extra]/FilePane.svelte]: I'm not sure on this, would including file creation time for last edit time be easier to understand? This would likely be an arbitrary choice.

framerail/src/routes/[x+2d]/register/+page.svelte
```diff
@@ -1,154 +1,133 @@
...
+        placeholder={data.internationalization?.["email.placeholder"]}
         required
         type="text"
+        bind:value={$form.email}
       />
-      <p class="description">{page.data.internationalization?.["email.info"]}</p>
+      <p class="description">{data.internationalization?.["email.info"]}</p>
+      {#if $errors.email}
+        <!-- Although `error-api.INVALID_EMAIL` is almost the same as this one, it should
+            get another name in `error-form` to avoid confusion -->
```
Comment (Zokhoi) [inline framerail/src/routes/[x+2d]/register/+page.svelte:74]: You can make new locale strings in fluent if you think it's necessary.

framerail/src/routes/[x+2d]/register/+page.svelte
```diff
@@ -1,154 +1,133 @@
...
+    <label for="locale">UNTRANSLATED:Locale</label>
+    <div class="input-container">
+      <!-- TODO: Implement a multi select component -->
+      <!-- I know it's ugly, but we can implement a better looking component later on -->
+      <select id="locale" name="locale" multiple required bind:value={$form.locale}>
+        <option value="en">UNTRANSLATED:en</option>
+        <option value="ko">UNTRANSLATED:ko</option>
+        <option value="pl">UNTRANSLATED:pl</option>
+        <option value="vi">UNTRANSLATED:vi</option>
+        <option value="zh_Hans">UNTRANSLATED:zh_Hans</option>
```
Comment (Zokhoi) [inline framerail/src/routes/[x+2d]/register/+page.svelte]:
```suggestion
        <option value="zh-Hans">UNTRANSLATED:zh_Hans</option>
```
We need to use hyphenated `zh-Hans` here for value. The user locale list will be used in javascript native localization functions and they only accept hyphenated versions for locales.

framerail/src/lib/server/load/user.ts
```diff
@@ -61,30 +64,41 @@ export async function loadUser(username?: string, request, cookies) {
...
+      response.data.user_session?.user?.user_id !== response.data.user.user_id
+
+    viewData.user = deleteSensitiveData(response.data.user, isViewingAnotherUser)
 
     // Get user avatar image
-    if (viewData.user.avatar_s3_hash !== null) {
-      const avatar = await getFileByHash(new Uint8Array(viewData.user.avatar_s3_hash))
+    if (
+      response.data.user.avatar_s3_hash !== null &&
+      response.data.user.avatar_s3_hash !== undefined
```
Zokhoi: The `avatar_s3_hash` will only be `null` and not `undefined` as we have discussed

framerail/src/routes/[slug]/[...extra]/FilePane.svelte
```diff
@@ -1,213 +1,290 @@
...
+    return (input: Parameters<SubmitFunction>[0]) => {
+      return new Promise<XMLHttpRequest>((resolve) => {
+        const xhr = new XMLHttpRequest()
+        xhr.onload = () => {
+          if (xhr.readyState === xhr.DONE) {
+            resolve(xhr)
+          }
+        }
+        xhr.open("POST", input.action, true)
+        xhr.send(formData)
```
Comment (Zokhoi) [inline framerail/src/routes/[slug]/[...extra]/FilePane.svelte]: Curious on the use of xhr here, is it not possible to use fetch and then call `resolve`, for example `fetch(...).then(res=>resolve(res))`?

framerail/src/routes/[slug]/[...extra]/FilePane.svelte
Comment (emmiegit) [reply framerail/src/routes/[slug]/[...extra]/FilePane.svelte]: Given that there is a "created at" display field above, if there's no "updated at" I feel like it's better to hide it.

That is, only show an "Updated at: [...]" when it's non-null.

Comment (Zokhoi) [reply framerail/src/routes/[slug]/[...extra]/FilePane.svelte]: This is for a table of files on a page, and some files may have been modified after creation and some may not, so the column will need to be there.

Comment (emmiegit) [reply framerail/src/routes/[slug]/[...extra]/FilePane.svelte]: Oh I see, I misunderstood.

In this case my suggestion would be to have "updated at" be blank or a dash or something to indicate that it's blank.

Comment (hoah2333) [reply framerail/src/routes/[slug]/[...extra]/FilePane.svelte]: I feel like using the same date for the "created at" filed looks a bit cleaner, but a dash is a solid aternative.

framerail/src/routes/[slug]/[...extra]/FilePane.svelte
Comment (hoah2333) [reply framerail/src/routes/[slug]/[...extra]/FilePane.svelte]: That was a mistake on my part, I was misled by [the file upload document](https://superforms.rocks/concepts/events#customrequest), wrongly thought XHR was the only way to send files along with extra form data.

framerail/src/lib/server/deepwell/pageFile.ts
```diff
@@ -159,32 +212,37 @@ export async function pageFileHistory(
...
 }
 
+/* ----- Page File Rollback ----- */
 export async function pageFileRollback(
   siteId: number,
-  pageId: Optional<number>,
+  pageId: number,
   userId: number,
-  fileId: string,
+  fileId: string | number,
```
emmiegit: I assume that file IDs being strings here is a holdover from when they used to use CUIDs (a string type)? Now they're just normal 64-bit integers.

framerail/src/lib/server/load/user.ts
```diff
@@ -61,30 +64,38 @@ export async function loadUser(username?: string, request, cookies) {
       errorStatus = 500
   }
 
-  if (errorStatus === null && username && viewData.user.slug !== username) {
-    redirect(308, `/-/user/${viewData.user.slug}`)
+  if (
```
emmiegit, line 68: Maybe a comment to indicate intent:
```suggestion
  // Handle user aliases
  if (
```

framerail/src/lib/server/load/user.ts
```diff
@@ -114,13 +125,97 @@ export async function loadUser(username?: string, request, cookies) {
...
+  return { ...viewData, view: response.type, internationalization, userEditForm }
+}
+
+function deleteSensitiveData(
+  user: UserModel,
+  isViewingAnotherUser: boolean
+): Partial<UserModel> {
+  let sensitiveKeys = ["password", "multi_factor_secret", "multi_factor_recovery_codes"]
+  if (isViewingAnotherUser) {
+    sensitiveKeys = [...sensitiveKeys, "email", "email_is_alias", "email_verified_at"]
```
emmiegit: For viewing another user, I'd also add:
* `email_validation_info` (new field, contains email)
* `email_validation_at` (new field)
* `name_changes_left`
* `last_name_change_added_at`

I'm unsure on `locales` but leaning to it also being removed. It could reveal something about the user's demographic info (e.g. bilingual, immigrant family speaking a different language at home) or region (e.g. a particular dialect or regional variant of a language). Users can explicitly specify their region etc if they wish via the `location` field.

emmiegit: Honestly maybe it's safer to have the opposite, a whitelist system where we set some keys as safe for export to self and export to others, and thus any new properties are by default excluded.

framerail/src/lib/server/deepwell/pageFile.ts
Zokhoi: The endpoint for file rollback accepts `Reference<'a>` but I'm not clear on what a "file slug" refers to. The `fileId` should nonetheless be `number` only to avoid confusion.

emmiegit: I believe it's filename, but I agree that that's not clear (it's not a "slug") and should be changed.

framerail/src/lib/server/load/user.ts
```diff
@@ -136,17 +137,48 @@ export async function loadUser(request: Request, cookies: Cookies, username?: st
...
+    "name",
+    "slug",
+    "avatar_s3_hash",
+    "user_page"
+  ]
   if (isViewingAnotherUser) {
-    sensitiveKeys = [...sensitiveKeys, "email", "email_is_alias", "email_verified_at"]
+    const safeKeys = [...baseSafeKeys]
+    return Object.fromEntries(
+      Object.entries(user).filter(([key]) => safeKeys.includes(key))
```
Zokhoi: Nit: You can just use `baseSafeKeys.includes()` if there is no modification here

framerail/src/routes/[x+2d]/user/+page.svelte
```diff
@@ -1,175 +1,162 @@
...
-    const res = await fetch(`/-/user`, {
-      method: "POST",
-      body: fdata
-    }).then((res) => res.json())
-    if (res?.message) {
-      errorPopupState.current = {
-        state: true,
-        message: res.message,
-        data: res.data
+  const { form, enhance } = superForm(
```
Comment (Zokhoi) [inline framerail/src/routes/[x+2d]/user/+page.svelte:32]: For deepwell apis that accept `Unset` like user edit, we need to remove the unchanged fields on form submit, otherwise it triggers things like the forbidden name filter on admin user profile update.

framerail/src/lib/server/load/user.ts
hoah2333: The current approach is just in case we need to add more fields later, which I think we almost certainly will.

Zokhoi: I can't really think of anything that will be safe to show to others but not to the user themselves, i.e. the whitelist for viewing other user profiles should be a subset of that for viewing their own.

framerail/src/routes/[x+2d]/user/+page.svelte
Comment (emmiegit) [reply framerail/src/routes/[x+2d]/user/+page.svelte:32]: If desired, I could also modify the unchanged fields logic to ignore instances where it is `Set` but to the same value?

(Also maybe the admin user should have a filter exemption?)

emmiegit:
The PR seems fine to me overall, though I lack the same frontend expertise that Zokhoi has.

Zokhoi:
Ran locally and verified the changes mostly work, now just waiting for the review comments to be resolved

framerail/src/routes/[x+2d]/user/+page.svelte
Comment (Zokhoi) [reply framerail/src/routes/[x+2d]/user/+page.svelte:32]: I'm not sure of the work it would take for the deepwell changes, if the workload is too large we can delegate that till later. For now the problem can be avoided by frontend removing unchanged fields.

And yeah the system and admin users should be exempted from the filters.

framerail/src/lib/server/load/user.ts
```diff
@@ -114,13 +126,128 @@ export async function loadUser(username?: string, request, cookies) {
...
+      data: error.data
+    })
+  }
+}
+
+export const userEditSchema = object({
+  name: string(),
+  realName: string(),
+  email: string(),
+  avatar: file(),
```
Zokhoi: The form submit fails when no file is attached: `Invalid type: Expected File but received undefined`.

Edit: All user edit fields should be optional in validation as the deepwell endpoint accept `Maybe::Unset`. `optional(string())` and `optional(file())` works.

framerail/src/routes/[x+2d]/user/+page.svelte
```diff
@@ -7,18 +7,54 @@
...
+  })
+
   const { form, enhance } = superForm(
     untrack(() => data.userEditForm),
     {
-      onResult: async ({ result }) => {
+      dataType: "json",
+      onSubmit: ({ jsonData }) => {
+        const { avatar, ...rest } = $form
+        lastSubmitted = rest
```
Comment (Zokhoi) [inline framerail/src/routes/[x+2d]/user/+page.svelte]: Doesn't this make `lastSubmitted` values the same as current submission before the change comparison?

Comment (hoah2333) [reply framerail/src/routes/[x+2d]/user/+page.svelte]: Yes. We don't need `lastSubmitted` to be the same as `submitForm`, because the latter filters out everything unchanged, and we still need those fields when user clicks edit button again.

Comment (Zokhoi) [reply framerail/src/routes/[x+2d]/user/+page.svelte]: I mean this line makes `lastSubmitted` and `rest` the same before they are checked against each other in `submitForm`, thus `submitForm` will always be empty. It should be moved to after `submitForm` creation.

Zokhoi (approved):

PR #2777, opened 2026-03-16
Extend filter immunity to seed-created users and system user types

emmiegit:
As [discussed in #2765](https://github.com/scpwiki/wikijump/pull/2765#discussion_r2942960409), for user filters, we extend `bypass_filters` to also apply for users created by the seeder, and all system-maintained user types, since such collisions may cause unrelated systems failings or should not apply to begin with.


emmiegit:
Fun PR number!

Zokhoi (approved):

PR #2797, opened 2026-03-23
Add client-side date formatting for ftml [[date]] block

jsthope:
This is related to  [WJ-1311](https://scuttle.atlassian.net/browse/WJ-1311) [WJ-1174](https://scuttle.atlassian.net/browse/WJ-1174) [WJ-1328](https://scuttle.atlassian.net/browse/WJ-1328)  [WJ-1106](https://scuttle.atlassian.net/browse/WJ-1106) 
This need to be used with [this ftml version](https://github.com/scpwiki/ftml/pull/106).

[WJ-1311]: https://scuttle.atlassian.net/browse/WJ-1311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WJ-1174]: https://scuttle.atlassian.net/browse/WJ-1174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WJ-1328]: https://scuttle.atlassian.net/browse/WJ-1328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WJ-1106]: https://scuttle.atlassian.net/browse/WJ-1106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


framerail/src/lib/client/dateformatter.ts
```diff
@@ -0,0 +1,608 @@
...
+type RelativeTimeUnit = "second" | "minute" | "hour" | "day"
+
+let icuPromise: Promise<IcuModule> | null = null
+
+const localeCache = new Map<string, ResolvedLocale>()
+const icuFormatterCache = new Map<string, unknown>()
+const intlDateTimeFormatterCache = new Map<string, Intl.DateTimeFormat>()
+const intlRelativeTimeFormatterCache = new Map<string, Intl.RelativeTimeFormat>()
+
+function normalizeDisplay(value: string) {
```
emmiegit: Could you add return type annotations here and on the other functions? (to the extent reasonable, of course)
```suggestion
function normalizeDisplay(value: string): string {
```

(I'll also note that I'm unfamiliar with the extent to which Typescript can infer return types)

emmiegit (approved):
Seems fine to me. Please get sign-off from @hoah2333 and/or @Zokhoi before merging.

hoah2333:
Should we parse the date into English on the server side first, and then re-render it for the user's locale on the client? As it stands, users might see an unformatted string for a moment before it suddenly switches to the formatted version, which isn't ideal.

jsthope:
I set it as a fallback because that's what Wikidot does but I also think it looks cleaner if it's not set. So I don't know, what you think @emmiegit / @Zokhoi ?

emmiegit:
The temporary unformatted string is not an ideal experience, but wouldn't something similar also happen if it had English as a fallback language?  It would flash English for a bit before getting localized.

Though if we did have an initial render, maybe doing it in the site's language makes sense? This way for a CN site, all the dates start in CN before getting localized by JS. This way CN users on that site don't see any interruption.

Zokhoi:
Yeah we should use site locale when it's available, which I think it is for everywhere that makes custom date formats available. Doesn't hurt to have English after the site locale as a fallback though.

jsthope:
For me, the cleanest solution would simply be to have no initial rendering. Even if it's translated according to the site's localization, an initial rendering should still adapt to the user's utc.

Zokhoi:
Well the problem comes in two parts, first the language and the timezone the date is shown in. These two could be decoupled and have their fallback list be different.

My thought for the language is, the date syntax is part of wikitext and not general page options, so the language should follow the overall wikitext (generally same as the site locale).

However for timezone, as a UX thing we should let user set their preference to show in a specific timezone or follow the site timezone. This could be done later on imo.

Zokhoi:
Then the second part of the problem is, where do we try to format the date? We can do it in the framerail server, or we can do it in the client webpage. If we can do it in the framerail server, then we can avoid flashing different formats to the user.

jsthope:
Yes, that makes sense to me. We can create a pre-formatted version in the site's language and then a client-side reprocessing to adapt to the time zone. For server-side formatting, we can use [the older version of ftml](https://github.com/scpwiki/ftml/pull/105). In this case, should we assume UTC+0 or one of the UTC values ​​from the localization?

Zokhoi:
See `framerail/src/lib/server/load/page.ts`, function `loadPage`, the part that calculates for `wiki-page-last-edit`. Similar logic can be used, and the user and site information should be available at that point.

jsthope:
Indeed but wouldn't it be better to do it in the ftml to avoid formatting per request? (given that the ftml already knows the site's locale)

Zokhoi:
I think the ftml renders are usually cached on deepwell's side, and if the user have the option to customize their preferred timezone shown later on, then we have to modify the renders anyhow. (If they want to see it in site timezone then no modification needed.)

Zokhoi:
So I guess the situation would be, we first cache a render in site language + timezone. Then if user has a specific timezone preference, we reformat the render in post, whether that be in framerail server or in client webpage. I think we're on the same page here.

jsthope:
Yes you're right that's a good idea. So we would need to check if the site timezone is the same as the client's, and then reformat if it differs. For a user who doesn't have the same timezone/no account, how can we ensure the framerail server is aware of this difference before requesting the page? Should we first use local reformatting before the client sends its timezone, and then switch to server-side formatting (framerail) for next requests?

Zokhoi:
I believe there isn't a situation where the server will need to send page content before knowing about the user preference (or lack thereof). The initial page request will include the `accept-language` header and user session cookie if they exist, and from there we can infer the remaining. For users that sent no preference, we can default to site timezone.

jsthope:
Okay I see. So you want to use the site's timezone and only adapt it server-side if the user has preferences. In my opinion, not adapting the date timezone by default could confuse users. But it's just a matter of design. Setting it to the site's timezone by default could also make sense.

jsthope:
The problem is that ``accept-language`` does not provide the timezone, so the timezone cannot be determined before a user preference is established :(

jsthope:
What are your thoughts on this @emmiegit / @hoah2333 ?

Zokhoi:
> In my opinion, not adapting the date timezone by default could confuse users.

We really don't have a good way to solve this when we need to follow the Wikidot syntax, which allows for the wikitext author to choose to hide the timezone in their formatting (and hiding being the default). In my opinion if the author wants to convey a precise time, then it's on them to format the date such that a timezone is included and not confuse their readers. Any timezone is better than none, the rest is just convenience.

hoah2333:
As for the timezone, we'd better align it with the language's default. For example, French would use GMT+1, and English, from what I've seen, usually follows American Eastern time.

jsthope:
Yeah that makes sense. I suggest that for now we go directly through ftml (it already knows the site's language) to format and localize the date (with timezone implicitly set when the date block is created). Then in the future, if we want to implement user preferences for this, we'll look back on it of whether to adapt it server-side in the framerail or client-side.

jsthope:
Thank you all three for your feedback. As [the rest is the responsibility of ftml](https://github.com/scpwiki/ftml/pull/106) this PR can be closed.

PR #2803, opened 2026-03-28
Upgrade redis crate in deepwell

emmiegit:
While `rsmq-async` still requires an older version of the crate, this is not blocking us from upgrading the version in use in the `deepwell` crate proper, and enables use of redis methods like `HMGET`.


emmiegit:
I am foolish and forgot things about this issue.

PR #2805, opened 2026-03-28
[WJ-40] [WJ-1359] add soft-delete and restore for forum posts

Syxless:
- Add `ForumPostService::delete()` to soft-delete individual posts without cascading to children
- Add `ForumPostService::restore()` to undelete posts, with `ForumPostNotDeleted` error check
- Add `ForumPostNotDeleted` error type in the `4500` range
- `list_structured()` always includes deleted posts to preserve tree structure
- Strip revision content from deleted posts in `build_post_nodes()` to avoid leaking data

### DIFF
--- a/deepwell/src/services/forum_post/service.rs
+++ b/deepwell/src/services/forum_post/service.rs
@@ -461,7 +461,6 @@ impl ForumPostService {
         GetStructuredForumPosts {
             forum_thread_id,
             start_post_id,
-            include_deleted: _,
             limit,
             max_depth,
         }: GetStructuredForumPosts,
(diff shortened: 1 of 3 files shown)


deepwell/src/services/forum_post/service.rs
```diff
@@ -346,7 +461,7 @@ impl ForumPostService {
         GetStructuredForumPosts {
             forum_thread_id,
             start_post_id,
-            include_deleted,
+            include_deleted: _,
```
emmiegit: If this is being ignored, would it make sense to remove the field from the struct entirely?

emmiegit (approved):
Seems good, one comment

deepwell/src/services/forum_post/service.rs
Syxless: It does!

emmiegit:
LGTM

PR #2806, opened 2026-03-29
Upgrade redis and rsmq_async dependency versions

emmiegit:
After https://github.com/DavidBM/rsmq-async-rs/issues/30, we can upgrade both `redis` for direct use and `rsmq_async` to take in those new changes.


loop-index (approved):
i love open source

Thanks for taking care of this!! 🙌

PR #2817, opened 2026-04-04
[WJ-1360] implement PageQueryService::find() with FoundPages output

Syxless:
- Rename `PageQueryService::execute()` to `PageQueryService::find()`
- Add `FoundPages`, `FoundPageRow`, and `FoundPageFields` output types
- Callers specify which fields to include via `FoundPageFields`, unneeded fields are `None`
- Populate page-table fields when requested, revision/vote fields left as TODO
- Remove old unused `PageQueryOutput` and `PageResult` structs

### DIFF
--- a/deepwell/src/services/page_query/structs.rs
+++ b/deepwell/src/services/page_query/structs.rs
@@ -291,6 +291,8 @@ pub struct FoundPageFields {
     pub title: bool,
     pub alt_title: bool,
     pub slug: bool,
+    pub page_category_id: bool,
+    pub page_revision_id: bool,
     pub tags: bool,
     pub created_at: bool,
     pub created_by: bool,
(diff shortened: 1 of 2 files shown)


deepwell/src/services/page_query/structs.rs
```diff
@@ -281,17 +278,65 @@ pub struct PageQuery<'a> {
...
+}
+
+/// A single page row in the query results.
+///
+/// Fields are optional because callers specify which fields
+/// they need via `FoundPageFields`. Fields not requested will
+/// be `None` to avoid unnecessary data transfer.
+#[derive(Serialize, Debug, Clone, PartialEq)]
+pub struct FoundPageRow {
+    pub page_id: i64,
```
emmiegit, line 311: You can't add it yet since you haven't added the `page_revision` join yet, but when you do, I think the field `page_revision_id` would be a good addition.

Additionally, I think you should add `page_category_id`, since that's something you can retrieve now.

deepwell/src/services/page_query/structs.rs
```diff
@@ -281,17 +278,65 @@ pub struct PageQuery<'a> {
...
+}
+
+/// The result of `PageQueryService::find()`.
+///
+/// Contains an ordered list of pages matching the query,
+/// with only the requested fields populated.
+#[derive(Serialize, Debug, Clone, PartialEq)]
+pub struct FoundPages {
+    pub pages: Vec<FoundPageRow>,
+    pub total: usize,
```
emmiegit: Since this is a derived property of `pages`, maybe it's better for it to be a method `pages`:
```suggestion
}

impl FoundPages {
    #[inline]
    pub fn total(&self) -> usize {
        self.pages.len()
    }
```

emmiegit (approved):

PR #2818, opened 2026-04-04
Add HTTP Range support to wws

jsthope:
This adds support for HTTP range requests in order to more optimally serve the new ftml blocks ``[[video]]`` and ``[[audio]]``.

This follow my [PR](https://github.com/scpwiki/ftml/pull/108) for [WJ-1041](https://scuttle.atlassian.net/browse/WJ-1041).

[WJ-1041]: https://scuttle.atlassian.net/browse/WJ-1041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wws/src/handler/file.rs
```diff
@@ -27,43 +27,248 @@ use crate::{
...
 };
-use axum_extra::response::Attachment;
 use s3::request::request_trait::ResponseDataStream;
+use std::fmt::Write;
 use wikidot_normalize::normalize;
 
-async fn fetch_file(
+const MULTIPART_BOUNDARY: &str = "wikijump_byteranges";
+
+// Reject requests with more than this many ranges to limit DoS surface
```
emmiegit: Make this and below into rustdocs:
```suggestion
/// Reject requests with more than this many ranges to limit DoS surface
```

wws/src/handler/file.rs
```diff
@@ -27,43 +27,248 @@ use crate::{
...
+const MAX_RANGES: usize = 10;
+
+// Maximum total bytes we'll buffer for a multipart/byteranges response.
+// Beyond this, the multipart request is rejected with 416
```
emmiegit:
```suggestion
/// Maximum total bytes we'll buffer for a `multipart/byteranges` response.
/// Beyond this, the multipart request is rejected with 416 (Range Not Satisfiable)
```

wws/src/handler/file.rs
```diff
@@ -27,43 +27,248 @@ use crate::{
...
+const MAX_MULTIPART_BYTES: u64 = 8 * 1024 * 1024;
```
emmiegit:
```suggestion
const MAX_MULTIPART_BYTES: u64 = 8 * 1024 * 1024; // 8 MiB
```

wws/src/handler/file.rs
```diff
@@ -27,43 +27,248 @@ use crate::{
...
+
+// HTTP Range support (see RFC 9110 §14)
+
+#[derive(Debug, Clone, Copy)]
+struct ByteRange {
+    start: u64,
+    end: u64,
+}
```
emmiegit: Noting that [`std::ops::Range<u64>`](https://doc.rust-lang.org/stable/std/ops/struct.Range.html) exists (but doesn't have a `len()` method).

wws/src/handler/file.rs
```diff
@@ -27,43 +27,248 @@ use crate::{
...
+                    start,
+                    end: end.min(file_size - 1),
+                });
+            }
+        } else {
+            return ParsedRange::None;
+        }
+
+        if ranges.len() > MAX_RANGES {
+            return ParsedRange::None;
```
emmiegit: I'd add an `error!` log for this case, something like
```rust
error!("Requested too many ranges, rejecting ({} > {})", ranges.len(), MAX_RANGES);
```

wws/src/handler/file.rs
```diff
@@ -27,43 +27,248 @@ use crate::{
...
+    None,
+
+    // One or more satisfiable byte ranges
+    Satisfiable(Vec<ByteRange>),
+
+    // Valid header but every range is unsatisfiable
+    NotSatisfiable,
+}
+
+fn parse_range_header(value: &str, file_size: u64) -> ParsedRange {
```
emmiegit: Add a comment linking the range request format? Would https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Range_requests#multipart_ranges be a good link?

wws/src/handler/file.rs
```diff
@@ -27,43 +27,248 @@ use crate::{
...
+            })
+            .collect();
+        let encoded = percent_encode_rfc5987(filename);
+        let value = format!(
+            "attachment; filename=\"{ascii_fallback}\"; filename*=UTF-8''{encoded}"
+        );
+        HeaderValue::from_bytes(value.as_bytes())
+            .unwrap_or_else(|_| HeaderValue::from_static("attachment"))
+    }
+}
```
emmiegit: Is there a reason you can't use `axum_extra::response::Attachment`?

If there is and `Attachment` is no longer being used, we should turn it off in the `axum-extras` dependency's features.

wws/src/handler/file.rs
```diff
@@ -27,43 +27,248 @@ use crate::{
...
+            | b'&'
+            | b'+'
+            | b'-'
+            | b'.'
+            | b'^'
+            | b'_'
+            | b'`'
+            | b'|'
+            | b'~' => {
+                out.push(*byte as char);
```
emmiegit: Better to avoid `as`:
```suggestion
                out.push(char::from(*byte));
```

wws/src/handler/file.rs
```diff
@@ -141,14 +677,141 @@ pub async fn handle_file_download(
...
+    fn suffix_larger_than_file() {
+        assert_eq!(ranges("bytes=-99999", 100), vec![(0, 99)]);
+    }
+
+    #[test]
+    fn multiple() {
+        assert_eq!(
+            ranges("bytes=0-99, 200-299", 12345),
+            vec![(0, 99), (200, 299)],
+        );
```
emmiegit: I think this one should have a few more cases. Without whitespace, open-ended, suffix, etc.

emmiegit (commented):
This is quite the PR, thanks for putting in all this work. I left a few comments, mostly smaller things.

wws/src/handler/file.rs
jsthope: ``axum_extra::response::Attachment`` doesn't compose well with the other headers we need. It builds a complete Response (body + Content-Type + Content-Disposition), but we now need to set many other headers around it: ETag, Accept-Ranges, Content-Range, Content-Length, etc. 
So using it would mean either mutating ``response.headers_mut()`` after the fact (kinda ugly :'/) or building the response manually anyway for the 206/416 paths.

wws/src/handler/file.rs
```diff
@@ -27,43 +27,255 @@ use crate::{
...
+                out.push(char::from(*byte));
+            }
+            _ => {
+                let _ = write!(out, "%{byte:02X}");
+            }
+        }
+    }
+    out
+}
+
```
emmiegit: With all the testing, I think the range parsing stuff is in a good state.

I didn't realize this was all in the route handler though, I think the last thing to do would be to split stuff out a bit. In the crate root you can see some helper modules focused on providing a particular utility.

So maybe like:
* `range.rs` - `ByteRange`, `parse_range_header`, the unit tests, etc
* `percent.rs` - `percent_encode_rfc5987`
* `attachment.rs` - `content_disposition_attachment` and its tests

Or another arrangement you find to be a natural set of divisions.

wws/src/handler/file.rs
```diff
@@ -27,43 +27,255 @@ use crate::{
...
+    },
     response::{IntoResponse, Response},
 };
-use axum_extra::response::Attachment;
 use s3::request::request_trait::ResponseDataStream;
+use std::fmt::Write;
 use wikidot_normalize::normalize;
 
-async fn fetch_file(
+const MULTIPART_BOUNDARY: &str = "wikijump_byteranges";
```
emmiegit: What exactly does this do?

emmiegit (approved):
Assuming you feel confident that it's working properly when run locally, I think this is good to go after my comments.

wws/src/handler/file.rs
jsthope: It specifies the boundary string used in HTTP multipart/byteranges responses:

```
--wikijump_byteranges
Content-Type: image/png
Content-Range: bytes 0-99/12345

<first chunk data>
--wikijump_byteranges
Content-Type: image/png
Content-Range: bytes 200-299/12345

<second chunk data>
--wikijump_byteranges--
```

After some thought, it might be a good idea to add a random hex to avoid unlikely cases where it appears in the data. 

see [RFC 2046](https://www.rfc-editor.org/rfc/rfc2046.html#section-5.1)

emmiegit: I see, and I agree. You could do similar to how we generate tokens, where it starts with `wikijump_byteranges_` and then has random info at the end: https://github.com/scpwiki/wikijump/blob/b371e25831df81ca6984bf8e06abb3f7e08782e6/deepwell/src/services/session/service.rs#L104-L105

Though looking at this now, it probably should use [`append_string()`](https://docs.rs/rand/0.10.0/rand/distr/trait.SampleString.html#tymethod.append_string):

```rust
let mut token = String::with_capacity(config.session_token_prefix.len() + config.session_token_length);
token.push_str(config.session_token_prefix);
Alphanumeric.append_string(&mut rng, &mut token, config.session_token_length);
```

So then for this case:
```rust
const MULTIPART_BOUNDARY_PREFIX: &str = "wikijump_byteranges_";
const MULTIPART_BOUNDARY_RANDOM_LENGTH: usize = 16;

let multipart_boundary = {
    let mut rng = thread_rng();
    let mut buffer = String::with_capacity(MULTIPART_BOUNDARY_PREFIX.len() + MULTIPART_BOUNDARY_RANDOM_LENGTG);
    buffer.push_str(MULTIPART_BOUNDARY_PREFIX);
    Alphanumeric.append_string(&mut rng, &mut buffer, MULTIPART_BOUNDARY_RANDOM_LENGTH);
    buffer
};
```

emmiegit: Also we should add a rustdoc explaining the boundary constant. A link to the RFC is probably sufficient.

jsthope:
Does this file structure suit you or did you envision something else @emmiegit ?

wws/src/handler/file.rs
```diff
@@ -20,87 +20,252 @@
...
+        }
+        ParsedRange::Satisfiable(ranges) => {
+            let total: u64 = ranges.iter().map(|r| r.len()).sum();
+            if total > MAX_MULTIPART_BYTES {
+                return build_or_500(
+                    Response::builder()
+                        .status(StatusCode::RANGE_NOT_SATISFIABLE)
+                        .header(header::CONTENT_RANGE, format!("bytes */{file_size}"))
+                        .header(header::ACCEPT_RANGES, "bytes")
+                        .body(Body::empty()),
```
emmiegit: This is repeated twice, maybe a `range_not_satisfiable()` helper function?

wws/src/attachment.rs
```diff
@@ -0,0 +1,87 @@
...
+    if filename.is_ascii() {
+        let escaped = filename.replace('\\', "\\\\").replace('"', "\\\"");
+        let value = format!("attachment; filename=\"{escaped}\"");
+        HeaderValue::from_str(&value)
+            .unwrap_or_else(|_| HeaderValue::from_static("attachment"))
+    } else {
+        let ascii_fallback: String = filename
+            .chars()
+            .map(|c| {
+                if c.is_ascii_graphic() && c != '"' && c != '\\' {
```
emmiegit, line 64: BTW, why can't `filename` have non-ASCII in it?

wws/src/range.rs
```diff
@@ -0,0 +1,280 @@
...
+
+#[derive(Debug, Clone, Copy)]
+pub struct ByteRange {
+    pub start: u64,
+    pub end: u64,
+}
+
+impl ByteRange {
+    pub fn len(self) -> u64 {
+        self.end - self.start + 1
```
emmiegit, line 38: Would benefit from a `debug_assert!(self.start < self.end)`.

emmiegit (approved):
Looks good to me, just some minor comments. Go ahead and merge when done!

wws/src/attachment.rs
jsthope: Non-ASCII filenames are supported via RFC 5987. See [line 41](https://github.com/scpwiki/wikijump/pull/2818/changes/BASE..c0dc216d4a13fe16da44fa459bece88f646613ff#diff-a2cccea9c7519f44c566a35b3f880c50c72dedfe4c13ac91b4b933a2f1734cc1R41) where we UTF-8 encode the original filename and [line 43](https://github.com/scpwiki/wikijump/pull/2818/changes/BASE..c0dc216d4a13fe16da44fa459bece88f646613ff#diff-a2cccea9c7519f44c566a35b3f880c50c72dedfe4c13ac91b4b933a2f1734cc1R43) where we provide both ASCII fallback and UTF-8 parameters for maximum compatibility :)

jsthope: I'll probably delete percent.rs and move it to attachment.rs to make things clearer

PR #2819, opened 2026-04-04
[WJ-1361] implement PageQueryService::select() with SelectedFields

0x5267:
- Add `SelectedFields`, `SelectedField`, `SelectedPageRow`, and `SelectedPages` types
- Implement `PageQueryService::select()` to extract requested fields from `FoundPages`
- Handles category extraction from slug, date formatting, hidden tag filtering
- Unimplemented fields (`score_votes`, `revisions`, `comments`, `children`, `size`) return null for now

### DIFF
--- a/deepwell/src/services/page_query/service.rs
+++ b/deepwell/src/services/page_query/service.rs
@@ -27,7 +27,7 @@ use crate::models::page_connection::{self, Entity as PageConnection};
 use crate::models::page_parent::{self, Entity as PageParent};
 use crate::models::{page_revision, text};
 use crate::services::{PageService, ParentService};
-use crate::utils::get_category_name;
+use crate::utils::{get_category_name, split_category};
 use sea_query::{Expr, Query};
 
 #[derive(Debug)]
@@ -513,14 +513,16 @@ impl PageQueryService {
                 } else {
                     None
                 },
-                slug: if fields.contains(SelectedField::PageSlug)
-                    || fields.contains(SelectedField::FullSlug)
-                {
+                full_slug: if fields.contains(SelectedField::FullSlug) {
                     row.slug.clone()
                 } else {
                     None
                 },
-                // Extract the category prefix from the slug (e.g. "scp" from "scp:scp-173").
+                page_slug: if fields.contains(SelectedField::PageSlug) {
+                    row.slug.as_deref().map(|s| split_category(s).1.to_owned())
+                } else {
+                    None
+                },
                 category: if fields.contains(SelectedField::Category) {
                     row.slug.as_deref().map(|s| get_category_name(s).to_owned())
                 } else {
@@ -551,7 +553,6 @@ impl PageQueryService {
                 } else {
                     None
                 },
-                // Hidden tags are those starting with '_'.
                 hidden_tags: if fields.contains(SelectedField::HiddenTags) {
                     row.tags.as_ref().map(|tags| {
                         tags.iter()
--- a/deepwell/src/services/page_query/service.rs
+++ b/deepwell/src/services/page_query/service.rs
@@ -27,6 +27,7 @@ use crate::models::page_connection::{self, Entity as PageConnection};
 use crate::models::page_parent::{self, Entity as PageParent};
 use crate::models::{page_revision, text};
 use crate::services::{PageService, ParentService};
+use crate::utils::get_category_name;
 use sea_query::{Expr, Query};
 
 #[derive(Debug)]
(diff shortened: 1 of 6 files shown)


deepwell/src/services/page_query/service.rs
```diff
@@ -483,4 +483,169 @@ impl PageQueryService {
...
+            .map(|row| {
+                let mut values = serde_json::Map::new();
+
+                for field in &fields.0 {
+                    let (key, value) = match field {
+                        SelectedField::PageId => {
+                            ("page_id", serde_json::Value::from(row.page_id))
+                        }
+                        SelectedField::SiteId => {
+                            ("site_id", serde_json::Value::from(row.site_id))
```
emmiegit: I have been using `use serde_json::Value as JsonValue;` as the import for this type.

Since you also use `serde_json::Map` you can import that as `JsonMap`.

deepwell/src/services/page_query/structs.rs
```diff
@@ -352,3 +352,63 @@ impl FoundPages {
...
+/// page query results. Each variant corresponds to a ListPages
+/// output variable like `%%title%%` or `%%created_at%%`.
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
+pub struct SelectedFields(pub Vec<SelectedField>);
+
+/// A single field to select from the query results.
+///
+/// Mirrors `PageQueryVariables` but is serializable and
+/// does not carry lifetime parameters.
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq, Hash)]
```
emmiegit:
```suggestion
#[derive(Deserialize, Serialize, Debug, Copy, Clone, PartialEq, Eq, Hash)]
```

deepwell/src/services/page_query/structs.rs
```diff
@@ -352,3 +352,63 @@ impl FoundPages {
         self.pages.len()
     }
 }
+
+/// Specifies which display fields the caller wants from the
+/// page query results. Each variant corresponds to a ListPages
+/// output variable like `%%title%%` or `%%created_at%%`.
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
+pub struct SelectedFields(pub Vec<SelectedField>);
```
emmiegit: Instead of a `Vec`, which has properties we don't need/want (O(n) membership, ordering, duplicates, stack allocation), I think the right type of choice here is a bitset.

I think [`enum_bitset`](https://docs.rs/enum-bitset/latest/enum_bitset/index.html) is a viable choice.

deepwell/src/services/page_query/service.rs
```diff
@@ -483,4 +483,169 @@ impl PageQueryService {
 
         Ok(FoundPages { pages: rows })
     }
+
+    /// Takes the output of `find()` and extracts only the fields
+    /// specified in `SelectedFields`, producing a `SelectedPages`
+    /// result with each page's values as a JSON map.
+    pub fn select(found: &FoundPages, fields: &SelectedFields) -> Result<SelectedPages> {
```
emmiegit:
```suggestion
    pub fn select(found: FoundPages, fields: SelectedFields) -> Result<SelectedPages> {
```

1. `SelectedFields` should be a bitset (which is an integer, which is `Copy`) (see other comments)
2. `FoundPages` should be passed owned, imo. It takes the values from fields (in particular strings), and if a caller needs to run multiple `select()` calls, they can run clone themselves.

deepwell/src/services/page_query/service.rs
```diff
@@ -483,4 +483,169 @@ impl PageQueryService {
...
+        info!(
+            "Selecting {} fields from {} pages",
+            fields.0.len(),
+            found.total()
```
emmiegit:
```suggestion
            found.total(),
```

deepwell/src/services/page_query/service.rs
```diff
@@ -483,4 +483,169 @@ impl PageQueryService {
...
+                                .slug
+                                .as_deref()
+                                .and_then(|s| s.split(':').next())
+                                .filter(|_| {
+                                    row.slug
+                                        .as_deref()
+                                        .map(|s| s.contains(':'))
+                                        .unwrap_or(false)
+                                })
+                                .unwrap_or("_default");
```
emmiegit: Don't duplicate logic, you should be using `split_category()` (in `crate::utils`) here.

deepwell/src/services/page_query/service.rs
```diff
@@ -483,4 +483,169 @@ impl PageQueryService {
...
+                            ("size", serde_json::Value::Null) // TODO
+                        }
+                    };
+
+                    values.insert(key.to_owned(), value);
+                }
+
+                SelectedPageRow {
+                    page_id: row.page_id,
+                    values,
```
emmiegit: I don't really understand, why use a JSON type here? Why not put it in normal rust struct form and have serde do the conversion to JSON?

deepwell/src/services/page_query/service.rs
```diff
@@ -483,4 +484,107 @@ impl PageQueryService {
...
+                } else {
+                    None
+                },
+                alt_title: if fields.contains(SelectedField::AltTitle) {
+                    row.alt_title
+                } else {
+                    None
+                },
+                slug: if fields.contains(SelectedField::PageSlug)
+                    || fields.contains(SelectedField::FullSlug)
```
emmiegit: Isn't this conflating two different things? Like for some page `system:recent-changes`, then:
* full slug is `system:recent-changes` (`slug`)
* category slug is `system` (`get_category_name(slug)`)
* page slug is `recent-changes` (`split_category(slug).1`)

emmiegit: Though honestly I feel like the better approach is to have these be separate fields in `SelectedPageRow` which has extracted fields map 1-to-1 to these fields.

deepwell/src/services/page_query/service.rs
```diff
@@ -513,14 +513,16 @@ impl PageQueryService {
...
                     None
                 },
-                // Extract the category prefix from the slug (e.g. "scp" from "scp:scp-173").
+                page_slug: if fields.contains(SelectedField::PageSlug) {
+                    row.slug.as_deref().map(|s| split_category(s).1.to_owned())
+                } else {
+                    None
+                },
                 category: if fields.contains(SelectedField::Category) {
                     row.slug.as_deref().map(|s| get_category_name(s).to_owned())
```
emmiegit, line 527: This is closer, though I was saying I think we should also modify the struct type `row` uses to have `full_slug`, `page_slug`, and `category`, so here we're just moving the field instead of processing it.

(That is, this processing, e.g. `get_category_name()`, is done when building `SelectedPageRow` instead of here)

PR #2823, opened 2026-04-05
[WJ-1373] Don't use low region for users

emmiegit:
Since Wikidot uses its compatibility IDs all the way down to 1, we cannot use the low region for our special users. Instead, we will use the negative ID range, since that will never get used by the normal user creation flow.

As part of this, I also ensure that seeding begins at these compatibility IDs:

image

I also take the opportunity here to formally enforce in code the 0 ID invariant (user ID 0 is never a valid user).

### DIFF
--- a/deepwell/Cargo.lock
+++ b/deepwell/Cargo.lock
@@ -947,7 +947,7 @@ dependencies = [
 
 [[package]]
 name = "deepwell"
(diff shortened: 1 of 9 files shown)


Syxless (approved):
LGTM!

Zokhoi (approved):

PR #2824, opened 2026-04-06
Framerail: Fix some param mismatch between forms and API

loop-index:
During testing, I noticed that revision rollback was not working and tracked it down to a mismatch between form submission schema and API schema. Combed through the form action list and made some changes wherever I noticed. 

Also created a error struct so typescript would stop complaining.


emmiegit (approved):
LGTM

PR #2840, opened 2026-04-11
[WJ-1367] Migrate existing UserPermissions stubs to new permission system

loop-index:
Cleaned up the existing UserPermissions stubs to use PermissionService instead. Additionally, implemented a batch permission check to query multiple permissions at once.

Closes [WJ-1367](https://scuttle.atlassian.net/browse/WJ-1366)

[WJ-1367]: https://scuttle.atlassian.net/browse/WJ-1367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deepwell/seeder/roles.json
+++ b/deepwell/seeder/roles.json
@@ -30,7 +30,8 @@
             "page:delete",
             "page:rename",
             "role:view",
-            "role:assign"
+            "role:assign",
+            "site:edit"
         ]
     },
     {
--- a/deepwell/src/services/permission/service.rs
+++ b/deepwell/src/services/permission/service.rs
@@ -230,44 +230,27 @@ impl PermissionService {
     pub async fn check_user_can(
         ctx: &ServiceContext<'_>,
         perm_ctx: &CheckPermissionContext<'_>,
-        PermissionInput {
-            resource_type,
-            resource_category,
-            action,
-        }: PermissionInput<'_>,
+        input: PermissionInput<'_>,
     ) -> Result<bool> {
+        let [result] = Self::batch_check_user_can(ctx, perm_ctx, [input]).await?;
+        Ok(result)
+    }
+
+    /// Batch check if a user has the specified permissions.
+    ///
+    /// Returns an array of booleans corresponding to each permission input.
+    /// Results are returned in the same order as the input array, best used with destructuring.
+    pub async fn batch_check_user_can<const N: usize>(
+        ctx: &ServiceContext<'_>,
+        perm_ctx: &CheckPermissionContext<'_>,
+        permissions: [PermissionInput<'_>; N],
+    ) -> Result<[bool; N]> {
         let user_id = perm_ctx.user_id;
         let site_id = perm_ctx.site_id;
         let page_reference = perm_ctx.page_reference.clone();
 
-        // Resolve category reference to ID for permission checking
-        let resource_category_id = match resource_category {
-            Some(reference) => {
-                resolve_category_reference(ctx, site_id, resource_type, reference).await?
-            }
-            None => None,
-        };
-
-        info!(
-            "Checking permission for user ID {:?} on site ID {} for resource {} of category {} with action {}",
-            user_id,
-            site_id,
-            resource_type,
-            resource_category_id.unwrap_or(0),
-            action
-        );
-
-        let make_error = || {
-            Error::new(
-                format!(
-                    "failed to check permission on resource {}:{} with action {}",
-                    resource_type,
-                    resource_category_id.unwrap_or(0),
-                    action
-                ),
-                ErrorType::Permission,
-            )
-        };
+        let make_error =
+            || Error::new("failed to check permissions", ErrorType::Permission);
 
         // Get all roles for this user (includes virtual roles)
         let role_ids: Vec<i64> = RoleService::get_all_roles_for_user_and_site(
@@ -284,8 +267,9 @@ impl PermissionService {
         .map(|r| r.role_id)
         .collect();
 
+        // Short-circuit: no roles means no permissions.
         if role_ids.is_empty() {
-            return Ok(false);
+            return Ok([false; N]);
         }
 
         // Lambda to check if user has any of the specified roles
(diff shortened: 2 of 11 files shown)


deepwell/seeder/roles.json
```diff
@@ -30,7 +30,8 @@
             "page:delete",
             "page:rename",
             "role:view",
-            "role:assign"
+            "role:assign",
+            "site:edit"
```
emmiegit, line 34: Off-topic: I guess `site:view` would be how a potential "private site" feature would be impleented?

deepwell/src/services/permission/service.rs
```diff
@@ -230,44 +230,27 @@ impl PermissionService {
...
     ) -> Result<bool> {
+        let [result] = Self::batch_check_user_can(ctx, perm_ctx, [input]).await?;
+        Ok(result)
+    }
+
+    /// Batch check if a user has the specified permissions.
+    ///
+    /// Returns an array of booleans corresponding to each permission input.
+    /// Results are returned in the same order as the input array, best used with destructuring.
+    pub async fn batch_check_user_can<const N: usize>(
```
emmiegit, line 243: I didn't know we could do this in stable now! Cool

deepwell/src/services/view/service.rs
```diff
@@ -263,7 +278,20 @@ impl ViewService {
...
+                            ctx,
+                            GetSiteBan {
+                                site_id,
+                                user_id: session.user.user_id,
+                            },
+                        )
+                        .await
+                        .or_raise(make_error)?,
+                        None => false,
+                    };
```
emmiegit, line 295: We should add a TODO here for IP bans (not implemented yet).

deepwell/tests/permission.rs
```diff
@@ -0,0 +1,480 @@
...
+use deepwell::services::user::{CreateUser, UserService};
+use deepwell::types::{Action, Reference, Resource, UserType};
+use serde_json::json;
+
+static FIXTURE_COUNTER: AtomicU64 = AtomicU64::new(0);
+const TEST_CATEGORY_NAME: &str = "test-category";
+
+fn next_n() -> u64 {
+    FIXTURE_COUNTER.fetch_add(1, Ordering::Relaxed)
+}
```
emmiegit, line 47: Each `#[tokio::test]` / `#[test]` runs as a separate test process, so I think `next_n()` at the start of a test is always going to return `0`.

deepwell/tests/permission.rs
```diff
@@ -0,0 +1,480 @@
...
+
+async fn create_user(ctx: &ServiceContext<'_>, fixture_n: u64, label: &str) -> i64 {
+    UserService::create(
+        ctx,
+        CreateUser {
+            user_type: UserType::Regular,
+            name: format!("Perm Test {fixture_n} {label}"),
+            email: format!("perm-{fixture_n}-{label}@email.com"),
+            locales: vec!["en".to_string()],
```
emmiegit:
```suggestion
            locales: vec![str!("en")],
```
is shorter

deepwell/tests/permission.rs
```diff
@@ -0,0 +1,480 @@
...
+            override_user_id: None,
+            ip_address: common::IP_ADDRESS,
+        },
+    )
+    .await
+    .expect("Failed to create test user")
+    .user_id
+}
+
+async fn check(
```
emmiegit, line 213:
```suggestion
#[must_use]
async fn check(
```

deepwell/tests/permission.rs
```diff
@@ -0,0 +1,480 @@
...
+            resource_type: resource,
+            resource_category: category_id.map(Reference::Id),
+            action,
+        },
+    )
+    .await
+    .expect("Permission check returned an error")
+}
+
+async fn batch_check<const N: usize>(
```
emmiegit, line 239:
```suggestion
#[must_use]
async fn batch_check<const N: usize>(
```

deepwell/tests/permission.rs
```diff
@@ -0,0 +1,480 @@
...
+        check(&runner, b, f.site_id, Resource::Page, cat, Action::Edit).await,
+        "user_b: should pass page:edit check in test-category"
+    );
+    // unscoped edit should fail
+    assert!(
+        !check(&runner, b, f.site_id, Resource::Page, None, Action::Edit).await,
+        "user_b: should fail page:edit without category"
+    );
+    // edit in other category should fail
+    let other_cat = Some(f.category_id + 9999);
```
emmiegit: I'm inclined to say that the test fixture should explicitly create an "other category" rather than just using a different ID, but this should work fine.

deepwell/tests/permission.rs
```diff
@@ -0,0 +1,480 @@
...
+            user_a,
+            user_b,
+            user_c,
+        }
+    }
+}
+
+// Test helpers
+
+async fn create_role(ctx: &ServiceContext<'_>, site_id: i64, name: &str) -> i64 {
```
emmiegit, line 137: Out of scope for this PR obviously, but we do need to create JSONRPC methods for these new role operations.

deepwell/tests/permission.rs
```diff
@@ -0,0 +1,480 @@
...
+            page_edit_permission,
+            json!({
+                "user_id": f.user_a,
+                "site_id": f.site_id,
+                "page": page.page_id,
+            }),
+        )
+        .can_edit,
+        "user_a should NOT have edit permission for page in test-category"
+    );
```
emmiegit, line 508: Nitpick: Could you also replicate these two tests, but passing in page slug instead of ID?

deepwell/seeder/roles.json
loop-index: To be fair, I'm kinda on the fence about gating admin panel being gated behind site:edit, because it's unclear what kind of resource it is. Is it a Page with a special category? I'm only putting it behind Site:Edit because logically you need that permission to reach admin console, but as you said when there's more Site: permissions then it might get confusing.

deepwell/tests/permission.rs
loop-index: I'm using an `AtomicU64` so it can be shared between tests.

deepwell/tests/permission.rs
loop-index: Yeah, I'll do that as part of a role operations/UX PR.

deepwell/tests/permission.rs
emmiegit: I was under the false impression that each test was a separate process, which I learned in creating an example test setup to elaborate on my comment.

```
running 4 tests
alpha
pid: 1051808
counter: 0
beta
pid: 1051808
counter: 1
test alpha ... ok
delta
pid: 1051808
counter: 2
gamma
pid: 1051808
counter: 3
test beta ... ok
test delta ... ok
test gamma ... ok

```

So this is fine, apologies for the confusion.

deepwell/seeder/roles.json
emmiegit: I would honestly say the admin panel should be its own permission entirely. `admin:manage`? `site:admin`? idk but it shouldn't be hitched onto something else.

Actually probably, `admin:` as a resource and then each _section_ of the panel is its own thing? Like `admin:ban`, `admin:forum`, `admin:permissions`, etc.?

emmiegit (approved):

deepwell/seeder/roles.json
loop-index: Why would each section of the admin panel be gated differently? Would multiple roles be able to access this resource? In that case, I think something like `SiteSettings` would be more appropriate, since "Admin Panel" kinda tie down the notion that there is an admin role.

emmiegit: Perhaps my thinking on this is being shaped too much by the way Wikidot currently does it. So to step through it:

My suggestion around splitting it up was since we have that ability, we could help sites refine their permissions more. (Since presently it's one big permission tied into wikidot admin)

The Wikidot admin panel has a variety of things, but they're all the sort of "biggest items". They can promote/demote moderators, accept new site member applications, change core site settings, change permissions, etc. It's a collection of various things, and I can see cases where you'd want to split it up.

If the issue is _naming_, then sure, I'm fine not using the word "admin" due to existing baggage. It could be "site panel"? "Site settings panel"?

PR #2850, opened 2026-04-18
[WJ-1368] Implement role hierarchy logic

loop-index:
Implement a tree-based role hierarchy system. A role hierarchy would look something like this:
```
Root
├─ Admin
│  ├─ Moderator
│  │  ├─ Anti-Harassment Staff
│  ├─ Editor
│  │  ├─ Rewrite Staff
│  │  ├─ Deletions Staff
│  ├─ Member
│  │  ├─ Guest
│  │  ├─ Everyone
(example only)
```
**Constraint**: Each role's permission set must be a subset of its ancestors' and a superset of its descendants'.

With that in mind, I have defined a series of operations on roles:
- Role hierarchy view:
  - Get roles for site (with basic info)
  - Add role, must attach to an existing role
  - Remove role, must either be a leaf node, or specify `reparent_children` to reparent children to its parent
  - Reparent role, move one node (and its subtree) to a new parent. Runs several validations:
    - Check that new parent's permission set must be a superset of this node's permission set.
    - Check that it's not a cycle (new parent must not be a descendant of this node)
- Role view:
  - Get role permissions, with a `human_readable_categories` flag to additionally resolve category IDs back to human-friendly slugs
  - Edit role info
  - Edit role permissions, also perform subset check to ensure new permission set does not violate hierarchy constraints.
  - Grant role to user
  - Revoke role from user

I actually have written the code for role endpoints and corresponding tests, but that will add like another thousand lines to this PR, so omitting for ease of review.


deepwell/src/database/seeder/mod.rs
```diff
@@ -541,50 +541,52 @@ pub async fn seed(state: &ServerState) -> Result<()> {
...
+                        resource_type: Resource::from_str(resource)
+                            .or_raise(make_error)?,
+                        resource_category: None,
+                        action: Action::from_str(action).or_raise(make_error)?,
+                    },
+                    [resource, category_slug, action] => PermissionInput {
+                        resource_type: Resource::from_str(resource)
+                            .or_raise(make_error)?,
+                        resource_category: Some(Reference::from(*category_slug)),
+                        action: Action::from_str(action).or_raise(make_error)?,
```
emmiegit, line 588: Optional suggestion: Since you do this a few times, you could have a `parse!` macro or similar which does `T::from_str($value).or_raise(make_error)?` (or `$value.parse().or_raise(make_error)?` if the type inference isn't required?)

deepwell/src/services/audit/structs.rs
```diff
@@ -441,6 +462,30 @@ impl<'a> AuditEvent<'a> {
...
+                ref old_permissions,
+                ref new_permissions,
+            } => {
+                let old_perms_json =
+                    serde_json::to_string(old_permissions).or_raise(make_error)?;
+                let new_perms_json =
+                    serde_json::to_string(new_permissions).or_raise(make_error)?;
+
+                RawAuditEvent {
+                    event_type: "permissions.update",
```
emmiegit, line 477: Nitpick: All the event types are singular, which would make this `permission.update`, though maybe it sounds weird / we're considering "permissions" a plural-only noun? (e.g. scissors) Then again, we do have `PermissionService`. Unsure.

It's not a big deal but worth bringing up for discussion.

deepwell/src/services/permission/resolvers.rs
```diff
@@ -61,6 +68,24 @@ impl CategoryResolver for PageCategoryResolver {
...
+        use crate::services::CategoryService;
+
+        match reference {
+            Reference::Id(id) => {
+                let category =
+                    CategoryService::get_optional(ctx, site_id, Reference::Id(id))
+                        .await?;
+                Ok(category.map(|c| c.slug))
+            }
+            Reference::Slug(slug) => Ok(Some(slug.to_string())),
```
emmiegit, line 86: Nitpick: Given that `Reference::Slug` is a pass-through, this seems like a reasonable use for `Cow::Borrowed`.

That is, the method signature being
```rs
    async fn resolve_slug<'slug>(
        ctx: &ServiceContext<'_>,
        site_id: i64,
        reference: Reference<'slug>,
    ) -> Result<Option<Cow<'slug, str>>>;
```

deepwell/src/services/permission/service.rs
```diff
@@ -24,117 +24,208 @@ use crate::models::prelude::{Role, RolePermission};
...
+                .or_raise(make_error)?;
+            if !resolved_permissions.is_subset(&parent_perms) {
+                bail!(Error::new(
+                    format!(
+                        "role ID {} has permissions not present in parent role ID {}",
+                        role.role_id, parent_id,
+                    ),
+                    ErrorType::RoleHierarchyViolation {
+                        role_id: role.role_id,
+                        parent_role_id: parent_id,
```
emmiegit, line 108: Given that this error case is used in two cases (permissions not a subset of parent and not a superset of child), I think it would be worth adding an enum designating what exactly the hierarchy violation was.

Doesn't have to be in this PR.

deepwell/src/services/permission/service.rs
```diff
@@ -24,117 +24,208 @@ use crate::models::prelude::{Role, RolePermission};
...
+                .iter()
+                .map(|perm| {
+                    let resource_category_id =
+                        perm.resource_category.as_ref().and_then(|r| {
+                            if let Reference::Id(id) = r {
+                                Some(*id)
+                            } else {
+                                None
+                            }
+                        });
```
emmiegit, line 180: Nitpick: I think this is better as a `match`:

```suggestion
                        perm.resource_category.as_ref().and_then(|r| match r {
                            Reference::Id(id) => Some(*id),
                            _ => None,
                        });
```

deepwell/src/services/permission/service.rs
```diff
@@ -360,4 +487,89 @@ impl PermissionService {
...
+
+        // Remove permissions from child role
+        for perm in &to_remove {
+            let resource_category_id = perm.resource_category.as_ref().and_then(|r| {
+                if let Reference::Id(id) = r {
+                    Some(*id)
+                } else {
+                    None
+                }
+            });
```
emmiegit, line 536: Ditto.

deepwell/src/services/permission/service.rs
```diff
@@ -360,4 +487,89 @@ impl PermissionService {
...
+            .await
+            .or_raise(make_error)?;
+
+        for grandchild in grandchildren {
+            Box::pin(Self::cascade_permission_removals(
+                ctx,
+                site_id,
+                grandchild.role_id,
+                &to_remove,
+            ))
```
emmiegit, line 568: I'm curious, why is pinning necessary here?

deepwell/src/services/role/service.rs
```diff
@@ -66,6 +71,13 @@ impl RoleService {
             )
         };
 
+        // Validate parent role if specified
```
emmiegit, line 74:
```suggestion
        // Validate parent role (if specified) exists
```

deepwell/src/services/role/service.rs
```diff
@@ -74,41 +86,43 @@ impl RoleService {
...
-        Ok(CreateRoleOutput { role_id })
+        Ok(created_role)
     }
 
     pub async fn update(
         ctx: &ServiceContext<'_>,
-        site_id: i64,
-        reference: Reference<'_>,
         UpdateRoleInput {
+            site_id: _,
```
emmiegit, line 114: Why is `site_id` unused? Like in other cases where it's passed in but not strictly needed, it's used to verify that the object being acted on (in this case the role) belongs to that site, to catch issues where somehow someone makes a request for a different site then they are on.

deepwell/src/services/role/service.rs
```diff
@@ -470,6 +466,236 @@ impl RoleService {
...
+                    ErrorType::CyclicRoleViolation {
+                        role_id,
+                        parent_role_id: parent_id
+                    },
+                ));
+            }
+
+            // Validate that the new parent role exists
+            let parent_role =
+                Self::get(ctx, site_id, parent_id.into())
```
emmiegit, line 506: Does `RoleService::get()` not fail if the role is deleted? `PageService::get()` and others does this.

deepwell/src/services/role/service.rs
```diff
@@ -470,6 +466,236 @@ impl RoleService {
...
+                    ErrorType::RoleNotFound,
+                ));
+            }
+
+            let is_proper_subset =
+                Self::validate_child_role_subset_of_parent(ctx, role_id, parent_id)
+                    .await
+                    .or_raise(make_error)?;
+
+            // Hacky solution to avoid running the expensive cycle check.
```
emmiegit, line 524: I think this conditional (as explained in the comment) makes sense. What part is hacky to you?

deepwell/src/services/role/service.rs
```diff
@@ -470,6 +466,236 @@ impl RoleService {
...
+            .filter(
+                role::Column::SiteId
+                    .eq(site_id)
+                    .and(role::Column::DeletedAt.is_null()),
+            )
+            .all(txn)
+            .await
+            .or_raise(make_error)?;
+
+        // Hashmap for quick lookup.
```
emmiegit, line 647:
```suggestion
        // Hashmap for quick lookup.
        // role_id -> parent_role_id
```

deepwell/src/services/role/structs.rs
```diff
@@ -57,3 +108,10 @@ pub struct GetUserRolesInput<'a> {
     pub user_id: Option<i64>,
     pub page_reference: Option<Reference<'a>>,
 }
+
+#[derive(Deserialize, Debug, Clone)]
+pub struct GetRolePermissionsInput<'a> {
+    pub site_id: i64,
+    pub role_reference: Reference<'a>,
+    pub human_readable_categories: bool,
```
emmiegit, line 116: In serde, should this have a default or should it always be specified?

deepwell/src/services/audit/structs.rs
loop-index: I guess I can rename it to something like `role.update_permissions`?

deepwell/src/services/role/structs.rs
loop-index: I can add a default of `false` considering it's the less costly path. Clients that need it will have to specify it.

deepwell/src/services/permission/service.rs
loop-index: From the error message itself: ```
a recursive `async fn` call must introduce indirection such as `Box::pin` to avoid an infinitely sized future```

More details: https://rust-lang.github.io/async-book/07_workarounds/04_recursion.html

deepwell/src/services/role/service.rs
loop-index: Missed this, will add a validation

deepwell/src/services/role/service.rs
loop-index: Will add that check to `get` path.

deepwell/src/services/role/service.rs
loop-index: I think I was just uneasy that the condition rely on an assumption (that tree is always valid) and if that assumption breaks then we can run into unwanted behavior.

deepwell/src/services/audit/structs.rs
emmiegit: Hmm, actually yeah. I think that's better because a "permission" here is an entity attached to a role.

deepwell/src/services/permission/service.rs
emmiegit: Ah, because it is a recursive invocation here as you go between each level. Makes sense.

deepwell/src/services/role/service.rs
emmiegit: Fair enough.

deepwell/src/services/role/service.rs
loop-index: On further checking, `RoleService::get_optional` already does deletion filter. Will remove the redundant check

deepwell/src/services/permission/service.rs
loop-index: Do you mean add an enum to designate the type of HierarchyViolation (ie a field of this variant) or do you mean splitting it into 2 different variants? Because the way I'm representing it is:

Let's say hierarchy is A > B > C. 
- If A < B then throw `violation { role: B, parent: A }`
- If B < C then throw `violation { role: C, parent: B }`

That way, the violation narrows down to one meaning: "This child role has more permissions than its parent"

emmiegit: Oh I see! I completely missed this subtlety. I think this satisfies my concern.

deepwell/src/services/permission/service.rs
```diff
@@ -144,14 +232,49 @@ impl PermissionService {
             )
         };
 
-        let role_permissions = RolePermission::find()
+        let role_permissions_iter = RolePermission::find()
             .filter(role_permission::Column::RoleId.eq(role_id))
             .order_by_asc(role_permission::Column::ResourceType)
             .order_by_asc(role_permission::Column::ResourceCategoryId)
             .order_by_asc(role_permission::Column::Action)
             .all(txn)
```
emmiegit, line 240: Optional, but if you're doing `.into_iter()` anyways, maybe try `.stream(txn)` here as the output type instead (vs gathering into `Vec`).

deepwell/src/services/role/service.rs
```diff
@@ -74,55 +86,62 @@ impl RoleService {
...
                     "failed to update role ID {}, changed by user ID {}",
-                    role.role_id, updating_user_id,
+                    role_id, updating_user_id,
                 ),
                 ErrorType::Role,
             )
         };
 
+        // Validate that the role belongs to the site
+        Self::get(ctx, site_id, role_id.into())
```
emmiegit, line 135:
```suggestion
        let _role = Self::get(ctx, site_id, role_id.into())
```
like the other cases?

deepwell/tests/permission.rs
```diff
@@ -522,3 +542,122 @@ async fn check_permission_endpoint() {
...
+        CategoryService::get_or_create(runner.context(), f.site_id, OTHER_CATEGORY_NAME)
+            .await
+            .expect("Failed to create other page category")
+            .category_id;
+
+    let role = RoleService::create(
+        runner.context(),
+        CreateRoleInput {
+            site_id: f.site_id,
+            name: "Test Role".to_string(),
```
emmiegit, line 571: You can use `str!` for brevity:
```suggestion
            name: str!("Test Role"),
```

emmiegit (approved):
Implementation and tests look good to me.

deepwell/src/services/permission/service.rs
loop-index: Will address the nits in a follow up PR.

PR #2851, opened 2026-04-19
[WJ-1378] Extract common page data into separate preload request

Zokhoi:
Extracts "viewer" data into its own preload api request, such that common data like `site` and `user_session` will always be present in the `page.data` object, regardless of the current route having its own loader function or not.

This is also done in preparation for always-present page components such as client login status box.

Note: first 3 commits are unrelated bug fixes, viewing them by commit will be easier.


framerail/src/lib/server/load/user.ts
```diff
@@ -1,8 +1,6 @@
 import defaults from "$lib/defaults"
 
-import { parseAcceptLangHeader } from "$lib/locales"
 import { authGetSession } from "$lib/server/auth/getSession"
-import { getFileByHash } from "$lib/server/deepwell/file"
```
hoah2333, line 5: I think this import is accidentally deleted.

hoah2333 (approved):

PR #2852, opened 2026-04-19
[WJ-1379] Update generate-models script to apply custom fixes

emmiegit:
We use `sea-orm-cli` to generate our Sea-ORM entity files ("model files"), but unfortunately it has a few areas where it doesn't quite generate the right thing and we've needed to manually fix those items.

Unfortunately, as time has passed, the number of corrections has grown considerably, making modifying these files quite a pain.

This attempts to return us to a "run and forget" situation, where our new `generate-models.py` script applies these fixes on our behalf. The changes are:

1. Inserting the `#[serde]` attributes for serializing datetime objects properly.
2. Using our crate's enum types instead of generic string values.
3. Applying custom patches to the file, at present this is just `site.patch` to account for https://github.com/SeaQL/sea-orm/issues/2358.

This way, we should be able to trust the output of this script without further worry. However, this does mean that whenever we make changes to these files (e.g. adding a new enum), we will have to modify the `scripts/generate-models.py` file too.


deepwell/src/models/site.rs
```diff
@@ -69,7 +69,9 @@ pub enum Relation {
     PageRevision,
     #[sea_orm(has_many = "super::role::Entity")]
     Role,
-    #[sea_orm(has_many = "super::site_domain::Entity")]
+    #[sea_orm(has_many = "super::role_permission::Entity")]
+    RolePermission,
+    #[sea_orm(has_many = "super::role_permission::Entity")]
```
loop-index:
```suggestion
    #[sea_orm(has_many = "super::site_domain::Entity")]
```

deepwell/scripts/generate-models.py
```diff
@@ -0,0 +1,268 @@
...
+        # We're making these separate methods for readability
+        #
+        # Sure it would be faster to do it all in one pass,
+        # but given that this script is run on-demand as part
+        # of development, it is not part of any hot path.
+        self.insert_timestamp_fields()
+        self.replace_enum_types()
+
+    def save(self):
+        if self.current_line_count != self.original_line_count:
```
loop-index, line 94: [nit]: Not important for now, but isn't this check a bit flaky? Just in case we add some corrections in the future that remove lines, possibly causing line counts to not change. A simple `dirty` flag should do the trick.

loop-index (approved):
Just a minor comment, otherwise LGTM. Thanks for taking this up 🙌

deepwell/scripts/generate-models.py
emmiegit: Yeah, that's probably better. When I started this script, it only consistently _added_ lines, but now that's not the case. I should use a flag instead.

PR #2853, opened 2026-04-19
[WJ-1379] Add "modified" flag to generate-models script 

emmiegit:
Follow-up to #2852. We can use a flag to mark whether the contents are "dirty" or not, and use that instead of checking line counts.

Tested by rerunning locally (no changes).


loop-index (approved):
LGTM

PR #2861, opened 2026-04-21
[WJ-1386] Add basic configuration infrastructure for email

emmiegit:
This adds the new configuration fields and secrets needed for the eventual addition of email sending methods in EmailService.

I also avoided some redundancy with the caddy tests by reusing `Config::integration_testing()` there.


deepwell/config.example.toml
```diff
@@ -285,7 +285,7 @@ last-update-ms = 0
 
 # Alternate, more compact representation of the above:
 # rerender-skip = [
-#     { job-depth = 5, last-update-ms = 100 },
+#     { job-depth = 3, last-update-ms = 100 },
```
loop-index, line 288: Just curious, why this change?

deepwell/src/services/caddy/test.rs
```diff
@@ -39,66 +39,12 @@ fn build_config(main_domain: &str, files_domain: &str) -> Config {
...
-        mock_mailcheck: true,
-        presigned_path_length: 0,
-        presigned_expiry_secs: 0,
-        maximum_blob_size: 0,
-        maximum_avatar_size: 0,
-        maximum_message_subject_bytes: 0,
-        maximum_message_body_bytes: 0,
-        maximum_message_recipients: 0,
-    }
+    let mut config = Config::integration_testing();
```
loop-index, line 42: [nit] This works, but would it be clearer to have something like `Config::base_config()` to extend from?

loop-index (approved):
Minor comments, LGTM.

deepwell/config.example.toml
emmiegit: It's not a change, it's what the configuration has, but this comment was never updated:
https://github.com/scpwiki/wikijump/blob/a182db8095107b97c95d1d5a140bb124edd9c3db/deepwell/config.example.toml#L271-L272

deepwell/src/services/caddy/test.rs
emmiegit: What would `base_config()` do? I don't understand what's being suggested here.

loop-index: My bad, I thought this was reusing test config for prod or something.

PR #2862, opened 2026-04-23
[WJ-1374] Add wikidot_user table

emmiegit:
This adds the `wikidot_user` table to Wikijump. This table's schema was designed to represent the data we are able to retrieve from Wikidot, without loosening or compromising the regular table's constraints and typing.

However, because not every user-created object will be a Wikijump user, we instead move all of our foreign key relations to a new `known_user` table, which simply stores the user IDs of all users "known" in our system. For instance, page revisions imported from Wikidot will have `wikidot_user` rows and be present in `known_user`, but aren't Wikijump users (maybe they became inactive a long time ago and won't ever make an account).

What this means is `wikidot_user` can make a user ID "real" (can be used in wiki data), but only Wikijump users, i.e. those in `user` are "real" users who you can do things like send messages to, etc.

The two tables also have another relationship. You can have a row in `wikidot_user` _and_ in `user`, which represents the case where a user, originally imported from Wikidot, then activates their account on Wikijump. In this case, the `wikidot_user` row stays as it is, the same as when we fetched it from Wikidot. But the Wikijump row in `user` is what would be modified with any live user changes (like a Wikijump user created brand new) effectively shadowing (hiding) the `wikidot_user` row.

This PR adds these two tables, and modifies the schema to use the new foreign key relation. I've updated the code to appropriately insert into `known_user` when new users are created via `UserService::create()`. A similar process will happen in `ImportService` for adding `wikidot_user` rows (not done yet).

Another change not yet added in this PR is adding support for _retrieving_ Wikidot user data. This means that, at present, trying to get any user info on data imported from Wikidot will fail with "user not found".

Tested by running integration tests, basic operations there are working as expected.


loop-index (approved):
LGTM

PR #2863, opened 2026-04-23
[chore] Upgrade rand and argon2 in deepwell

emmiegit:
Companion to #2849 

I'm using a release candidate version of `argon2` since `0.6.0` is needed for upgrading `rand` but it's not fully out yet.

### DIFF
--- a/deepwell/src/services/password.rs
+++ b/deepwell/src/services/password.rs
@@ -20,10 +20,7 @@
 
 use super::prelude::*;
 use crate::utils::assert_is_csprng;
-use argon2::{
-    Argon2, PasswordHash, PasswordHasher, PasswordVerifier, password_hash::SaltString,
-};
-use rand::thread_rng;
+use argon2::{Argon2, PasswordHash, PasswordHasher, PasswordVerifier};
 use tokio::time;
 
 #[derive(Debug)]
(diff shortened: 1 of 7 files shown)


deepwell/src/services/password.rs
```diff
@@ -35,15 +32,9 @@ impl PasswordService {
...
-        // Create and verify CSPRNG
-        let mut rng = thread_rng();
-        assert_is_csprng(&rng);
-
-        // Create Argon-2 context, salt, and then hash the password
         let argon2 = Argon2::default();
-        let salt = SaltString::generate(&mut rng);
         let hash = argon2
-            .hash_password(password.as_bytes(), &salt)
+            .hash_password(password.as_bytes())
```
emmiegit, line 37: The new `.hash_password()` generates a salt for you. There is another method that can be used if one wishes to provide a salt themselves.

deepwell/src/services/password.rs
```diff
@@ -109,13 +100,9 @@ impl PasswordService {
...
 
     fn verify_internal(password: &str, hash: &str) -> Result<()> {
-        // Parse PHC string
-        let hash = PasswordHash::new(hash).map_err(convert_argon_error)?;
-
-        // Create Argon-2 context, then verify the password
         let argon2 = Argon2::default();
         argon2
-            .verify_password(password.as_bytes(), &hash)
+            .verify_password(password.as_bytes(), hash)
```
emmiegit, line 105: This can parse the password hash from the string for us now.

deepwell/src/services/password.rs
loop-index: Not too familiar with this crate, does this new version of argon handle salting internally?

loop-index (approved):
lgtm

PR #2868, opened 2026-04-25
Reformat rust imports by crate

emmiegit:
Compare with #2869 

This changes the import style in deepwell and wws to be by crate.

I'm not sure which style of imports is better for us to use going forward. I see reasons to use either, and I think I lean towards per-crate, but I wanted feedback on what people prefer.


emmiegit:
Consensus was to group by module.

PR #2869, opened 2026-04-25
Reformat rust imports by module

emmiegit:
Compare with #2868 

This changes the import style in deepwell and wws to be by module.

I'm not sure which style of imports is better for us to use going forward. I see reasons to use either, and I think I lean towards per-crate, but I wanted feedback on what people prefer.

### DIFF
--- a/wws/.rustfmt.toml
+++ b/wws/.rustfmt.toml
@@ -5,3 +5,6 @@ edition = "2024"
 max_width = 90
 use_field_init_shorthand = true
 use_try_shorthand = true
+
+# Enable when stabilized:
+# imports_granularity = "Module"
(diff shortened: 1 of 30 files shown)


wws/.rustfmt.toml
```diff
@@ -5,3 +5,6 @@ edition = "2024"
 max_width = 90
 use_field_init_shorthand = true
 use_try_shorthand = true
+
+# Enable when stabilized:
+# imports_granularity = "Module"
```
loop-index, line 10:
```suggestion
imports_granularity = "Module"
```

loop-index (approved):
Consensus was to group by module.

wws/.rustfmt.toml
emmiegit: No, this needs to stay commented. It would require we run formatting only via the nightly channel. I noted it in the config so that when it's stabilized we can easily uncomment it then.

PR #2870, opened 2026-04-27
[WJ-1369] Role management endpoints and tests

loop-index:
- Added role endpoints (CRUD, Grant/Revoke)
- Added tests
- Added a permission endpoint/function to return `DecoratedPermission`s. This is a list of permissions that a role has, with additional metadata on which can be added/removed to preserve hierarchy.

This is done in preparation of a rudimentary role management UI. I'll spend some time to design some mockups for it.


deepwell/src/services/role/service.rs
```diff
@@ -132,7 +132,7 @@ impl RoleService {
         };
 
         // Validate that the role belongs to the site
-        Self::get(ctx, site_id, role_id.into())
+        let _role = Self::get(ctx, site_id, role_id.into())
```
emmiegit, line 135: (Outside of the scope of this PR, just a thought)

Since we're doing this type of check in several places, maybe we should start adding an `assert_exists()` method or something like it.

deepwell/src/types/permissions.rs
```diff
@@ -77,3 +77,34 @@ impl FromStr for Permission {
...
+        }
+    };
+}
+
+// Define all valid permission types.
+define_permission_types! {
+    Page => [View, Edit, Create, Delete, Rename],
+    Role => [View, Edit, Assign],
+    Site => [View, Edit],
+}
```
emmiegit, line 110: Big fan of this macro

deepwell/tests/role.rs
```diff
@@ -0,0 +1,573 @@
...
+            "ip_address": common::IP_ADDRESS,
+        }),
+    );
+
+    assert_eq!(role.name, "Moderator");
+    assert_eq!(role.description, "A moderator role");
+    assert!(!role.is_virtual);
+
+    // Check that the role appears in the site role listing
+    let roles = run_endpoint!(runner, list_site_roles, json!({ "site_id": f.site_id }),);
```
emmiegit:
```suggestion
    let roles = run_endpoint!(runner, list_site_roles, json!({ "site_id": f.site_id }));
```

deepwell/tests/role.rs
```diff
@@ -0,0 +1,573 @@
...
+            "role_reference": child.role_id,
+            "new_permissions": [
+                { "resource_type": "page", "resource_category": null, "action": "view" },
+                { "resource_type": "page", "resource_category": null, "action": "create" }, // Not in parent
+            ],
+            "cascade_removals": false,
+            "updating_user_id": SYSTEM_USER_ID,
+            "ip_address": common::IP_ADDRESS,
+        }),
+    );
```
emmiegit, line 271: I really appreciate the addition of these tests.

emmiegit (approved):

deepwell/src/services/role/service.rs
loop-index: Will keep this in mind.

PR #2882, opened 2026-04-29
Add assert_exists() methods

emmiegit:
For each of the cases where we currently fetch an object, not for the sake of getting it but verifying that it exists (and returning an error if not).

I only added it for cases where we currently use this pattern.

### DIFF
--- a/deepwell/src/services/role/service.rs
+++ b/deepwell/src/services/role/service.rs
@@ -73,7 +73,7 @@ impl RoleService {
 
         // Validate parent role (if specified) exists
         if let Some(parent_id) = parent_role_id {
-            let _parent_role = Self::get(ctx, site_id, parent_id.into())
+            Self::assert_exists(ctx, site_id, parent_id.into())
                 .await
                 .or_raise(make_error)?;
         }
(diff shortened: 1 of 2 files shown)


deepwell/src/services/role/service.rs
```diff
@@ -132,7 +132,7 @@ impl RoleService {
         };
 
         // Validate that the role belongs to the site
-        let _role = Self::get(ctx, site_id, role_id.into())
+        Self::get(ctx, site_id, role_id.into())
```
loop-index:
```suggestion
        Self::assert_exists(ctx, site_id, role_id.into())
```

deepwell/src/services/role/service.rs
loop-index:
```suggestion
            Self::assert_exists(ctx, site_id, parent_id.into())
```

loop-index: idk why this is showing up as changing a blank line for me.

loop-index (approved):
Appreciate you doing this, I left some suggestions on some other spots that could use the new function too.

PR #2883, opened 2026-05-02
Permission logic enhancement

loop-index:
Made some changes to the permission category scoping logic, so it's more intuitive and straightforward.

Before:
- For a category, check if any roles have a permission scoped to that category
- If a list of such roles exists, intersect it with the user's roles
- If no such roles exist, fallback to checking if any roles have the unscoped permission
- Intersect with user's role to check
-> Convoluted, hard to grasp, user permissions is not source of truth

After:
- Get all user roles, then construct a set of all permissions for that user
- Category check: Since RolePermission schema is (site_id, resource_type, resource_category_id, action, ...) we can simplify the category scope check down to an EXIST check
- Check if user has permission with appropriate scope

Also changed cache logic to cache per-user per-permission. Works more intuitively now imo.

### DIFF
--- a/deepwell/migrations/20220906103252_deepwell.sql
+++ b/deepwell/migrations/20220906103252_deepwell.sql
@@ -989,6 +989,8 @@ CREATE TABLE role_permission (
     UNIQUE (site_id, role_id, resource_type, resource_category_id, action)
 );
 
+CREATE INDEX idx_site_perms ON role_permission (site_id, resource_type, resource_category_id, action);
+
 -- User role assignments (many-to-many)
 CREATE TABLE user_role (
     user_id BIGINT NOT NULL REFERENCES known_user(user_id),
--- a/deepwell/src/services/permission/cache.rs
+++ b/deepwell/src/services/permission/cache.rs
@@ -42,7 +42,7 @@ impl PermissionCache {
     /// Build Redis cache key to lookup user permissions for a specific site.
     fn site_user_key(site_id: Option<i64>, user_id: Option<i64>) -> String {
         format!(
-            "site:{}:user:{}",
+            "permission:site:{}:user:{}",
             site_id
                 .map(|id| id.to_string())
                 .unwrap_or(SITE_NOT_SET_KEY.to_owned()),
@@ -61,7 +61,10 @@ impl PermissionCache {
         let category_id_str = resource_category_id
             .map(|id| id.to_string())
             .unwrap_or(DEFAULT_CATEGORY_KEY.to_owned());
-        Cow::Owned(format!("{}:{}:{}", resource, category_id_str, action))
+        Cow::Owned(format!(
+            "permission:{}:{}:{}",
+            resource, category_id_str, action
+        ))
     }
 
     /// Check if an action should be cached.
@@ -129,7 +132,7 @@ impl PermissionCache {
     /// Invalidate the cache for a specific site.
     pub async fn invalidate_site(ctx: &ServiceContext<'_>, site_id: i64) -> Result<()> {
         let mut redis = ctx.redis();
-        let pattern = format!("site:{}:*", site_id);
+        let pattern = format!("permission:site:{}:*", site_id);
         let make_error = || {
             Error::new(
                 format!("Failed to invalidate permission cache for site {}", site_id),
--- a/deepwell/src/services/permission/cache.rs
+++ b/deepwell/src/services/permission/cache.rs
@@ -95,10 +95,7 @@ impl PermissionCache {
                 Error::new("Permission cache read error", ErrorType::Permission)
             })?;
 
-        Ok(match has_permission {
-            Some(val) => Some(val == "1"),
-            None => None,
-        })
+        Ok(has_permission.map(|val| val == "1"))
     }
 
     /// Set a user's permission value in the cache.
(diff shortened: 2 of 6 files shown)


deepwell/migrations/20220906103252_deepwell.sql
```diff
@@ -989,6 +989,8 @@ CREATE TABLE role_permission (
     UNIQUE (site_id, role_id, resource_type, resource_category_id, action)
 );
 
+CREATE INDEX idx_site_perms ON role_permission (site_id, resource_type, resource_category_id, action);
```
emmiegit, line 992: I'm curious if this is actually needed. When you do an `EXPLAIN` with and without, is it able to leverage the `UNIQUE` index above to do the same work, or is it slower?

deepwell/src/services/permission/cache.rs
```diff
@@ -31,22 +31,37 @@ use ftml::info;
...
 
 #[allow(dead_code)]
 impl PermissionCache {
-    /// Build Redis cache key for a permission hash.
-    fn key(site_id: i64, resource_type: Resource, action: Action) -> String {
-        format!("perm:{}:{}:{}", site_id, resource_type, action)
+    /// Build Redis cache key to lookup user permissions for a specific site.
+    fn site_user_key(site_id: Option<i64>, user_id: Option<i64>) -> String {
+        format!(
+            "site:{}:user:{}",
```
emmiegit: This format makes sense, but is no longer namespaced liked before. I think it makes sense to prepend both keys with `permission:`.
```suggestion
            "permission:site:{}:user:{}",
```

deepwell/src/services/permission/cache.rs
```diff
@@ -31,22 +31,37 @@ use ftml::info;
...
+    /// Build a hash field key for the permission
+    fn permission_key(
+        resource: Resource,
+        resource_category_id: Option<i64>,
+        action: Action,
+    ) -> Cow<'static, str> {
+        let category_id_str = resource_category_id
+            .map(|id| id.to_string())
+            .unwrap_or(DEFAULT_CATEGORY_KEY.to_owned());
+        Cow::Owned(format!("{}:{}:{}", resource, category_id_str, action))
```
emmiegit:
```suggestion
        Cow::Owned(format!("permission:{}:{}:{}", resource, category_id_str, action))
```

deepwell/src/services/permission/cache.rs
```diff
@@ -58,97 +73,63 @@ impl PermissionCache {
...
+            })?;
 
         Ok(())
     }
 
+    /// Invalidate the cache for a specific site.
     pub async fn invalidate_site(ctx: &ServiceContext<'_>, site_id: i64) -> Result<()> {
         let mut redis = ctx.redis();
-        let pattern = format!("perm:{}:*", site_id);
+        let pattern = format!("site:{}:*", site_id);
```
emmiegit:
```suggestion
        let pattern = format!("permission:site:{}:*", site_id);
```

deepwell/src/services/permission/resolvers.rs
```diff
@@ -53,16 +53,19 @@ impl CategoryResolver for PageCategoryResolver {
...
-            Reference::Id(id) => Ok(Some(id)),
+            Reference::Id(id) => Ok(Some(*id)),
             Reference::Slug(slug) => {
-                let category =
-                    CategoryService::get_optional(ctx, site_id, Reference::Slug(slug))
-                        .await?;
+                let category = CategoryService::get_optional(
+                    ctx,
+                    site_id,
+                    Reference::Slug(slug.clone()),
```
emmiegit: Nitpick:
```suggestion
                    Reference::Slug(cow!(&slug)),
```

emmiegit (approved):
I agree, I think this new approach is a lot more comprehensible / easy to understand. Thanks for this!

deepwell/migrations/20220906103252_deepwell.sql
loop-index: with:
```
 Limit  (cost=0.14..7.95 rows=1 width=4) (actual time=0.034..0.034 rows=0 loops=1)
   Buffers: shared hit=1
   ->  Index Only Scan using idx_site_perms on role_permission  (cost=0.14..7.95 rows=1 width=4) (actual time=0.033..0.033 rows=0 loops=1)
         Index Cond: ((site_id = 1) AND (resource_type = 'page'::text) AND (resource_category_id = 5) AND (action = 'view'::text))
         Heap Fetches: 0
         Buffers: shared hit=1
 Planning Time: 0.077 ms
 Execution Time: 0.061 ms
(8 rows)
```

without:
```
 Limit  (cost=0.27..4.29 rows=1 width=4) (actual time=0.045..0.045 rows=0 loops=1)
   Buffers: shared hit=5
   ->  Index Only Scan using role_permission_site_id_role_id_resource_type_resource_cate_key on role_permission  (cost=0.27..4.29 rows=1 width=4) (actual time=0.044..0.044 rows=0 loops=1)
         Index Cond: ((site_id = 1) AND (resource_type = 'page'::text) AND (resource_category_id = 5) AND (action = 'view'::text))
         Heap Fetches: 0
         Buffers: shared hit=5
 Planning:
   Buffers: shared hit=107
 Planning Time: 0.806 ms
 Execution Time: 0.083 ms
(10 rows)
```
Pretty interesting, I thought the excess `role_id` in the unique constraint index will break it out of the index scan or something but seems like postgres is able to resolve it. I think the only concern remaining is index size, since there's one more dimension to store.

PR #2893, opened 2026-05-06
[Experiment] Add request context to store common request metadata

loop-index:
Attempt to enhance request cycle by passing common request data (user session, site ID, page reference) out-of-band through request headers.

Request context flow:
- In Framerail, session token is fetched from cookies and optional site ID + page ref is passed down from frontend. `RequestContext` constructed from these data.
- Utilize JSONRPC client's `ClientParams` hook to receive this context and populate the appropriate HTTP headers
- Created a middleware layer on deepwell to extract these headers and reconstruct the `RequestContext` on backend. This is then persisted in the `Extensions` parameter of the JSONRPC module [docs](https://docs.rs/jsonrpsee-server/0.26.0/jsonrpsee_server/struct.ServerBuilder.html#method.set_http_middleware)
- During JSONRPC method registration, extract the request context from `Extensions` and store it in `ServiceContext`. Now all downstream services can access this data.

By making this an OOB flow, existing endpoints can be gradually migrated to this new pattern. I've included a proof-of-concept with `pageEditPermission` endpoint. It now passes no arguments and rely entirely on the RequestContext to carry necessary data.

This data also matches the permission context, so my future plan is to migrate the permission fetch entirely to pre-method as well.


deepwell/src/endpoints/page.rs
```diff
@@ -233,20 +233,28 @@ pub async fn page_edit(
 
 pub async fn page_edit_permission(
     ctx: &ServiceContext<'_>,
-    params: Params<'static>,
+    _params: Params<'static>,
 ) -> Result<PageEditPermissionOutput> {
-    let input: PageEditPermission = parse!(params, Page);
+    let request_ctx = ctx.request_context();
```
emmiegit: Nitpick: I think `request_context` is a bit redundant of a name given it's already part of a "context" object. Maybe just `request`?

deepwell/src/endpoints/page.rs
```diff
@@ -233,20 +233,28 @@ pub async fn page_edit(
...
+    let page_reference = request_ctx.page_reference().cloned().ok_or_else(|| {
+        Error::new(
+            "page reference is required for edit permission check",
+            ErrorType::Page,
+        )
+    })?;
```
emmiegit: Given that these "cannot proceed without required data" checks will be prevalent with this new context struct, we should make fetch methods for them. Like a `.page_ref()` that returns `Result` / error if not present.

deepwell/src/endpoints/page.rs
```diff
@@ -233,20 +233,28 @@ pub async fn page_edit(
...
+
     info!(
-        "Checking edit permission for page {:?} in site ID {}",
-        input.page, input.site_id,
+        "Checking edit permission for page {:?} in site ID {:?}",
+        page_reference,
+        request_ctx.site_id().unwrap()
```
emmiegit: Same deal here. Instead of doing `unwrap()` everywhere (which as a very strong assertion, we should avoid except for obvious cases), we should extract desired fields at the start with the error handling, and then use those bindings throughout.

deepwell/src/middleware.rs
```diff
@@ -0,0 +1,98 @@
...
+    S: Service<Request<Body>>,
+{
+    type Response = S::Response;
+    type Error = S::Error;
+    type Future = S::Future;
+
+    fn poll_ready(
+        &mut self,
+        cx: &mut Context<'_>,
+    ) -> Poll<std::result::Result<(), Self::Error>> {
```
emmiegit: Nitpick: We have the `StdResult` alias for the stdlib type.

deepwell/src/middleware.rs
```diff
@@ -0,0 +1,98 @@
...
+        self.service.poll_ready(cx)
+    }
+
+    fn call(&mut self, mut request: Request<Body>) -> Self::Future {
+        let session_token: Option<String> = request
+            .headers()
+            .get("X-Session-Token")
```
emmiegit: In wws, I've been making sure all internal headers are `X-Wikijump-` prefixed. I think it would be good for us to maintain something similar (maybe `X-Deepwell-`?) here even though these requests aren't ever from outside HTTP clients.

emmiegit (commented):
I never thought of using the HTTP middleware like this! I kept thinking of parsing out the JSONRPC body in some way, I think is a better solution.

I had some comments, mostly about use semantics for the added concepts.

emmiegit (approved):

PR #2896, opened 2026-05-11
[Enhancement] Consolidating request context propagation

loop-index:
Following up from #2893 

On the framerail side, I've added a request hook ([doc](https://svelte.dev/docs/kit/hooks)) that runs before every framerail request, gathering user token, site ID and page slug to persist into `ServerLoadEvent.locals`, allowing that context to be accessed throughout the request's call chain. This is possible because the required data is persisted on the browser through one way or another:
- `session_token` via cookies
- `site_id` via Caddy headers
- `page` slug via URL params

Additionally, I've implemented the proposed `user_has_permission` in `ServiceContext` as it now contains all necessary data.

Again as a POC, I've modified the `page_edit_permission` call chain to make use of this. Now the full chain becomes:
```
Browser calls `pageEditPermission` via form submission
↓
Framerail hook runs, populating RequestContext in locals
↓
pageEditPermissionAction gets the RequestContext from locals and passes it to deepwell's RPC client
↓
deepwell RPC client serializes RequestContext fields into headers, which is intercepted by deepwell middleware to inject into ServiceContext
↓
Downstream consumers access RequestContext from ctx
```
Request body no longer needs to pass these fields, and handlers no longer need to drill these values down the chain.

### DIFF
--- a/deepwell/src/middleware.rs
+++ b/deepwell/src/middleware.rs
@@ -80,6 +80,7 @@ where
             .get("X-Deepwell-Page")
             .and_then(|v| v.to_str().ok())
             .map(|s| {
+                // TODO: Figure out if this can cause a bug: if the slug is a number, it will be parsed as an ID instead of a slug.
                 s.parse::<i64>()
                     .map(Reference::Id)
                     .unwrap_or_else(|_| Reference::Slug(Cow::Owned(s.to_owned())))
--- a/deepwell/src/services/context.rs
+++ b/deepwell/src/services/context.rs
@@ -111,7 +111,7 @@ impl<'txn> ServiceContext<'txn> {
 
     #[inline]
     /// Internal method to update the request context, for use in testing only.
-    pub fn _set_request(&mut self, request_ctx: RequestContext) {
+    pub fn set_request_for_test(&mut self, request_ctx: RequestContext) {
         self.request_ctx = request_ctx;
 
         // Clear cached permissions since the user context has changed.
@@ -198,8 +198,10 @@ impl<'txn> ServiceContext<'txn> {
             || Error::new("Failed to check user permissions", ErrorType::Permission);
 
         let perms = self.user_permissions().await.or_raise(make_error)?;
-        PermissionService::_has_permission(self, user_id, perms, site_id, permission)
-            .await
-            .or_raise(make_error)
+        PermissionService::permission_in_set_helper(
+            self, user_id, perms, site_id, permission,
+        )
+        .await
+        .or_raise(make_error)
     }
 }
--- a/deepwell/src/services/context.rs
+++ b/deepwell/src/services/context.rs
@@ -24,12 +24,15 @@ use crate::error::prelude::*;
 use crate::locales::Localizations;
 use crate::models::session::Model as SessionModel;
 use crate::services::blob::MimeAnalyzer;
-use crate::types::Reference;
+use crate::services::permission::PermissionService;
+use crate::types::{Permission, Reference};
 use redis::aio::MultiplexedConnection as RedisMultiplexedConnection;
 use rsmq_async::Rsmq;
 use s3::bucket::Bucket;
 use sea_orm::DatabaseTransaction;
+use std::collections::HashSet;
 use std::sync::Arc;
+use tokio::sync::OnceCell;
 
 /// Per-request context derived from HTTP headers by the middleware layer.
 #[derive(Debug, Clone, Default)]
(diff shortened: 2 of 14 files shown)


loop-index:
Some tests are breaking, which is due to RequestContext being populated via middleware and not from the request body. Will think of a way to make this more ergonomic to use, because the alternative is every test will need to populate its own request ctx.

deepwell/src/services/permission/service.rs
```diff
@@ -422,110 +422,123 @@ impl PermissionService {
...
-                    continue;
-                } else {
-                    info!(
-                        "Cache miss for user ID {:?} on site ID {} for resource {} of category {:?} with action {}",
-                        user_id, site_id, resource, resource_category, action,
-                    );
-                }
-            }
+    /// Helper function to check if a permission is present in (the user's) permission set.
+    pub async fn _has_permission(
```
emmiegit: I'm a bit confused by the semantics of this function.

If it's a helper method for `PermissionService`, why is it `pub`? Do we expect other callers to use this?

Also, I don't think we use underscore prefixing to indicate an inner/helper method, but generally it being not-`pub` and having a name describing what the helper does. Maybe here, `has_specific_permission()`? Or something else to suggest that this is the internal check for one user/permission pair.

deepwell/src/services/context.rs
```diff
@@ -105,8 +110,12 @@ impl<'txn> ServiceContext<'txn> {
     }
 
     #[inline]
-    pub fn set_request(&mut self, request_ctx: RequestContext) {
+    /// Internal method to update the request context, for use in testing only.
+    pub fn _set_request(&mut self, request_ctx: RequestContext) {
```
emmiegit: If it's for testing then how about:

```suggestion
    pub fn set_request_for_test(&mut self, request_ctx: RequestContext) {
```

deepwell/src/middleware.rs
```diff
@@ -80,6 +80,7 @@ where
             .get("X-Deepwell-Page")
             .and_then(|v| v.to_str().ok())
             .map(|s| {
+                // TODO: Figure out if this can cause a bug: if the slug is a number, it will be parsed as an ID instead of a slug.
```
emmiegit, line 83: Oh good catch. This will definitely cause issues, hm.

I guess the solution is parsing off of a prefix?

```rs
if let Some(s) = v.strip_prefix("slug:") {
    return Reference::Slug(Cow::Owned(s.to_owned()));
}

if let Some(s) = v.strip_prefix("id:") {
    return s.parse::<i64>().map(Reference::Id)?;
}

// error
```

Though maybe this should also just be a case where we add this type of serialization logic to `Reference` itself.

Anyways, just some thoughts, doesn't have to be this PR of course.

deepwell/tests/page.rs
```diff
@@ -42,6 +43,15 @@ async fn basic_edit() {
     let site_id = output.site.site_id;
     assert_eq!(output.site.slug, SITE_SLUG, "Site slug doesn't match");
 
+    // Set request context to populate params for the internal permission check.
+    // TODO: Figure out a better way to handle this gap.
```
emmiegit: Honestly, I feel like this is fine. Instead of needing to set it again and again for each request, we set it per test, which seems like a reasonable balance between not repeating things and avoiding carry-over behavior between different tests.

framerail/src/hooks.server.ts
```diff
@@ -0,0 +1,20 @@
...
+  const { siteId } = loadSiteInfo(request.headers)
+  const page_slug = params.slug
+  const sessionToken = cookies.get("wikijump_token")
+
+  storeRequestContext(locals, sessionToken, siteId, page_slug)
+
+  // Continue processing the request
+  const response = await resolve(event)
+  return response
+}
```
emmiegit, line 20: Very cool way of handling this!

emmiegit (approved):
Looks good, small suggestions mostly about naming.

hoah2333 (approved):

deepwell/src/services/permission/service.rs
loop-index: I made it pub because I want to reuse it in `ServiceContext::user_has_permission`. I can rename it to `PermissionService::permission_in_set_helper` or something to make it clearer.

deepwell/src/middleware.rs
loop-index: That could work, but we would have to carry that convention anywhere we want to pass references between framerail and deepwell. An ideal solution would be a strongly typed transmission format (eg. Protobuf) instead of JSON, but that's probably too large of a change.

deepwell/src/services/permission/service.rs
emmiegit: You can make it `pub(crate)` so it can be used in other services but not outside the crate. That's probably the best way to do it, there are a handful of methods like that.

deepwell/src/middleware.rs
emmiegit: Mhmm, yeah. We'll have to think about it. And I do agree, a typed transmission format would be ideal, though I'm not sure how feasible it would be to actually do.

Zokhoi (approved):

emmiegit (approved):

PR #2913, opened 2026-05-18
[WJ-1265] Implement page locks

loop-index:
Implemented backend service for page locks.


deepwell/src/services/page_lock/service.rs
```diff
@@ -0,0 +1,333 @@
...
+                    .user_has_permission(Permission {
+                        resource_type: Resource::Page,
+                        resource_category: page_category_id.map(Reference::Id),
+                        action: Action::BypassLock,
+                    })
+                    .await
+                    .or_raise(make_error)?,
+                PageLockType::AuthorOrPermissionOnly => {
+                    // Check if the user is the author of the page
+                    let attributions = RelationService::get_page_attributions(
```
loop-index, line 281: Maybe UserRoles should be added to request ctx as well considering that permissions are being fetched already.

deepwell/src/endpoints/page_lock.rs
```diff
@@ -0,0 +1,104 @@
...
+    let request = ctx.request();
+    let site_id = request
+        .site_id()
+        .or_raise(|| Error::new("No site ID found", ErrorType::PageLock))?;
+    let user_id = request
+        .user_id()
+        .or_raise(|| Error::new("No user ID found", ErrorType::PageLock))?;
+    let page_ref = request
+        .page_reference()
+        .or_raise(|| Error::new("No page reference found", ErrorType::PageLock))?;
```
emmiegit: Nitpick: `exn` (and maybe Rust more broadly) error messages are supposed to start lowercase, since they're meant to be sentence fragments that can be included in a longer error message.

deepwell/src/services/page_lock/service.rs
```diff
@@ -0,0 +1,333 @@
...
+        let mut updated_reason = input.reason.unwrap_or_default();
+
+        if let Some(old_lock) = existing_lock {
+            if !input.override_existing {
+                bail!(Error::new(
+                    format!(
+                        "An active lock already exists for page {:?}, please remove it first.",
+                        page_ref
+                    ),
+                    ErrorType::PageLock
```
emmiegit: I think "lock already exists" should be a separate `ErrorType`. Certainly, we'd want a specific message for it in the UI.

deepwell/src/services/page_lock/service.rs
```diff
@@ -0,0 +1,333 @@
...
+        };
+
+        // Fetch the page to be locked
+        let page_id = match page_ref {
+            Reference::Id(page_id) => page_id,
+            _ => {
+                PageService::get(ctx, site_id, page_ref.borrow())
+                    .await
+                    .or_raise(make_error)?
+                    .page_id
```
emmiegit, line 59: FYI that `PageService::get_id()` exists to convert reference to page ID - however since we want to assert that the page actually exists here (since we're about to apply an object onto it), doing `PageService::get()` is fine.

deepwell/src/services/page_lock/service.rs
```diff
@@ -0,0 +1,333 @@
...
+                    ),
+                    ErrorType::PageLock
+                ));
+            } else {
+                // Update lock reason to include override message
+                updated_reason = [
+                    updated_reason,
+                    format!("Overriding previous {} lock", old_lock.lock_type),
+                ]
+                .join("\n");
```
emmiegit: Hmm, I'm not a fan of this. For one, it's hardcoded English in the lock reason, and two, I feel like the fact that this is an override should be inferrable from the lock history, right? If you see lock A -> B, then it was replaced. If you see lock A -> (none) -> B, then you see it was removed, and some time later a new, different lock was added.

deepwell/src/services/page_lock/service.rs
```diff
@@ -0,0 +1,333 @@
...
+
+        // Create the page lock
+        let new_lock = page_lock::ActiveModel {
+            page_id: Set(page_id),
+            user_id: Set(user_id),
+            lock_type: Set(input.lock_type),
+            reason: Set(updated_reason),
+            expires_at: Set(input.expires_at),
+            from_wikidot: Set(input.from_wikidot),
+            created_at: Set(OffsetDateTime::now_utc()),
```
emmiegit: Convenience function:
```suggestion
            created_at: Set(now()),
```

deepwell/src/services/page_lock/service.rs
```diff
@@ -0,0 +1,333 @@
...
+        // Resolve page reference to ID
+        let page_id = match page_ref {
+            Reference::Id(page_id) => page_id,
+            _ => {
+                PageService::get(ctx, site_id, page_ref.borrow())
+                    .await
+                    .or_raise(make_error)?
+                    .page_id
+            }
+        };
```
emmiegit: This is the ideal case for replacing with `PageService::get_id()` (don't resolve on ID, only slug).

deepwell/src/services/page_lock/service.rs
```diff
@@ -0,0 +1,333 @@
...
+
+        // Fetch the active lock to be removed
+        let maybe_lock = Self::get_active_lock_for_page(ctx, page_id)
+            .await
+            .or_raise(make_error)?;
+
+        // If no lock, return
+        let page_lock = match maybe_lock {
+            Some(lock) => lock,
+            None => return Ok(()),
```
emmiegit: Thought: Maybe this method should return an `Option`? `Ok(None)` if no lock to remove, `Ok(Some(_))` with the lock that was removed (idk which fields we'd want to include there though). Then again, maybe such information isn't useful so we shouldn't bother.

deepwell/src/services/page_lock/service.rs
```diff
@@ -0,0 +1,333 @@
...
+        // Fetch the page to get its ID
+        let page_id = match page_ref {
+            Reference::Id(page_id) => page_id,
+            _ => {
+                PageService::get(ctx, site_id, page_ref.borrow())
+                    .await
+                    .or_raise(make_error)?
+                    .page_id
+            }
+        };
```
emmiegit:
```suggestion
        let page_id = PageService::get_id(ctx, site_id, page_ref.borrow())
            .await
            .or_raise(make_error)?
            .page_id
        };
```

deepwell/src/services/page_lock/service.rs
```diff
@@ -0,0 +1,333 @@
...
+
+        // Fetch all historical locks for the page, including expired and deleted ones
+        let locks = PageLock::find()
+            .filter(page_lock::Column::PageId.eq(page_id))
+            .all(txn)
+            .await
+            .or_raise(make_error)?;
```
emmiegit, line 209: We should explicily order by lock ID; the history should always be in a consistent order, and the most logical one is chronological.

deepwell/src/services/page_lock/service.rs
```diff
@@ -0,0 +1,333 @@
...
+        let active_lock = PageLock::find()
+            .filter(
+                Condition::all()
+                    .add(page_lock::Column::PageId.eq(page_id))
+                    .add(page_lock::Column::DeletedAt.is_null())
+                    .add(
+                        Condition::any()
+                            .add(
+                                page_lock::Column::ExpiresAt
+                                    .gt(OffsetDateTime::now_utc()),
```
emmiegit:
```suggestion
                                page_lock::Column::ExpiresAt.gt(now()),
```

deepwell/src/services/page_lock/service.rs
emmiegit: This seems sensible to me.

deepwell/src/services/page_lock/service.rs
```diff
@@ -0,0 +1,333 @@
...
+                                action: Action::BypassLock,
+                            })
+                            .await
+                            .or_raise(make_error)?
+                }
+            };
+            Ok(can_bypass)
+        } else {
+            // No active lock, so no need to bypass
+            Ok(true)
```
emmiegit: Thought: We should disinguish between "no lock exists" and "lock exists but user has permission to bypass".

One use case I'm thinking of is when a staff member has permission to edit despite a lock on a page, but they're editing in a personal capacity. For instance, a collaborative log can be locked due to being full or not accepting contributions, and presently moderators aren't going to be impeded in attempting to edit. Whereas, if we return the lock result, then the UI could display a little "⚠️ This page is locked for reason `foo bar`, but you have permission to edit it" type message.

emmiegit:
Exciting changes!

deepwell/src/endpoints/page_lock.rs
loop-index: Gotcha, will fix this in other places as well.

deepwell/src/services/page_lock/service.rs
loop-index: Got it, will remove this and move this logic to frontend.

deepwell/src/services/page_lock/service.rs
loop-index: Makes sense, will make a specialized output.

deepwell/src/services/page_lock/structs.rs
```diff
@@ -42,3 +41,9 @@ pub struct CreatePageLockInput {
     pub override_existing: bool,
     pub ip_address: IpAddr,
 }
+
+#[derive(Deserialize, Debug, Clone)]
+pub struct CheckLockBypassOutput {
+    pub lock_present: bool,
+    pub can_edit: bool,
+}
```
emmiegit, line 49: Hmm, yeah I think this is better than an enum.

emmiegit (approved):

PR #2921, opened 2026-05-22
[WJ-1373] [WJ-272] Add wikidot user serialization type

emmiegit:
This PR adds the `wikidot_user` table to the database, which represents user records imported from Wikidot. This concept allows us to describe user data for backfilled data, like imported page revisions from Wikidot. From a UI perspective, this lets us show a "this is a wikidot user, here's limited information" view when such a revision/user is clicked on.

However, if that user activates a proper Wikijump account, then a row in the `user` table is created, and that shadows the Wikidot user record.

Wikidot user records are immutable, meant to describe the user at the time of import. Any live or mutable operations (changing a bio, making a new page edit, updating user MFA) must happen on real (Wikijump) user records.

In code:
* The concept of `User` has been added, with two variants (Wikidot and Wikijump).
* `UserService::get()` and `::get_optional()` have been modified to return this new enum.
* This new enum has methods to assert / extract specifically the Wikijump or Wikidot variants from it, producing an error otherwise.
* `UserService::get_real()` and `::get_real_optional()` have been added to fetch Wikijump user records only.
* Because we now have a second place inserting into `known_user` (thus why this table was chosen), I have added helper functions for interacting with it.
* When adding users, there is a check to ensure that it is not overwriting a Wikidot user of that type (unless it's that user being activated).
* User-related stubs in `ImportService` have been implemented. For instance, uploading a Wikidot user avatar as part of this process.
* I've also implemented audit logging to `ImportService`.
* The endpoint `user_activate_from_wikidot` has been implemented.

Most notable however, is the serialization for `User`. When a `User::Wikijump` is being serialized, it is identical to the output of the normal Wikijump user record (i.e. what framerail is used to now). However, if it is a `User::Wikidot`, then it sets `"user_type": "wikidot"` and the fields are different. In order to support Wikidot user types in these places, framerail will need to discriminate against this value of `user_type` and have different processing that considers the different set of fields in this kind of record.

This is required since, ultimately, a Wikidot user record is not the same as Wikijump one. We can tell if it has been deleted, but not when. We obviously have no access to password or email information. This means that attempting to cram in Wikidot users into the regular `user` table will result in columns having different meanings between the two (for instance, a fictional `deleted_at` value for deleted users), which is a confusing and non-obvious way to represent that data.

I have also added integration tests for this change.


emmiegit:
It's been a bit, going to self-merge.

PR #2935, opened 2026-05-31
[WJ-1266][WJ-1267] Implement Page Lock UI

loop-index:
Implemented a `LockPane` to manage page locks.

Wikidot Layout:
image
Wikijump Layout:
image


framerail/src/lib/server/load/page.ts
```diff
@@ -1115,3 +1141,77 @@ const pageRestoreSchema = object({
...
+    return fail(500, {
+      form,
+      message: error.message,
+      code: error.code,
+      data: error.data
+    })
+  }
+}
+
+const pageLockSchema = object({
```
emmiegit, line 1181: Is there a difference with doing `object({ .. })` vs just `{ .. }`?

framerail/src/routes/[slug]/[...extra]/LockPane.svelte
```diff
@@ -0,0 +1,359 @@
...
+    index: number
+  ): keyof typeof data.internationalization {
+    if (lock.deleted_at) {
+      // Heuristic check to determine if a lock was overridden:
+      // If next lock was created within a short time of last lock being deleted
+      if (index > 0) {
+        const newerLock = lockHistory[index - 1]
+        const deletedAt = new Date(lock.deleted_at).getTime()
+        const newerCreatedAt = new Date(newerLock.created_at).getTime()
+        if (Math.abs(newerCreatedAt - deletedAt) < 5000) {
```
Comment (emmiegit) [inline framerail/src/routes/[slug]/[...extra]/LockPane.svelte:80]:
```suggestion
        if (Math.abs(newerCreatedAt - deletedAt) < 5000) {  // 5 seconds
```

emmiegit (approved):
LGTM

framerail/src/lib/server/deepwell/page.ts
```diff
@@ -341,6 +341,62 @@ export async function pageRestore(
...
+  from_wikidot: boolean
+  lock_type: string
+  page_id: number
+  user_id: number
+  reason: string
+}
+export async function pageLockHistory(
+  reqContext: RequestContext
+): Promise<PageLockModel[]> {
+  return client.request("page_lock_get_history", {}, reqContext)
```
Zokhoi: Deepwell api functions should use explicit page reference instead of only inferring from context, since when accessing the site's default page, the `params.slug` will be null and the page reference cannot be inferred.

framerail/src/lib/server/load/page.ts
Zokhoi: This `object` is the one imported from valibot for schema validation purposes.

framerail/src/lib/server/deepwell/page.ts
loop-index: Fair point for now, but ideally the request context should always be able to get the right "page" it's on unless absolutely not applicable (ie. forum pages, inter-site pages like user profile...) I'll think a bit more to make it work, will send in a commit to change.

Zokhoi: We should also consider that the context system with page slug would be more appropriate for a web client, and less so for an api client, so if we were to create a public api backed by deepwell in the future, it would still be better to use explicit page references.

loop-index: I see, agreed. Made some changes to pass in explicit page refs.

PR #2973, opened 2026-06-19
Create wikidot page lock in ImportService

emmiegit:
This part of the method was stubbed, and we have `PageLockService` now so we should use it here.


loop-index (approved):
LGTM

PR #2981, opened 2026-06-29
Updated deploy script

r2bean2:
this added a menu based deployment script using python-questionary, which is an outside dependecy. it is incldued in pip and most other package managers though.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -30,7 +30,7 @@ Once it's installed, ensure it is running:
 
 ## Building and Running
 
-**The helper script `install/local/deploy.py` is the primary way to manage a local Wikijump installation.**
+**The helper script `install/local/deploy-helper.py` is the primary way to manage a local Wikijump installation.**
 
 Under the hood, it is running the following:
 
@@ -55,11 +55,14 @@ The "action" corresponds to actions that `docker-compose` can do. Some common ac
 
 Note that in `docker-compose.yaml`, there are configuration options for the domains to use. For development purposes, these are set to `wikijump.localhost`. This is the domain you will be connecting to, e.g. `https://www.wikijump.localhost`. The TLD `.localhost` is just like the usual `localhost` domain. Even when running locally, HTTPS is used. Because this certificate is self-signed, you will need to dismiss the certificate warning.
 
-**Thus, you can run the following to start a local instance:**
+**all you need to do to use the script is run:**
 
 ```
-$ install/local/deploy.py up
+$ install/local/deploy-helper.py
 ```
+and select the options you want to use 
+
+**for your first build, select the up option from the actions menu**
 
 Because this first needs to build the images and sources, _this will take some time_. Docker's build cache will prevent future rebuilds from taking as long, though there are circumstances where a full rebuild will be necessary (such as updated dependencies).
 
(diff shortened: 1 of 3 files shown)


Zokhoi:
Uh, maybe you should use a grammar and spelling checker for all the strings and documentation, there are quite a bunch of spelling mistakes I've spotted in them.

r2bean2:
Ya I am sorry about that, I did rush on the final documentation because it was late and I was going on a trip the next day, plus I have dyslexia, so I’m not really good at spelling in the first place. I’ll look it over. I don’t have alot of time this or next week, but I’ll try to get around to it, and if not, I’ll get it done week after next.

docs/development.md
```diff
@@ -30,7 +30,7 @@ Once it's installed, ensure it is running:
 
 ## Building and Running
 
-**The helper script `install/local/deploy.py` is the primary way to manage a local Wikijump installation.**
+**The helper script `install/local/deploy-helper.py` is the primary way to manage a local Wikijump installation.**
```
emmiegit, line 33: The direct utility is useful for invocation of commands, the other is useful for exploring options or newer developers:
```suggestion
**You can use the helper script `install/local/deploy-help.py` to interactively navigate deployment for a local Wikijump installation.** Alternatively, if you're more familiar with Docker you can use `install/lock/deploy.py` directly.
```

docs/development.md
```diff
@@ -55,11 +55,14 @@ The "action" corresponds to actions that `docker-compose` can do. Some common ac
...
-**Thus, you can run the following to start a local instance:**
+**all you need to do to use the script is run:**
 
 ```
-$ install/local/deploy.py up
+$ install/local/deploy-helper.py
 ```
+and select the options you want to use 
+
+**for your first build, select the up option from the actions menu**
```
emmiegit, line 65: This seems kind of unnecessary. Either the reader takes the route of using the interactive script (where how to start a new instance is spelled out for them), or they want to learn more about the underlying Docker commands, in which case the original is what they would want to read.

docs/development.md
```diff
@@ -71,11 +74,16 @@ Because this first needs to build the images and sources, _this will take some t
...
+If you want to entirely _reset_ the containers, as their data is otherwise persistent even across restarts, use the down option in the helper script
+
+**you may find entering your choices in thge deploy script every time to be anoying. to skip stright to selecting the docker command, run:**
 
-```sh
-$ install/local/deploy.py down
 ```
+$ install/local/deploy-helper.py -s 
+```
+**you can also use --skip instead of -s**
```
emmiegit, line 84: Hmm, if people want to skip over options in the interactive helper, maybe the point of the helper should be changed to providing guidance on how to run different commands?

So you navigate the menu to find a thing you want to do (and haven't done before), and can run it like that, but it also shows you the `deploy.py` / Docker command for it, so you can just run that in the future.

emmiegit (commented):
My thoughts are that perhaps the script should be more "interactive documentation" and less "replacement for `deploy.py`", given that presently I feel like it is trying to do both and is weaker for it. A focus on one would make it more useful.

I also agree with Jochoi that there are a number of spelling etc. issues. I would make PR suggestions here, but I feel like the wider purpose of the script should be addressed first.

r2bean2:
I do see your point, but the reason I made the -s option was because I personally find it easier to just hit down arrow a couple of times then to add the docker command to the end of the other script. As it stands, I don’t even use the original script, because it’s just easier for me to use normal docker compose, with a little config stuff(sudo context is not so use, chdir isn’t really relevant when skipping the script, and you can get dev bindings by default with a little one time bashrc config. Maybe I’ll make a script for that). I do want to know, how hard was getting the dependencies to work? I may create pre compiled binaries if it’s hard

docs/development.md
r2bean2: It does do that now, it prints the command and asks for confirmation, and then prints it again.

PR #3024, opened 2026-07-20
[WJ-1168] Implement site bans

WackDog:
Implement site bans, including permanent and limited time bans.

- Added endpoints for site bans
- Added auto cleanup of expired bans through worker
- Removes/revokes user site membership and roles when banned
- Prevents user from receiving membership and getting roles when banned
- Added audit events for manual ban adding/removing
- Added integration tests covering basically all of the above

Preventing banned users from making applications is still TODO in site_ban.rs as advised by @emmiegit as the site application infrastructure is not there yet, I am happy to circle back to that todo after this is done.


deepwell/src/endpoints/site_ban.rs
```diff
@@ -0,0 +1,125 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+use super::prelude::*;
+use crate::models::relation::Model as RelationModel;
+use crate::services::relation::{CreateSiteBan, GetSiteBan, RemoveSiteBan, SiteBanData};
+use std::net::IpAddr;
+
+#[derive(Deserialize, Debug, Clone)]
+struct SetSiteBanInput {
```
emmiegit: Would `CreateSiteBanInput` be a better name?

deepwell/src/services/relation/site_ban.rs
```diff
@@ -59,24 +65,117 @@ impl RelationService {
...
+                reason: &metadata.reason,
             },
         )
         .await
         .or_raise(make_error)?;
 
-        // TODO: remove site member applications
-        // TODO: remove site roles
+        Ok(())
+    }
```
emmiegit, line 135: All site ban creation / removal events should be in the audit log. If I'm trying to see why user Alice was unbanned but I'm not seeing an entry for her unbanning, the log isn't serving its purpose.

For automated unbans, the acting user is [`SYSTEM_USER_ID`](https://github.com/scpwiki/wikijump/blob/develop/deepwell/src/constants.rs#L25) and the IP address is [localhost](https://doc.rust-lang.org/std/net/struct.Ipv6Addr.html#associatedconstant.LOCALHOST) (since Wikijump itself is performing the action).

deepwell/src/services/audit/structs.rs
```diff
@@ -60,6 +60,18 @@ pub enum AuditEvent<'a> {
...
+        site_id: i64,
+        user_id: i64,
+        banning_user_id: i64,
+        banned_until: Option<Date>,
+        reason: &'a str,
+    },
+    SiteBanRemove {
+        site_id: i64,
+        user_id: i64,
+        unbanning_user_id: i64,
```
emmiegit, line 73: We should have the ability to log reasons for unbans as well (e.g. linking a successful appeal)

deepwell/src/services/audit/structs.rs
```diff
@@ -300,6 +312,40 @@ impl<'a> AuditEvent<'a> {
...
+            AuditEvent::SiteBanCreate {
+                site_id,
+                user_id,
+                banning_user_id,
+                banned_until,
+                reason,
+            } => RawAuditEvent {
+                event_type: "site_ban.create",
+                ip_address,
+                user_id: Some(banning_user_id),
```
emmiegit: Main `user_id` here should be the banned user, not the actor.

deepwell/src/services/audit/structs.rs
```diff
@@ -300,6 +312,40 @@ impl<'a> AuditEvent<'a> {
...
+                site_id: Some(site_id),
+                page_id: None,
+                extra_id_1: Some(user_id),
+                extra_id_2: None,
+                extra_string_1: banned_until.map(|date| Cow::Owned(date.to_string())),
+                extra_string_2: Some(Cow::Borrowed(reason)),
```
emmiegit: Maybe swap these, so `reason` is the first string field in both?

emmiegit: The other thing some audit events do is add metadata as JSON here.

deepwell/src/api.rs
```diff
@@ -323,6 +323,11 @@ async fn build_module(app_state: ServerState) -> Result<RpcModule<ServerState>>
     register!("site_update", site_update);
     register!("site_domain", site_get_domain);
 
+    // Site bans
+    register!("site_ban_set", site_ban_set);
+    register!("site_ban_get", site_ban_get);
+    register!("site_ban_remove", site_ban_remove);
```
emmiegit, line 329: (Not a blocker for this PR) We will eventually need an endpoint for listing bans for a site (i.e. in the admin panel equivalent)

deepwell/src/endpoints/site_ban.rs
WackDog: Agreed, renamed to CreateSiteBanInput

deepwell/src/services/relation/site_ban.rs
WackDog: Updated to use normal audit remove path with wikijump performing the action

deepwell/src/services/audit/structs.rs
WackDog: Added reason as a required field in audit event for both manual and automatic removals, with automatic already prewritten. Can still submit an empty string if no reason is needed.

deepwell/src/services/audit/structs.rs
WackDog: Swapped as suggested.

deepwell/src/services/audit/structs.rs
WackDog: Swapped as suggested.

deepwell/src/api.rs
WackDog: I can add this to the Jira if it is not there already

emmiegit: This could also be a follow-up Jira item.

deepwell/src/services/relation/site_ban.rs
```diff
@@ -114,4 +215,70 @@ impl RelationService {
...
+            }
+
+            Self::remove_site_ban_with_audit(
+                ctx,
+                RemoveSiteBan {
+                    site_id: site_ban.dest_id,
+                    user_id: site_ban.from_id,
+                    removed_by: SYSTEM_USER_ID,
+                },
+                IpAddr::V4(Ipv4Addr::LOCALHOST),
```
emmiegit: The seeder also uses IPv6:
```suggestion
                IpAddr::V6(Ipv6Addr::LOCALHOST),
```

(Maybe we should make this another constant, the "system IP address")

deepwell/src/services/relation/site_ban.rs
```diff
@@ -59,24 +65,119 @@ impl RelationService {
...
 
-        // TODO: remove site member applications
-        // TODO: remove site roles
+        Ok(())
+    }
 
-        create_operation!(
-            ctx, SiteBan, Site, site_id, User, user_id, created_by, &metadata,
-            make_error,
+    pub async fn remove_site_ban_with_audit(
```
emmiegit, line 136: I don't think it makes sense to have both `remove_site_ban()` and `remove_site_ban_with_audit()`. The default action for removing a site ban should include doing audit logging.

Is this a case where `remove_site_ban()` is generated by the macro? Since it's a bit complicated, we can merge this as-is and I'll work on modifying it to allow `remove_site_ban()` to be an override which also does the audit logging.

deepwell/src/services/relation/site_ban.rs
```diff
@@ -114,4 +215,70 @@ impl RelationService {
 
         Ok(())
     }
+
+    /// Soft-deletes all active site bans whose expiry date has passed.
+    ///
+    /// Permanent bans, represented by a null `banned_until`, are left active.
```
emmiegit: Suggestion:
```suggestion
    /// Permanent bans, represented by a null `banned_until`, are left unmodified.
```

deepwell/src/services/relation/site_ban.rs
WackDog: Switched

deepwell/src/services/relation/site_ban.rs
WackDog: Yes it is generated by the macro, have left it for now as per your suggestion

deepwell/src/services/relation/site_ban.rs
WackDog: Updated.

emmiegit (approved):
I'll follow up with the relation generation macro

PR #3039, opened 2026-07-28
[WJ-1168] Add proc macro for generating relation structs and methods

emmiegit:
Up to this point, we used a normal `macro_rules!` macro to generate the code for relations. This had a few problems, such as it being cumbersome to disable generation of creation structs or methods, and the inability to exclude `metadata` as a field if it wasn't used for that relation (e.g. is `()`).

With the new requirement to permit disabling `remove_*()` method generation as well, I decided that the code was getting too complicated and it was time to use a procedural macro to generate the code instead. This allows for specifying a relation in a more readable syntax, as well as handling some niceties for us (such as doing the `PascalCase` to `snake_case` conversion for fields automatically).

The arguments accepted are as follows:

```rust
impl_relation! {
    // Required. Name of relation in PascalCase.
    name => NameOfRelation,

    // Required. Name and object type of "dest".
    dest => destination_field: Page,

    // Required. Name and object type of "from".
    from => from_field: User,

    // Optional. Associated data type. Default is ().
    data => RelationExtraData,

    // Optional. How to generate the creation code for the relation. Default is true.
    //
    // Values are:
    //  * true   - Generate public struct and public creation method.
    //  * false  - Generate public struct and private "inner" creation method.
    //             Public creation method must be defined by the programmer.
    //  * extern - Generate neither a struct nor creation method.
    //             Both must be defined by the programmer.
    create_fn => false,

    // Optional. How to generate the removal code for the relation. Default is true.
    // Same values as above.
    remove_fn => true,
}
```

With this done, I changed the site member and ban relations to add the audit logging, and did some refactoring to make the struct handling in `endpoints/` consistent.

Naturally, I also deleted the old macro definition.


deepwell/src/services/relation/user_bot_owner.rs
```diff
@@ -59,15 +59,13 @@ pub struct UserBotMetadata {
...
-    owner_user,
-    UserBotMetadata,
-    NO_CREATE_IMPL_OR_STRUCT,
-);
+impl_relation! {
+    name => UserBotOwner,
+    dest => bot_user: User,
+    from => owner_user: User,
+    data => UserBotMetadata,
+    create_fn => false,
```
WackDog: Should this be `create_fn => extern` rather than `false`? The previous invocation used NO_CREATE_IMPL_OR_STRUCT and the parser test also models UserBotOwner using `extern` . My understanding is that false still generates the public creation struct, whereas extern generates neither the struct or method.

deepwell/relation-impl-derive/src/case.rs
```diff
@@ -0,0 +1,35 @@
+//! Helper utilities in processing different naming cases.
+
+/// Convert `PascalCase` to `snake_case`.
+pub fn pascal_to_snake_case(value: &str) -> String {
+    let mut output = String::new();
+    for (i, ch) in value.char_indices() {
+        if i > 0 && ch.is_uppercase() {
+            output.push('_');
+        }
```
WackDog, line 9: How should names with acronyms be handled here? This implementation would make all uppercase letters a seperate word like SCPName would be s_c_p_name. Could be intentional ofc

deepwell/relation-impl-derive/src/test.rs
```diff
@@ -0,0 +1,148 @@
+use crate::parse::RelationSettings;
+use crate::types::{GenerateMethod, RelationObjectType};
+use syn::Type;
+
+#[test]
```
WackDog, line 5: Non-blocking: would it be useful to add compile tests for the generated output, perhaps with trybuild?

deepwell/src/services/relation/site_member.rs
```diff
@@ -70,9 +72,77 @@ impl RelationService {
...
+
+        AuditService::log(
+            ctx,
+            ip_address,
+            AuditEvent::JoinSiteMember {
+                user_id,
+                site_id,
+                joining_user_id: created_by,
+            },
         )
```
WackDog, line 95: Non-blocking: are there integration tests covering these new site-member audit events?

deepwell/relation-impl-derive/src/case.rs
emmiegit: This is just the standard `snake_case` conversion algorithm (used by `serde` etc.)

For the case where you have "SCP Name" then in PascalCase it would be `ScpName`.

WackDog: gotcha, lgtm

deepwell/relation-impl-derive/src/test.rs
emmiegit: Hmm, that's worth considering. I originally thoguht about doing tests (in the proc macro crate) for parsing and expansion, but then I thought it would be easier to just ensure `deepwell` itself built rather than setting up a duplicate system for testing. But maybe `trybuild` is closer to what I want here.

If I do implement this, it would be in a follow-up PR.

deepwell/src/services/relation/site_member.rs
emmiegit: There are not.

emmiegit: Added in the most recent commits.

deepwell/src/services/relation/user_bot_owner.rs
emmiegit: When it's `false`, it creates a private creation method and struct, which is used in the custom public creation method. But you're right, there isn't a way to disable public struct creation except with `extern`; we need an option for "create private/inner, but nothing public".

emmiegit: I've updated the PR to add separate settings for "create private/inner struct+method only" and "create private/inner struct+method, and a public struct for convenience".

deepwell/relation-impl-derive/src/expand.rs
```diff
@@ -0,0 +1,448 @@
...
+            #[derive(Deserialize, Debug, Copy, Clone)]
+            pub struct #remove_struct {
+                pub #dest_name: i64,
+                pub #from_name: i64,
+                pub removed_by: i64,
+            }
+        }
+    } else {
+        quote! {}
+    };
```
WackDog, line 399: Does `remove_fn => private_only` currently leave the generated method without its input struct?
For `PrivateWithoutStruct`, `generate_pub_struct` is false, so this branch emits no `Remove{Relation}` struct. However,  the generated `remove_*_inner()` method below still destructures `Remove{Relation}`. Should this mode generate a private removal struct, or should `private_only` not be supported for `remove_fn`?

deepwell/relation-impl-derive/src/parse.rs
```diff
@@ -0,0 +1,162 @@
...
+                //
+                //  create_fn => public
+                "create_fn" => {
+                    error_if_set!(create_fn);
+                    let option = GenerateMethod::parse(input)?;
+                    create_fn = Some(option);
+                }
+
+                // Designate how the remove method should be generated
+                // This key is optional, default is "true".
```
WackDog, line 118: should this say the default is public, with the example using `remove_fn => public`? The parser doesn't accept true anymore

WackDog:
New commits LGTM, only remaining this is still this comment, https://github.com/scpwiki/wikijump/pull/3039/changes#r3692016963

emmiegit:
Also, noting that I played around a bit with `trybuild`, and while I think it's a useful tool, this would take a bit too much work to get it up when our main use here is kind of niche. (Which is also why it's hard to set up here)

deepwell/relation-impl-derive/src/expand.rs
emmiegit: I was having trouble conceptualizing this for a while, but you're right, this is a problem. Thinking on the definition of `private_only`, I think we should just have a private removal struct.

emmiegit:
I've pushed a fix for the comment you had above. Now, for `remove_fn => private_only`, it generates a private removal struct to match the private inner removal method. Tested by setting that field on a random relationship and building.

deepwell/relation-impl-derive/src/expand.rs
```diff
@@ -0,0 +1,454 @@
...
+                create_struct_def = Some(quote! {
+                    #[derive(Deserialize, Debug, Clone)]
+                    pub struct #create_struct {
+                        pub #dest_name: i64,
+                        pub #from_name: i64,
+                        pub created_by: i64,
+                    }
+                });
+            }
+        }
```
WackDog, line 274: Could `create_fn => private_only` have the same issue here when the relation has no metadata?

In that case, `generate_pub_struct` is false, so this branch doesn’t create a `Create{Relation}` struct. But the generated `create_*_inner()` method still appears to use `Create{Relation}` as its input type

Should this branch generate a private struct in that case, like the updated remove_fn handling does?

emmiegit: Hmm. I'd have to check at home to be sure, but you may be right. This case is not presently used in any relations since the "inner" methods generally act on or consider some metadata.

WackDog (approved):
LGTM

deepwell/relation-impl-derive/src/expand.rs
emmiegit: You are correct, this case wasn't considered. As discussed, I'll make a follow-up PR.

PR #3054, opened 2026-08-03
[WJ-1354 Pt.3] Migrate old frontend components

Zokhoi:
Third step in the series for the old front end components migration. (Should be final?)

This migrates basically all of the [components](https://github.com/scpwiki/wikijump/tree/legacy-php/web/modules/components) package from the old front end.


hoah2333 (approved):
lgtm

PR #3061, opened 2026-08-16
Serve user avatar through wws

Zokhoi:
This allow user avatars to be served through wws by user id (with s3 hash caching), such that the blob data need not be fetched though deepwell.

### DIFF
--- a/framerail/src/lib/server/load/user.ts
+++ b/framerail/src/lib/server/load/user.ts
@@ -1,7 +1,6 @@
 import defaults from "$lib/defaults"
 
 import { authGetSession } from "$lib/server/auth/getSession"
-import { getFileByHash } from "$lib/server/deepwell/file"
 import { translate } from "$lib/server/deepwell/translate"
 import { userEdit, userView } from "$lib/server/deepwell/user"
 import { loadSiteInfo } from "$lib/server/load/site-info"
@@ -75,17 +74,6 @@ export async function loadUser(
       parentData.user_session?.user?.user_id !== response.data.user.user_id
 
     viewData.user = sanitizeUserData(response.data.user, isViewingAnotherUser)
-
-    // Get user avatar image
-    if (response.data.user.avatar_s3_hash !== null) {
-      const avatar = await getFileByHash(
-        new Uint8Array(response.data.user.avatar_s3_hash)
-      )
-      const dataurl = `data:${avatar.type};base64,${Buffer.from(
-        await avatar.arrayBuffer()
-      ).toString("base64")}`
-      viewData.user.avatar = dataurl
-    }
 
     translateKeys = {
       ...translateKeys,
--- a/wws/src/handler/avatar.rs
+++ b/wws/src/handler/avatar.rs
@@ -85,8 +85,8 @@ async fn fetch_blob(
 
     Ok((
         BlobData {
-            size: file_size.unwrap_or(0),
-            mime: mime_type.unwrap_or("text/plain".into()),
+            size: file_size.expect("File size should be known"),
+            mime: mime_type.expect("File mime type should be known"),
         },
         body,
     ))
--- a/wws/src/handler/avatar.rs
+++ b/wws/src/handler/avatar.rs
@@ -115,6 +115,7 @@ pub async fn handle_user_avatar(
     };
 
     let result = Response::builder()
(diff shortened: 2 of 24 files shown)


wws/src/handler/avatar.rs
```diff
@@ -0,0 +1,130 @@
...
+        Ok((result, status_code)) => {
+            assert_eq!(
+                status_code,
+                StatusCode::OK,
+                "head_object() succeeded but did not reply 200",
+            );
+            (result.content_length, result.content_type)
+        }
+        Err(error) => {
+            error!(s3_hash = s3_hash, "Cannot get blob metadata: {error}",);
```
emmiegit:
```suggestion
            error!(s3_hash = s3_hash, "Cannot get blob metadata: {error}");
```

wws/src/handler/avatar.rs
```diff
@@ -0,0 +1,130 @@
...
+            .await;
+
+            return Err(response);
+        }
+    };
+
+    Ok((
+        BlobData {
+            size: file_size.unwrap_or(0),
+            mime: mime_type.unwrap_or("text/plain".into()),
```
emmiegit: Have these ever been `None` in your testing? I feel like we should error on that case rather than getting defaults, I'm not sure assuming the size is correct.

wws/src/handler/avatar.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+pub async fn handle_user_avatar(
+    State(state): State<ServerState>,
+    Path(user_id): Path<String>,
+    headers: HeaderMap,
+) -> Response {
+    info!(user_id = user_id, "Returning user avatar",);
+
+    let Ok(user_id_i64) = user_id.parse::<i64>() else {
```
emmiegit: Since `user_id` (the string) isn't used in the success case, we can just shadow it:
```suggestion
    let Ok(user_id) = user_id.parse::<i64>() else {
```

wws/src/state.rs
```diff
@@ -255,4 +255,58 @@ impl ServerStateInner {
...
+        headers: &HeaderMap,
+        user_id: i64,
+    ) -> ResponseResult<String> {
+        match self.get_avatar(user_id).await {
+            Ok(Some(avatar_s3_hash)) => Ok(avatar_s3_hash),
+            Ok(None) => {
+                error!(
+                    user_id = user_id,
+                    "Cannot complete request, no such user or no avatar set",
+                );
```
emmiegit, line 287: How does this work if a user doesn't have an avatar set? Shouldn't that just fall back to a default (on the frontend) rather than erroring?

Zokhoi: Right now in my wip code for framerail client status box, it checks if the `user.avatar_s3_hash` is present, and serves the default avatar if false.

I'm also thinking if serving default avatar from wws would make sense for modules that list a lot of users, such as vote list or listusers, as there would be fewer fields to transmit per user.

wws/src/handler/avatar.rs
Zokhoi: I just tested uploading though minio web ui, and for unknown files it uploaded as the `application/octet-stream` mime type. I guess this would be better as a default mime, as this indicates an "arbitrary binary" instead of text.

Not explicitly providing the content length is also an option here, though not sure if the right one.

emmiegit: Try uploading via a local deepwell API call. All blobs uploaded to S3 are run through magic to get their correct MIME type, so I believe it should be populated properly.

wws/src/state.rs
emmiegit: Makes sense, so it only requests if there is an avatar. I think this is fine then.

wws/src/handler/avatar.rs
Zokhoi: Changed to using `.expect()` now.

framerail/src/lib/server/load/user.ts
```diff
@@ -1,7 +1,6 @@
 import defaults from "$lib/defaults"
 
 import { authGetSession } from "$lib/server/auth/getSession"
-import { getFileByHash } from "$lib/server/deepwell/file"
```
emmiegit, line 4: I believe that, with this, `getFileByHash` is now unused and can be removed.

emmiegit (approved):

PR #3088, opened 2026-09-01
Split deepwell coverage into separate workflow

emmiegit:
Current build issues with recursion depth are due to nightly Rust. I have two changes here, splitting the workflow so coverage is separate (relevant since we are now avoiding the tarpaulin container for the main test build), and to no longer use nightly for building either. Originally, these two processes were in the same workflow to avoid having to repeat the test setup twice.

### DIFF
--- a/.github/workflows/deepwell.yaml
+++ b/.github/workflows/deepwell.yaml
@@ -18,11 +18,104 @@ on:
       - prod
 
 jobs:
-  test_and_coverage:
-    name: Build, Test and Coverage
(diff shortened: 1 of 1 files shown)


emmiegit:
@gitguardian Not a real secret, used only for temporary test containers in GitHub CI.

emmiegit:
Okay, it's working now. Next is to rebase it together and update the PR description.

Zokhoi (approved):

